PR #18, opened 2018-06-05
Create my profile page

### LINKED ISSUE #8: As a client I am able to see my user profile

jswk:
The page shows read-only user information, as fetched from check-in.
Authorization for profile page is created, along with 401 error page.
A link is added to the navbar, which is shown when user is logged in.

Fixes #8.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,4 +4,10 @@ require "turbolinks"
 
 class ApplicationController < ActionController::Base
   include Turbolinks::Redirection
+
+  private
+
+  def authenticate_user!
+    render :file => "public/401", :status => :unauthorized, :layout => false unless current_user
+  end
 end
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -4,4 +4,10 @@
 
 class ApplicationController < ActionController::Base
   include Turbolinks::Redirection
+
+  private
+
+    def authenticate_user!
```
mkasztelnik: Devise has `authenticate_user!` method builded in. If it is used in before filter than user will be redirected into logging page. In this case we should redirect user to checkin

jswk: Is the redirection to check-in already in place? If not, how should it be set up?

mkasztelnik: This can be useful https://github.com/plataformatec/devise/wiki/How-To:-Redirect-to-a-specific-page-when-the-user-can-not-be-authenticated

mkasztelnik:
One small remark about 401 template page and this PR is ready to be merged :)

PR #19, opened 2018-06-06
Remove default session devise routes

mkasztelnik:
Since marketplace will be 100% based on external IDP (configured using omniauth) no session related URLs should be defined. Here session routes are removed, only logout option remains.


bwilk:
@mkasztelnik could you link documentation for devise/sessions route??

mkasztelnik:
https://github.com/plataformatec/devise/wiki/How-To:-Change-the-default-sign_in-and-sign_out-routes

Issue #44, opened 2018-06-15
As registered user I'm able to see the history of previously ordered services


roksanaer:

mkasztelnik:
Dedicated section "My services" is already implemented. It shows all user orders.

Issue #41, opened 2018-06-15
As registered user I'm able to see the status of ordered and active services


roksanaer:

mkasztelnik:
Done

Issue #40, opened 2018-06-15
As a user I'm able to place an order of the selected service


roksanaer:

mkasztelnik:
First prototype will include:
  * Order model connected with service and user
  * Order note - place where updates (e.g. messages from service provider) and status changes will be reported

Two new views will be created:
  1. /orders - show all user orders (pagination will be used here)
  2. /orders/:id - show order detail, notes and status changes

On /services/:id additional form "order" will be added.

mkasztelnik:
Implemented in the scope of #49

Issue #29, opened 2018-06-11
JIRA integration for marketplace 


bwilk:
Marketplace will be integrated with JIRA for order management. 
There is a need for implementation of integration with JIRA system. 

Malgorzata K. should be able to answer questions about:
- when JIRA for EOSC-hub will be ready
- how to access it 
- etc.

bwilk:
I've already asked Malgorzata about details so I'll report as soon as I get the answers.

bwilk:
https://docs.cyfronet.pl/display/FID/JIRA

mkasztelnik:
Currently we have following integrations:
  * when Order is created issue is created in JIRA
  * when issue status is changed in Jira it is reflected in the marketplace
  * when user add question to the order comment is created in assigned Jira issue
  * when comment is created in Jira it is reported as service provider message in the marketplace

So I beliete this can be closed. @bwilk, @roksanaer can you confirm this or am I missing something?

roksanaer:
confirm - 4 mentioned elements are fulfilling our requirements for the integration

PR #49, opened 2018-06-15
Order creation

mkasztelnik:
Basic ordering capabilities. Logged in user can click `Order` button on service page and as a conclusion ordering process begins. At the beginning `Order` record is created in our database with first `OrderChange` record. This change order status into `created` and set appropriate message. At the end of order creation process, the delayed job is triggered to register order in an external system (JIRA in our case). 

`Order::RegisterJob` invokes `Order::Register` service which is a placeholder for JIRA integration. At the end, it adds new order change (change status from `created` into `registered` with the dedicated message). 

This PR is related to following user stores:
  * #44 
  * #42 
  * #41
  * #40

I'm not selecting these issues as "closable" because my implementation is not styled and views are ugly - thus according to EU "read" understanding this is not ready.


app/views/orders/_order.html.haml
```diff
@@ -0,0 +1,9 @@
+%dt
+  Service:
+  = link_to order.service.title, order.service
```
bwilk: here you should link to order not service

app/models/order.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class Order < ApplicationRecord
+  STATUSES = {
+    created: "created",
+    registered: "registered",
```
bwilk: what does 'register' mean here?

app/services/order/register.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def call
+    register_in_jira! &&
+    update_status!
+  end
+
+  private
+
+    def register_in_jira!
+      # TODO: Jira integration. This method should throw error on errors which
```
bwilk: @michal-szostak take this into account while solving #29

app/models/order.rb
mkasztelnik: An order is registered in 3rd party service (e.g. JIRA)

app/views/orders/_order.html.haml
mkasztelnik: done

mkasztelnik:
All reported issues solve, this is ready to be merged after @bwilk approve.

PR #64, opened 2018-06-28
Configure dev, test database though ENV variables

mkasztelnik:
On @michal-szostak request dedicated way to configure PostgreSQL though environment variables was added. Take a look into `README.md` database section for details.


michal-szostak:
Everything works as expected, approved

PR #97, opened 2018-07-04
59 new layout

kosmidma:
* Header and pagination styling
* Adding static elements in Service


app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -101,4 +107,44 @@ input:-moz-placeholder, textarea:-moz-placeholder { /* Firefox 18- */
...
+      border-bottom: 1px solid $body-color;
+      font-weight: 800;
+    }
+  }
+  .text-dark{
+    color: $body-color;
+  }
+}
+.media {
+  border-bottom: 1px solid #d5dadd;
```
michal-szostak: Is it reusable? Shouldnt it be an extendable style or at least color constant?

app/views/services/_service.html.haml
```diff
@@ -1,3 +1,31 @@
...
+          %span.small.ml-1.font-weight-bold
+            (2.3/5)
+          %a{href: "", class: "small ml-3"}
+            5 opinions
+
+    .ml-5
+      %form.d-flex.justify-content-end
+        .form-group.form-check
+          %input{type: "checkbox", class: "form-check-input", id: "example"}
+          %label{class: "form-check-label", for: "example"}
```
michal-szostak: Label is separate from input, and the input has unset id. this causes clicking on *any* checkbox label to check first service in the list (first rendered element with given id).
It should be reworked to have unique IDs (i.e. including index / id of the service) or to not have id and having label enclose input.

app/views/services/_services.html.haml
```diff
@@ -1,4 +1,22 @@
...
+          20
+      %li.page-item
+        %a.page-link
+          30
+      %li.page-item.disabled
+        %a.page-link.text-dark
+          Items on page
+
+  .mt-3.mb-4
+    = will_paginate services, class: 'asdf'
```
michal-szostak: Remove "asdf" class? It does not exist anyway

app/views/services/_service.html.haml
kosmidma: I removed the id

app/views/services/_services.html.haml
kosmidma: fixed

mkasztelnik:
We should avoid of using asset pipeline because it will be removed in next rails version, this is also connected with the place where images should be placed. Right now some of them are placed into `app/assets/images` - this is wrong location. Please take a look into https://github.com/rails/webpacker#usage and use webpack to manage all possible assets.

mkasztelnik:
Maybe small element to improve, right now there is no spacing between the footer and main page

footer

Maybe small spacing can be added there? @abacz? The rest is ok for now :+1:

PR #129, opened 2018-07-12
51-create-jira-issue-on-order

michal-szostak:
# What is in this PR

This PR contains JIRA integration from the marketplace to JIRA.
This includes 2 things:
1) JIRA issue is created when new order is created.
2) comment is created when question to order is created.

Additionally jira rake task has been created which can verify
whether jira connection settings are correct

# What is *not* in this PR

Error handling on the JOB level has not been done as there 
is no consensus about how it should be handled.
Linking of question (order_change) to jira issue has not been done
as the question is only a virtual entity, and including 
additional specific fields in order_change table does not seem
like a scalable idea (@mkasztelnik - maybe you have some better ideas?)

# Where are my issues going to be?

The default project for development is 
https://jira.plgrid.pl/jira/browse/EHD, but all connection 
parameters can be overwriten by specific ENV variables (check README.md
for more information)


app/services/order/register_question.rb
```diff
@@ -1,18 +1,30 @@
...
+
+    comment = issue.comments.build
+    if comment.save(body: @question)
+      # :TODO: maybe question / order change should be extended to
+      # contain jira comment id and status, for easier disaster recovery
+      # and tracking
+      true
+    else
+      # :TODO: set errors on @question
+      raise JIRACommentCreateError.new(@question)
```
mkasztelnik: this can be simplified into `comment.save(body: @question) || raise JIRACommentCreateError.new(@question)`

app/services/order/register.rb
```diff
@@ -14,10 +21,19 @@ def call
...
 
     def register_in_jira!
-      # TODO: Jira integration. This method should throw error on errors which
-      #       can be connected with temporary jira issue. Than registration
-      #       process will be retired by delayed job.
-      true
+      issue = JIRA_CLIENT.Issue.build
+      if issue.save(fields: { summary: "Requested realization of #{@order.service.title}",
+                           project: { key: JIRA_PROJECT_KEY },
+                           issuetype: { id: JIRA_ISSUE_TYPE_ID } })
```
mkasztelnik: Where can we find information about who requested service?

app/services/order/register.rb
```diff
@@ -14,10 +21,19 @@ def call
...
+        @order.jira_active!
```
mkasztelnik: This method will do DB update and 2 lines bellow another DB save. Maybe we can have only one?

config/boot.rb
```diff
@@ -3,4 +3,9 @@
 ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
 
 require "bundler/setup" # Set up gems listed in the Gemfile.
-require "bootsnap/setup" # Speed up boot time by caching expensive operations.
+
+# enabling bootsnap breaks debugging, the problem apparently exists only for rails 5.2
+# after upgrading rails it can be removed
```
mkasztelnik: Can you elaborate more about this issue? Some link to official rails issue would be perfect.

config/initializers/jira.rb
```diff
@@ -0,0 +1,20 @@
...
+    context_path: jira_config_env["context_path"],
+    auth_type: :basic,
+    use_ssl: (/^https\:\/\// =~ jira_config_env["url"])
+}
+
+JIRA_OPTIONS = options
+
+JIRA_CLIENT = JIRA::Client.new(options)
+JIRA_PROJECT_KEY = jira_config_env["project"]
+JIRA_ISSUE_TYPE_ID = jira_config_env["issue_type_id"]
```
mkasztelnik: Maybe instead of using constants we can have simple model class? Something like (in `app/models/jira/client.rb`):

```ruby
class Jira::Client < JIRA::Client
  def initialize
    # read required options and initialize jira client
    options = { ... }
    super(options)
  end

  self.project_key
    ...
  end

  self.issue_type
    ...
  end
end
```

Than we have more OO style of programming:

```ruby
Jira::Client.new.some_method
Jira::Client.issue_type
```

This has an additional future plus: according to @roksanaer there can be multiple jira backends. If such a situation will happen we can simply refactor `Jira::Client` constructor to receive `service` as a parameter and use it to configure the dedicated client. We can go even further and instead of inheritance we can create composition with more specialized methods:

```ruby
class Jira::Client
  def register_order(order)
    ...
  end

  def new_comment(order_change)
    ...
  end
end
```

config/initializers/jira.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+JIRA_CONFIG = YAML.load(ERB.new(File.read("#{Rails.root}/config/jira.yml")).result)
```
mkasztelnik: This can be simplified:

```ruby
Mp::Application.config_for(:jira)
```

config/jira.yml
```diff
@@ -0,0 +1,28 @@
+default: &default
+  username: <%= ENV["MP_JIRA_USERNAME"] || Rails.application.credentials.jira[:development][:username] %>
```
mkasztelnik: You can use `Rails.env` to check in which environment application is started. This should allow you to remove the very similar code from this configuration.

db/migrate/20180705112318_add_issue_id_to_orders.rb
```diff
@@ -0,0 +1,5 @@
+class AddIssueIdToOrders < ActiveRecord::Migration[5.2]
+  def change
+    add_column :orders, :issue_id, :integer
```
mkasztelnik: Are we storing only issue number (e.g. 1), not whole issue id (`MP-1`)?

db/migrate/20180705112813_add_issue_status_to_orders.rb
```diff
@@ -0,0 +1,6 @@
+class AddIssueStatusToOrders < ActiveRecord::Migration[5.2]
+  def change
+    add_column :orders, :issue_status, :integer, default: 1, null: false
+    change_column_default :orders, :issue_status, 2
```
mkasztelnik: Why not delete this line and update default in line 3?

lib/tasks/jira.rake
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require "colorize"
+
+namespace :jira do
+  desc "Check JIRA configuration"
+  error = false
+
+  def print_ok!
```
mkasztelnik: Maybe simply `ok!` ?

lib/tasks/jira.rake
```diff
@@ -0,0 +1,141 @@
...
+        # following tests require project to exist, if it does not return
+        next
+      end
+
+      begin
+        issue_type = JIRA_CLIENT.Issuetype.find(JIRA_ISSUE_TYPE_ID)
+
+        issue_name = issue_type.attrs["name"]
+
+        puts "Trying to manipulate issue..."
```
mkasztelnik: Should we be able to do jira content manipulation on production? If not then this should be only runnable in the development environment.

spec/jobs/order/register_job_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+require "rails_helper"
+
+RSpec.describe Order::RegisterJob do
+  let(:order_owner) { create(:user) }
+  let(:register_service) { instance_double(Order::Register) }
+  order = nil
+
+  before(:each) do
+    order = create(:order, user: order_owner)
+    expect(Order::Register).to receive(:new).
```
mkasztelnik: You want to check `new` method invocation of the intention of the text is to check if service is invoked? If the second one than `expect` here can be replaced with `allow`.

spec/jobs/order/register_job_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  it "triggers registration process for order" do
+    expect(register_service).to receive(:call)
+    described_class.perform_now(order)
+  end
+
+  it "handles exception thrown by Order::Register" do
+    expect(register_service).to receive(:call).and_raise(Order::Register::JIRAIssueCreateError.new(order))
+
+    described_class.perform_now(order)
```
mkasztelnik: Is seems like this test is not finished yet because error handling mechanism is not defined yet. Maybe some `pending "Waiting for error handling to be implemented"` should be added here?

spec/jobs/order/register_job_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Order::RegisterJob do
+  let(:order_owner) { create(:user) }
+  let(:register_service) { instance_double(Order::Register) }
+  order = nil
```
mkasztelnik: `order` can be created here instead of each. The effect is the same.

spec/factories/order/questions.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :question, class: Order::Question do
+    sequence(:text) { |n| "question #{n} message" }
+    to_create { |instance| Order::Question::Create.new(instance) }
```
mkasztelnik: Can you explain the purpose of this line?

spec/jobs/order/register_question_job_spec.rb
```diff
@@ -20,13 +20,18 @@
   end
 
   it "does nothing when change is not a question" do
-    question = create(:order_change,
-                      message: "State change",
-                      status: :registered,
-                      order: order)
+    question.author = nil
```
mkasztelnik: This can lead to more complication understanding test. Let's assume that our test case will be long (more than can be shown on one screen). If we are using more lets and what more modifying objects created in let we need to scroll each time we want to understand the test. I propose to have `order_change` creation inline when it is specific for the only test (and setting author to `nil` is making it specific).

spec/services/order/register_question_spec.rb
```diff
@@ -3,5 +3,31 @@
 require "rails_helper"
 
 RSpec.describe Order::RegisterQuestion do
-  pending "Specs should be added when service implementation will be done"
+  let(:user) { create(:user) }
+  let(:service) { create(:service) }
+  let(:order) { create(:order, user: user, service: service, issue_id: 1) }
+  let(:question) { "Question message" }
+  let(:comment) { double("Comment") }
```
mkasztelnik: Is it overridden in line 15?

README.md
```diff
@@ -70,6 +70,47 @@ or you can also use `systemctl`, it shouldn't matter which one you use.
...
+default username and password is: `admin/admin`.
+Make sure that environmental variables are set as follows:
+
+```
+MP_JIRA_PROJECT=MP 
+MP_JIRA_USERNAME=admin
+MP_JIRA_CONTEXT_PATH=/jira
+MP_JIRA_PASSWORD=admin
+MP_JIRA_URL=http://localhost:2990
+MP_JIRA_ISSUE_TYPE_ID=10000  #this might be different
```
mkasztelnik: Since we have a lot o ENV which can be set up in development it is time to introduce [`dotenv`](https://github.com/bkeepers/dotenv) in another PR

mkasztelnik:
@michal-szostak very nice PR, small changes requested. We can store additional information in `OrderChange` model if needed.

app/services/order/register.rb
michal-szostak: There is no information, because details of the order has not been determined yet. As you can see summary is also a placeholder.
I assume, that when it is decided order issuer will be added to summary, or something.

app/services/order/register.rb
michal-szostak: What do you mean? `issue.save` is not DB update. Or do you mean in another method, where order is created / updated?

config/boot.rb
michal-szostak: Finding solution to this problem took several sites, but first link from google gives this:
https://github.com/Shopify/bootsnap/issues/168

config/initializers/jira.rb
michal-szostak: I'll leave aside the question about multiple jira backends, as I was discussing it in the beginning...
Either way, I do agree that model class seems like a much more elegant solution

config/initializers/jira.rb
michal-szostak: And it will properly expose ENV, and everything?

config/jira.yml
michal-szostak: So `Rails.application.credentials.jira[:development]` will become `Rails.application.credentials.jira[Rails.env]`?

db/migrate/20180705112318_add_issue_id_to_orders.rb
michal-szostak: We store project, elsewhere, and JIRA client API's `find` method works on ids

db/migrate/20180705112813_add_issue_status_to_orders.rb
michal-szostak: Because I want all existing issues to have issue_status = 1, and new ones to have issue_status = 2 by default

lib/tasks/jira.rake
michal-szostak: if you say so 🆗

lib/tasks/jira.rake
michal-szostak: Well, I also wonder about that. I think @bwilk or @roksanaer are better persons to ask this.
Personally I would agree

spec/jobs/order/register_job_spec.rb
michal-szostak: It works, it checks whether exception is caught. But specific handling should be checked elsewhere...
Or not?

app/services/order/register.rb
mkasztelnik: `@order.jira_active!` is doing db update inline, next you are changing model update and once again save - so another db update.

spec/jobs/order/register_job_spec.rb
michal-szostak: I had some problem with expect in that case... But maybe if I put it inside of the `let` block...
I will check it out.

config/initializers/jira.rb
mkasztelnik: yeap, take a look into documentation https://apidock.com/rails/Rails/Application/config_for

config/jira.yml
mkasztelnik: yeap

spec/jobs/order/register_job_spec.rb
mkasztelnik: Then maybe you should check if this job is not rescheduled.

db/migrate/20180705112813_add_issue_status_to_orders.rb
mkasztelnik: If this is working as you describe than neat :smile: Please add a comment with the explanation because the intention is not obvious here.

spec/jobs/order/register_job_spec.rb
michal-szostak: I need to make sure that service `call` is called and that job handles error and not breaks. So expect is correct for this use case.

spec/services/order/register_question_spec.rb
michal-szostak: Well... if you look really closly those are two different variables - one has `s` at the end 😄

README.md
michal-szostak: Very good idea, let's do that

spec/jobs/order/register_question_job_spec.rb
michal-szostak: I understand the logic, but it will bloat code a little. Isn't there "the best of both worlds" solution?

spec/factories/order/questions.rb
michal-szostak: My understanding is that FactoryBot normally uses active record new? Or something like that. But `Order::Question` is a *virtual active record* which is created by using service. That's why I had to override `to_create` so that following block will be used...

But  I will say that I have no confidence in this part of the code, I don't really know how to handle `Order::Question` model,

app/services/order/register.rb
michal-szostak: So it is also saving model?
If that is the case, then sure it should be avoided. 
I have a question - how should I set value of an enum field without using numbers?

app/services/order/register_question.rb
michal-szostak: I left it this way so that setting of `@question` erros can be done without refactoring it again.

spec/jobs/order/register_question_job_spec.rb
mkasztelnik: Unfortunately not (at least into my knowledge). What is more this can be simplified if you don't need additional elements to check, e.g.:

```ruby
question = create(:order_change, author: nil)
```

spec/factories/order/questions.rb
mkasztelnik: I don't believe that it will work since `Order::Question::Create` require an order to work (it is not set by this factory).

app/services/order/register.rb
mkasztelnik:
```ruby
order.update_attributes(issue_id: issue.id, jira_status: :jira_active)
```
this also returns `true` on success DB save, so it should simplify this block of code a bit :)

bwilk: @michal-szostak I'm not sure if this has been already communicated - yes @order.jira_active! does save the model

bwilk: what @mkasztelnik is a better option though

app/services/order/register.rb
```diff
@@ -14,10 +21,19 @@ def call
...
+        @order.issue_id = issue.id
+        @order.save
+        true
+      else
+        @order.jira_errored!
```
bwilk: same here

lib/tasks/jira.rake
bwilk: I'd agree. Simply add condition on the environement type here so we could manipulate in dev and skip it in prod

README.md
bwilk: @michal-szostak would you take care of that in a separate PR?

spec/factories/order/questions.rb
michal-szostak: Instance already has order which is passed when `create` is called

README.md
michal-szostak: Sure, I'll do that

spec/factories/order/questions.rb
mkasztelnik: Can you point me into code line when this is set?

michal-szostak: A, I see - you want order to be created automatically if it's not provided, ok I'll add it

PR #130, opened 2018-07-13
122 style profile view

kosmidma:
Profile picture does not work. I understand that I have to use a webpckage but I do not know how. Mark's instructions seem too broad for me to read all of it. I just need to know how to get a picture and what path to write in a template or css.


app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -327,6 +336,36 @@ textarea:-ms-input-placeholder { /* IE 10+ */
...
+  border-color: $primary;
+}
+
+.profile-img {
+  background: url(../../assets/images/profile-img.png) top left no-repeat;
+  width: 192px;
+  height: 192px;
+}
+
+.fa-star, .fa-star-half-alt, .fa-angle-right{
```
michal-szostak: Setting color globally for specific glyphicon does not seem like a good pattern.
We should either use `span.fa.fa-star.rating` or some class on the rating stars wrapper.

app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -35,6 +35,14 @@ a {
   color: $body-color !important;
 }
 
+h1 {
+  font-size: 1.7rem;
+}
+
+h2 {
```
michal-szostak: Check whether `h1` and `h2` sizing can be changed via bootstrap variables

Issue #136, opened 2018-07-16
propagate the marketplace with categories


roksanaer:
Create 5 basic categories in the MP:
Compute
Platforms
Thematic services
Identity & Security
Professional Services

Issue #139, opened 2018-07-16
categorisation of the eosc-hub services


roksanaer:
connect created services with existing categories
connects with #138

Issue #138, opened 2018-07-16
Propagate the MP with eosc-hub services


roksanaer:
basing on https://marketplace.eosc-hub.eu and available documentation create entries for eosc-hub services

PR #140, opened 2018-07-18
[#136, #138, #139] Fill categories and services in database

goreck888:
Add categories and services to the marketplace from production.
Add loops for creating services, categories and service_categories.
Filling the database is done by parsing yaml.

Fixes: #136, #138, #139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Styling Order Placement view (@kosmidma)
 - Add dotenv gem (@michal-szostak)
 - Change checkin service (@wziajka)
+- Add filling database with services, categories and relations between them in seeds.rb (@goreck888)
 
 ### Changed
 - Upgrade Sprockets gem to avoid CVE-2018-3760 vulnerability (@mkasztelnik)
--- a/db/data.yml
+++ b/db/data.yml
@@ -1,0 +1,328 @@
+categories:
+
+    Compute:
+        name: &computeParent Compute
+        description: General computing services including VM management, containers and job processing
+    Platforms:
+        name: &platformsParent Platforms
+        description: Applications on Demand gives you access to online applications and application-development and hosting frameworks to support compute-intensive data analysis.
+    Thematic services:
+        name: &thematicServicesParent Thematic services
+        description: High-level services, often discipline-specific for research communities.
+    Identity:
+        name: &identityParent Identity & Security
+        description: Identity management, authorisation and other services to ensure trust.
+    Professional Services:
+        name: &professionalServicesParent Professional Services
+        description: Training, certification, consultancy and other non-technical services.
+
+services:
+
+    Cloud compute:
+        title: Cloud compute
+        parent: *computeParent
+        tagline: Run virtual machines on-demand with complete control over computing resources
+        description: "With Cloud Compute you can deploy and scale virtual machines on­demand. Cloud Compute offers guaranteed computational resources in a secure and isolated environment with standard API access, without the overhead of managing physical servers.
+
+                     Cloud Compute offers the possibility to select pre­configured virtual appliances (e.g. CPU, memory, disk, operating system or software) from a catalogue replicated across all EGI cloud providers.
+
+
+                     To access the service you need to:
+
+                         Request access selecting one of the service options below.
+                         Set the relevant parameters and add the service to the cart.
+                         Go to the cart and submit your order.
+                         You will receive an email summarising your order.
+                         You will be contacted by the EGI support team.
+
+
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -39,8 +39,10 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Conditional software version placeholder in frontend (@bwilk)
 - Service owner can manage owned services (@mkasztelnik)
 - Add rating for Service (@kmarszalek)
+- Add filling database with services, categories and relations between them in seeds.rb (@goreck888)
 - Add JIRA integration (marketplace to jira) (@michal-szostak)
 
+
```
mkasztelnik: Unnecessary new line - please remove it.

db/data.yml
```diff
@@ -0,0 +1,328 @@
...
+        name: &professionalServicesParent Professional Services
+        description: Training, certification, consultancy and other non-technical services.
+
+services:
+
+    Cloud compute:
+        title: Cloud compute
+        parent: *computeParent
+        tagline: Run virtual machines on-demand with complete control over computing resources
+        description: "With Cloud Compute you can deploy and scale virtual machines on­demand. Cloud Compute offers guaranteed computational resources in a secure and isolated environment with standard API access, without the overhead of managing physical servers.
```
mkasztelnik: It was agreed with @roksanaer that we will use markdown for service description formatting.

db/seeds.rb
```diff
@@ -9,13 +9,26 @@
...
 all_categories = []
-computing = Category.create_with(name: "Computing").find_or_create_by(name: "Computing")
-all_categories << computing
-all_categories << Category.create_with(name: "HPC", parent: computing).find_or_create_by(name: "HPC")
-all_categories << Category.create_with(name: "Cloud", parent: computing).find_or_create_by(name: "Cloud")
-all_categories << Category.create_with(name: "Data").find_or_create_by(name: "Data")
+
+yaml_hash["categories"].each do |category, hash|
+  if hash["parent"].nil?
+    all_categories << Category.create_with(name: hash["name"], description: hash["description"]).find_or_create_by(name: hash["name"])
```
mkasztelnik: This can be simplified into:

```ruby
parent = hash["parent"} && Category.find_by(name: hash["parent"])
Category.create!(name: hash["name"], description: hash["description"], parent: parent)
```

mkasztelnik: This also assumes that children are defined after parent. If such an assumption is your intend than it should be described in the readme.

db/seeds.rb
```diff
@@ -9,13 +9,26 @@
...
+  else
+    all_categories << Category.create_with(name: hash["name"], description: hash["description"], parent: Category.find_by(name: hash["parent"])).find_or_create_by(name: hash["name"])
+  end
+end
+
+puts "Generating services"
+all_services = []
+all_service_categories = []
+yaml_hash["services"].each do |service, hash|
+  all_services << Service.create_with(title: hash["title"], tagline: hash["tagline"], description: hash["description"]).find_or_create_by(title: hash["title"])
```
mkasztelnik: Can you explain why `create_with(...).find_or_create_by(...) is used?

db/seeds.rb
```diff
@@ -9,13 +9,26 @@
...
+  all_service_categories << ServiceCategory.create_with(service_id: Service.find_by(title: hash["title"]).id, category_id: Category.find_by(name: hash["parent"]).id, main: true).find_or_create_by(service_id: Service.find_by(title: hash["title"]).id, category_id: Category.find_by(name: hash["parent"]).id)
```
mkasztelnik: If you want to search several categories you can use:

```ruby
Category.where(id: [1,2,3,4])
```

db/seeds.rb
goreck888: I just used the same methods which were used before my changes.'

I know that without find_or_create_by(...) records don't save in the base. But even if they do, that will prevent duplicating records.

db/seeds.rb
goreck888: I know, but in yaml category is a string, it's easier to find category by its name than by id.
Maybe I'm wrong.

db/seeds.rb
goreck888: Ok, I'll correct it.

db/data.yml
goreck888: Ok, I have to know what is the formatting pattern and then I'll fill it.

db/seeds.rb
bwilk: Use create() instead

bwilk: When creating a service - assign category in Service model or in Category model instead of creating ServiceCategory bean itself

Issue #61, opened 2018-06-27
Add scss & haml validation to overcommit


michal-szostak:
Overcommit precommit hook should validate scss and haml files to check syntax and formatting.

mkasztelnik:
If this will be added into overcommit than it should also be added into travis build

PR #141, opened 2018-07-18
[#61] HAML & SCSS validation to overcommit

michal-szostak:
# What is in this PR

In this PR validation for scss (via `scss-lint`) and HAML (via `haml-lint`) has been 
added to overcommit.

These checks has been added to preCommit hook, which means that running
```
overcommit --run
```

Will call them, which means that no modification to `.travis.yml` is required.

Additionally all offending files have been fixed according to new rules.

# Sensible defaults

Almost all defaults for HAML linting have been left *as is*, except for some minor things like:

- changed maximum line length to 120 characters
- changed maximum file length to 150 lines
- enforced id to be placed before classes

As for SCSS more fine tuning has been done. I dropped more restrictive validators, as they would create more problems than benefits IMHO. The most important validators which are enabled are:

- enforcing trailing newline after `}`
- enforcing space before `{`
- enforcing space after `:`
- no inline color allowed
- proper usage of `:` with pseudo selectors and pseudo elements
- and couple of smaller ones.

Fixes: #61

### DIFF
--- a/.haml-lint.yml
+++ b/.haml-lint.yml
@@ -1,0 +1,12 @@
+linters:
+  ViewLength:
+    enabled: true
+    max: 150
+
+  LineLength:
+    enabled: true
+    max: 120
+
+  ClassesBeforeIds:
+    enabled: true
+    EnforcedStyle: 'id'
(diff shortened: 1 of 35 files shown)


michal-szostak:
It seems that no reviewer is available to double check this pull request. As the changes to the code base are minimal I am closing it.

PR #143, opened 2018-07-20
[#113] User can add open access services to my services

### LINKED ISSUE #113: As a registered user I am able to reach a service, that does not require ordering process (it's open to use), from the service detail view
https://v81k42.axshare.com/#g=1&p=service-detail_v3 view should not include Order Specification tab
instead of 'proceed to order' button there is 'Add to my services list' and 'Go to the service' ( @abacz should decide here).
Button 'Go to the service' should open a new tab with the URL connected with a service.
Button 'Add to my services list' are responsible for activate the open service, it should add it to 'My services' list as 'Active'. Service order detail is probably not needed in this case

From the UI point of view, on 'my services list' open services (which do not have the service order detail view) should be somehow distinguished

martaswiatkowska:
User can add to 'my service' open access service.
Open access service is available for all logged users.
User can go to open access service website from service details view 
If user add open access service to "my services" order will change status to ready(i dont know is this good status @roksanaer ???)

User can add open access service in backoffice.
If service is open access then connected url is required. 

Fix #113

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -17,6 +17,7 @@ class Service < ApplicationRecord
   validates :title, presence: true
   validates :description, presence: true
   validates :tagline, presence: true
+  validates :connected_url, presence: true, url: true, if: :open_access?
 
   after_save :set_first_category_as_main!, if: :main_category_missing?
 
--- a/app/services/order/create.rb
+++ b/app/services/order/create.rb
@@ -7,11 +7,17 @@ class Order::Create
 
   def call
     @order.created!
+    @service = Service.find_by(id: @order.service_id)
 
     if @order.save
       @order.new_change(status: :created, message: "Order created")
       OrderMailer.created(@order).deliver_later
-      Order::RegisterJob.perform_later(@order)
+
+      if !@service.open_access
+        Order::RegisterJob.perform_later(@order)
+      else
+        Order::ReadyJob.perform_later(@order)
+      end
     end
 
     @order
--- a/app/services/order/ready.rb
+++ b/app/services/order/ready.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class Order::Ready
+  def initialize(order)
+    @order = order
+  end
+
+  def call
+    update_status! &&
+    notify!
+  end
+
+  private
+
+    def update_status!
+      @order.new_change(status: :ready,
+                        message: "Your order is ready")
+    end
+
+    def notify!
+      OrderMailer.changed(@order).deliver_later
+    end
+end
--- a/app/views/services/_description.html.haml
+++ b/app/views/services/_description.html.haml
@@ -7,8 +7,10 @@
       - if policy(Order.new).create?
         = form_for(Order.new) do |f|
           = f.hidden_field :service_id, value: service.id
-          %button{ class: "btn btn-primary", type: "submit" } Order
-
+          %button{ class: "btn btn-primary" }
+            = write_button_text
+          - if @service.open_access?
+            = link_to "Go to the service", service.connected_url, class: "btn btn-primary", target: "_blank"
 
       %h3.mb-3.mt-3 Run virtual machines on-demand with complete control over computing resources.
       %p.mt-4 With Cloud Compute you can deploy and scale virtual machines on-demand. Cloud Compute offers guaranteed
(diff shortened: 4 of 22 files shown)


app/services/order/create.rb
```diff
@@ -7,11 +7,14 @@ def initialize(order)
...
   def call
     @order.created!
+    @service = Service.find_by(id: @order.service_id)
 
-    if @order.save
-      @order.new_change(status: :created, message: "Order created")
-      OrderMailer.created(@order).deliver_later
+    @order.new_change(status: :created, message: "Order created")
+    OrderMailer.created(@order).deliver_later
+    if @order.save && !@service.open_access
```
mkasztelnik: This condition looks weird: so what will happened when order is not saved but it is open_access service?

app/services/order/ready.rb
```diff
@@ -0,0 +1,24 @@
...
+    update_status! &&
+    notify!
+  end
+
+  private
+
+    def update_status!
+      @order.new_change(status: :ready,
+                        message: "Your order is ready")
+      true
```
mkasztelnik: This `true` is not needed since `new_change` returns true when the new change is created successfully.

app/models/service.rb
```diff
@@ -17,6 +17,7 @@ class Service < ApplicationRecord
   validates :title, presence: true
   validates :description, presence: true
   validates :tagline, presence: true
+  validates :connected_url, presence: true, url: true
```
mkasztelnik: Maybe simply `service_url` or `url`?

app/views/services/_description.html.haml
```diff
@@ -7,8 +7,13 @@
       - if policy(Order.new).create?
         = form_for(Order.new) do |f|
           = f.hidden_field :service_id, value: service.id
-          %button{ class: "btn btn-primary", type: "submit" } Order
-
+          %button{ class: "btn btn-primary" }
+            - if @service.open_access?
+              Add to my services
+            - else
+              Order
```
mkasztelnik: This is else should be moved into helper method.

spec/factories/orders.rb
```diff
@@ -5,5 +5,9 @@
     status :created
     user
     service
+
+    factory :open_access_order do
```
mkasztelnik: Can you explain why this factory is needed?

spec/factories/services.rb
```diff
@@ -6,5 +6,11 @@
     sequence(:description) { |n| "service #{n} description" }
     sequence(:terms_of_use) { |n| "service #{n} terms of use" }
     sequence(:tagline) { |n| "service #{n} tagline" }
+    sequence(:connected_url) { "https://sample.url" }
```
mkasztelnik: This URL will be used for all services? If not maybe it should be moved into `open_access_service` factory?

spec/features/orders_spec.rb
```diff
@@ -10,6 +10,8 @@
 
     let(:user) { create(:user) }
     let(:service) { create(:service) }
+    let(:open_access_service) { create(:open_access_service) }
+    let(:open_access_order) { create(:open_access_order, user: user, service: open_access_service) }
```
mkasztelnik: Both variables are used only in several tests in other tests these variables are not needed. My proposition is to use simple local variables instead of creating global variables.

app/models/service.rb
martaswiatkowska: I think that service_url is a bit misleading with rails routing. In my opininion connected_url is is more appropriate because it's url to external web page

app/services/order/create.rb
martaswiatkowska: Nothing should happen. I made mistake here

spec/factories/orders.rb
martaswiatkowska: It's not need, sory.

spec/factories/services.rb
martaswiatkowska: if open_acces is true then services must have this url but if false then does not have to. I will do validation for that

roksanaer:
@martaswiatkowska 
"If a user adds open access service to "my services" order will change status to ready(i don't know is this good status @roksanaer ???)" - to discuss on Tuesday 24th.
In my opinion, there should be an order status and a service status separately. After adding a service to 'My services' list, there should be an order created with status indicating that it was handled positively (e.g. accepted, at the moment I do not know the schema - apologies) and a service status should be set to e.g. active. But this might be a separated point of view, it should be discussed within broader context (one of the topics for Tuesday)

PR #147, opened 2018-07-25
[#52] JIRA Webhooks

### LINKED ISSUE #52: Webhook integration with JIRA
Marketplace should be able to listen for changes in tickets in the connected provider's JIRA.
If updated issue is connected to the marketplace order, the order should update as well.

michal-szostak:
# What is in this PR
This PR contains JIRA integration from JIRA instance to the the marketplace.
This functionality is chiefly realized by `app/controllers/rest/webhooks/jiras_controller.rb` controller.

It receives notifications from jira and updates order if changed issue has been connected to it.

Additionally a lot of refactoring has been done concerning jira rake task. Everything has been placed in separated classes and tested. Basic groundwork for testing rake tasks has also been added.

# What should I do to check whether it works
Provided you've set up your local JIRA and started marketplace with correct settings (see `README.md`) do the following:
1. Log into MP
2. Go to any service
3. Place order
4. Go to jira to the issue connected to created order
5. Change ticket status
6. In MP, on order details page new order change should be visible
7. Add question to the order
8. Go back to jira, new comment should have been created

# How can I test it?
Unfortunatelly automatic branch deployment will not help you with this.
The only way to test it is to start your own jira instance on your development machine. 
But do not panic, it is much easier then you may think (generally needs only a couple of lines in the terminal to setup atlassian sdk).

`README.md` has all the information needed for connecting local JIRA instance with marketplace

Fixed #52


lib/jira/setup.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def call
+      @client.mp_project
+      puts "Project #{@client.jira_project_key} already exists"
+    rescue JIRA::HTTPError => e
+      if e.response.code == "404"
+        p = @client.Project.build
+        unless p.save(key: @client.jira_project_key,
+                      name: @client.jira_project_key,
+                      projectTemplateKey: "com.atlassian.jira-core-project-templates:jira-core-project-management",
```
bwilk: Maybe this should be configurable instead of hardcoded. This would allow for changes without re-release - just a suggestion I'm not insisting

spec/lib/jira/setup_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  it "call should check project existence" do
+    expect(jira_client).to receive(:mp_project).and_return(true)
+    expect { setup.call }.to output("Project #{jira_project_key} already exists\n").to_stdout
+  end
+
+  it "call should pass proper arguments to project.save" do
+    expect(project).to receive(:save).with(key: jira_client.jira_project_key,
+                                           name: jira_client.jira_project_key,
+                                           projectTemplateKey: "com.atlassian.jira-core-project-templates:jira-core-project-management",
```
bwilk: See above in comment for Jira::Setup

app/models/jira/checker.rb
```diff
@@ -0,0 +1,171 @@
...
+  block_error_handling :check_issue_type!
+  block_error_handling :check_project!
+  block_error_handling :check_create_issue!
+  block_error_handling :check_update_issue!
+  block_error_handling :check_add_comment!
+  block_error_handling :check_delete_issue!
+  block_error_handling :check_workflow!
+  block_error_handling :check_webhook!
+
+  def check_connection!
```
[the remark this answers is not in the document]
bwilk: Ok. I was wrong - sorry. You don't replace them. You declare new ones

app/controllers/rest/webhooks/jiras_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  rescue_from(ActiveRecord::RecordNotFound) do
+    render json: { message: "No correlating order found" }
+  end
+
+  def initialize
+    @jira_client = Jira::Client.new
+  end
+
+  def create
+    unless request.filtered_parameters.fetch("secret", "") == @jira_client.webhook_secret
```
mkasztelnik: Authorization can be moved into `before_action` block.

app/controllers/rest/webhooks/jiras_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      render json: { message: "Secret does not match" }, status: :bad_request
+    else
+      issue_id = request.filtered_parameters["issue"]["id"]
+
+      order = Order.find_by!(issue_id: issue_id)
+
+      request.filtered_parameters["changelog"].fetch("items", []).each do |change|
```
mkasztelnik: A lot of logic is here, maybe you could move it into service?

app/models/jira/checker.rb
```diff
@@ -0,0 +1,171 @@
+# frozen_string_literal: true
+
+class Class
+  def block_error_handling(*attrs)
```
mkasztelnik: It is quite dangerous to add such a method into base classes. I know that `Jira::Checker` is loaded only in the rake task but still this is quite dangerous.

mkasztelnik: Instead, you can create a basic class and use inheritance or include (as a concern) to inject this behavior.

config/routes.rb
```diff
@@ -30,6 +30,12 @@
     resources :services
   end
 
+  namespace :rest do
```
mkasztelnik: Maybe `api` is a better name?

lib/jira/console_checker.rb
```diff
@@ -0,0 +1,118 @@
...
+      abort("jira:check exited with errors")
+    end
+
+    def error_and_abort!(e, indent = 1)
+      print " FAIL".red << "\n"
+
+      case e
+      when Errno::ECONNREFUSED
+        puts "ERROR".red + ": Could not connect to JIRA: #{ @checker.client.jira_config["url"] }"
+        self.abort!
```
mkasztelnik: is `self` needed here?

michal-szostak: I suppose not, I like to have clear indication that it's a instance method and not some block or something else... But if that is not ruby convention I can cut it.

app/controllers/rest/webhooks/jiras_controller.rb
michal-szostak: I think that it hasn't grown that big yet.
In the near future there will be more logic added there (mostly connected to comments)
I think that at that time refactoring this out into service will be a good idea.

app/models/jira/checker.rb
michal-szostak: I tried to do something like that and it wouldn't work (trust me, overloading `Class` was the last thing I tried, but all the examples found in the *internets* point me to do it this way.

If you can provide me with a working example how to do this without extendig `Class` I would be very gratefull.

config/routes.rb
michal-szostak: Ok, I can change it, additionally it will be a good test to check whether I implemented named routes properly

PR #155, opened 2018-07-31
[#133] User can deactivate open_access service

### LINKED ISSUE #133: As a registered user I am able to reach an open service, that does not require ordering process and is already activated for me, from the service detail view
if the user is reaching the service via EOSC-Portal not for the first time (clicking the button for the first time), the button states: "Go to the service". Clicking the button:
- opens a new tab with the given service

A ‘Go to the service’ button is visible as well that removes the service from ‘My services’ list
If a user clicks on 'Stop using a service':
- the service is removed from 'My services' list

The detailed view of the service has 'Add to my services list' button again and ‘stop using the service' button is not available anymore

martaswiatkowska:
User can stop using open_access service by clicking on "Stop using the service" button. 
Button "Stop using the service" is displayed on details view if open_access service is activated (status: ready) 

Fix #133


app/models/order.rb
```diff
@@ -27,8 +28,8 @@ class Order < ApplicationRecord
   validates :user, presence: true
   validates :status, presence: true
 
-  def active?
-    !(ready? || rejected?)
+  def during_preparation?
```
bwilk: Are you sure about that ??? I mean the name?

martaswiatkowska: It's not perfect but I have no idea how to name this function. Maybe better in_preparation?

mkasztelnik:
draft?

Marek Kasztelnik

pon., 27 sie 2018, 09:46 użytkownik Marta Świątkowska <
notifications@github.com> napisał:

app/models/order.rb
mkasztelnik: maybe `draft?`

CHANGELOG.md
```diff
@@ -48,6 +48,9 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Add and styling static elements to sidebar (@kosmidma)
 - Update service view (@kosmidma)
 - Styling My Services view (@kosmidma)
+- Add and styling static elements to sidebar (@kosmidma)
+- AStyling My Services view (@kosmidma)
```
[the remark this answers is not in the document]
martaswiatkowska: But are on master. I did commit squash

app/helpers/service_helper.rb
```diff
@@ -21,4 +21,10 @@ def print_rating_stars(rating)
   def write_button_text
     @service.open_access ? "Add to my services" : "Order"
   end
+
+  def active?(service)
+    # user can have only one open_access order with status ready
+    Order.find_by(service: service.id, user: current_user, status: Order.statuses["ready"]) &&
+      service.open_access
```
kmarszalek: change condition order

app/services/order/update.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class Order::Update
+  def initialize(order)
+    @order = order
+  end
+
+  def call
+    if !(@order.deactivated? || @order.rejected?)
```
kmarszalek: ?

martaswiatkowska: This condition is needed because its not possible to deactivate already deactivated or rejected order.
DeactivateJob is called because in this pull request only deactivation is possible in Order::update.
Yes, should be called OrderMailer.updated

spec/jobs/order/deactivate_job_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+require "rails_helper"
+
+RSpec.describe Order::DeactivateJob do
+  let(:user) { create(:user) }
+  let(:deactivate_service) { instance_double(Order::Deactivate) }
+  let(:order) {
+    order = create(:open_access_order, user: user)
+    expect(Order::Deactivate).to receive(:new).
+        with(order).and_return(deactivate_service)
+    next order
```
kmarszalek: can you explain this?

app/controllers/orders_controller.rb
```diff
@@ -27,6 +27,21 @@ def create
     end
   end
 
+  def update
+    order = Order.find_by(service_id: params[:id], user_id: current_user.id,
```
mkasztelnik: `user: current_user`

app/controllers/orders_controller.rb
```diff
@@ -27,6 +27,21 @@ def create
...
+                          status: Order.statuses["ready"])
```
mkasztelnik: `status: :ready`

app/controllers/orders_controller.rb
```diff
@@ -27,6 +27,21 @@ def create
...
+
+    authorize(order)
+    Order::Update.new(order).call
+
+    if order.persisted?
```
mkasztelnik: It will always be persisted because you are upgrading order not creating a new one :) You should check if object if valid (`order.valid?`).

app/helpers/service_helper.rb
```diff
@@ -22,6 +22,13 @@ def write_button_text
     @service.open_access ? "Add to my services" : "Order"
   end
 
+  def active?(service)
+    # user can have only one open_access order with status ready
+    service.open_access && Order.find_by(service: service.id,
+                                         user: current_user,
+                                         status: Order.statuses["ready"])
```
mkasztelnik: `status: :ready`

app/services/order/update.rb
mkasztelnik: What if order will be deactivated or rejected after calling `call` method but before delayed job is executed?

app/views/services/_description.html.haml
```diff
@@ -7,9 +7,14 @@
       - if policy(Order.new).create?
         = form_for(Order.new) do |f|
           = f.hidden_field :service_id, value: service.id
-          %button{ class: "btn btn-primary" }
-            = write_button_text
-          - if @service.open_access?
+          - if !active?(service)
```
mkasztelnik: Maybe `active?` method can be moved into policy?

spec/features/orders_spec.rb
```diff
@@ -20,10 +20,24 @@
     end
 
     scenario "I see order open acces service button" do
-      @open_access_service = create(:open_access_service)
-      visit service_path(@open_access_service)
+      open_access_service = create(:open_access_service)
```
mkasztelnik: This test is doing much more than just test if there is an open access service button.

spec/features/orders_spec.rb
```diff
@@ -20,10 +20,24 @@
     end
 
     scenario "I see order open acces service button" do
```
mkasztelnik: `access`

spec/features/orders_spec.rb
```diff
@@ -20,10 +20,24 @@
...
+      open_access_service = create(:open_access_service)
+
+      visit service_path(open_access_service)
 
       expect(page).to have_text("Add to my services")
+      expect(page).to_not have_text("Stop using the service")
+      expect(page).to have_text("Go to the service")
+    end
+
+    scenario "I see deactivate service button" do
```
mkasztelnik: I propose to improve it a bit into something like "I can deactivate active open access service"

spec/features/orders_spec.rb
```diff
@@ -38,19 +52,18 @@
...
-    scenario "I can order open acces service" do
-      @open_access_service = create(:open_access_service)
-      @open_access_order = create(:order, service: @open_access_service, user: user)
+    scenario "I can order open access service" do
+      open_access_service = create(:open_access_service)
 
-      visit service_path(@open_access_service)
+      visit service_path(open_access_service)
 
       expect { click_button "Add to my services" }.
```
mkasztelnik: IMHO this test is enough, "I see order open access service button" is not needed because it is included in this test (you are clicking required link)

spec/jobs/order/deactivate_job_spec.rb
mkasztelnik: This is not the right syntax. `next` can be used in for loops to skip current execution and start new iteration. In ruby the last line is used to return value. So you can convert it into:

```ruby
order = create(:open_access_order, user: user)
expect(Order::Deactivate).to receive(:new).
￼         with(order).and_return(deactivate_service)
order
```
or even better you can use magic rails `tap` method:
```ruby
create(:open_access_order, user: user).tap do |order|
  expect(Order::Deactivate).to receive(:new).
￼         with(order).and_return(deactivate_service)
end
```

What I really don't like in this code is that `let` is meant to be used to create variables with complex initialization which is generic enough to be reused in many tests. Here you are adding an expectation. So now this is not generic order (it is really open access service order not normal order - also naming can be improved here) but the order which is expected to be deactivated. Such expectation is a part of test and should not be a part of variable creation.

spec/services/order/deactivate_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it "changes order status into deactivate on success" do
+    described_class.new(order).call
+
+    expect(order).to be_deactivated
+  end
+
+  it "sent email to order owner" do
+    order.new_change(status: :deactivated, message: "Order has been deactivated")
```
mkasztelnik: Why this line is needed? According to previous tests `Order::Deactivate` should create this change.

spec/services/order/deactivate_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    expect { described_class.new(order).call }.
+      to change { ActionMailer::Base.deliveries.count }.by(1)
```
mkasztelnik: Maybe you should check what kind of email is sent do the user?

spec/services/order/update_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Order::Update do
+  let(:user) { create(:user) }
+  let(:service) { create(:open_access_service) }
```
mkasztelnik: You are creating a special service here, maybe the name should reflect this specialty?

spec/services/order/update_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it "triggers deactivate order in external system" do
```
mkasztelnik: What do you mean by the external system here?

spec/services/order/deactivate_spec.rb
martaswiatkowska: let define a memoized helper method that is called in a `before` hook. In my let I create an order so in every test I work on another order.

https://relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let

app/views/services/_description.html.haml
martaswiatkowska: But in this method I check if service is open_access and  if there is already an order for that service with status :ready. I think it's not necessary.

spec/services/order/update_spec.rb
martaswiatkowska: Jira. I get this code from @michal-szostak

martaswiatkowska: Jira, I took this code from @michal-szostak

mkasztelnik: The problem is that I don't see any verification in the test that "external system" is notified. You are only checking that `Order::DeactivateJob` job was created.

app/services/order/update.rb
martaswiatkowska: You mean the situation when user click several times button "stop using service" before delayed job is executed?

app/helpers/service_helper.rb
```diff
@@ -21,4 +21,9 @@ def print_rating_stars(rating)
   def write_button_text
     @service.open_access ? "Add to my services" : "Order"
   end
+
+  def is_active?(service)
+    # user can have only one open_access order with status ready
```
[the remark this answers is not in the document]
martaswiatkowska: service is open access not order

app/controllers/orders_controller.rb
```diff
@@ -19,14 +19,29 @@ def create
     authorize(order_template)
     order = Order::Create.new(order_template).call
 
-    if order.persisted?
+    if order.valid?
```
mkasztelnik: In create you can use `persisted?` because new record is created.

app/controllers/orders_controller.rb
```diff
@@ -19,14 +19,29 @@ def create
...
       redirect_to order_path(order)
     else
       redirect_to service_path(order.service),
                   alert: "Unable to create order"
     end
   end
 
+  def update
+    order = Order.find_by(service_id: params[:id], user: current_user.id,
```
mkasztelnik: Maybe bang method should be used here (`find_by!`)?

app/views/services/_description.html.haml
```diff
@@ -7,9 +7,14 @@
...
+            %button{ class: "btn btn-primary" }
+              = write_button_text
```
mkasztelnik: This is quite a generic name for the helper method. Try to use a more specialized name.

app/views/services/_description.html.haml
mkasztelnik: The name is very generic. It can lead into misunderstanding like in my previous comment, where I was not sure what this method is doing.

app/helpers/service_helper.rb
```diff
@@ -22,6 +22,13 @@ def write_button_text
     @service.open_access ? "Add to my services" : "Order"
   end
 
+  def active?(service)
+    # user can have only one open_access service order with status ready
+    service.open_access && Order.find_by(service: service.id,
```
mkasztelnik: It is not recommended to add db queries in the helper method. If something like this is required maybe it can be moved into the controller?

app/views/services/_description.html.haml
mkasztelnik: Also try to not use negation. This `if` condition can be converted into:

```haml
- unless active?
```

or you can change order of the blocks (so if active? than random active specific elementsl else not active specific elements)

spec/features/orders_spec.rb
```diff
@@ -116,6 +98,38 @@
 
       expect(page).to have_text("Question cannot be blank")
     end
+
+    context "when service is open_access" do
```
mkasztelnik: maybe simply "when open access service"?

spec/features/orders_spec.rb
```diff
@@ -116,6 +98,38 @@
...
+      let(:open_access_service) { create(:open_access_service) }
+
+      scenario "I cannot see 'Stop using the service' button" do
```
mkasztelnik: Maybe "I cannot stop using not ordered service"?

app/services/order/update.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Order::Update
```
mkasztelnik: This service name is very misleading. It is not updating order (updating is a very generic concept) but deactivating service. I propose to remove this service and invoke directly deactivate job in the controller.

app/views/services/_description.html.haml
```diff
@@ -7,9 +7,14 @@
...
+          - else
+            = link_to "Stop using the service", order_path(service),
```
mkasztelnik: Please consider better url naming. Right now we have:

```
PUT /orders/{id}
```

but the `id` is not an order id but `Service` `id`. This is very misleading. Maybe better path would be:

```
PATH /services/{id}/deactivate
```
?

app/controllers/orders_controller.rb
martaswiatkowska: "It will always be persisted because you are upgrading order not creating a new one :) You should check if object if valid (order.valid?)."
your comment above ;D

app/services/order/deactivate.rb
```diff
@@ -0,0 +1,23 @@
...
+  def call
+    update_status! &&
+    notify!
+  end
+
+  private
+
+    def update_status!
+      @order.new_change(status: :deactivated,
+                        message: "Your order has beed deactivated")
```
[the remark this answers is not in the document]
martaswiatkowska: Order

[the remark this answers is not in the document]
mkasztelnik: Looks strange for me. I'm using service not order so deactivation should be connected with word service not order IMHO.

app/views/services/_description.html.haml
martaswiatkowska: So it will be "generate_button_text" :blush:

app/models/order.rb
```diff
@@ -28,8 +29,8 @@ class Order < ApplicationRecord
   validates :user, presence: true
   validates :status, presence: true
 
-  def active?
-    !(ready? || rejected?)
+  def draft?
+    !(ready? || rejected? || deactivated?)
```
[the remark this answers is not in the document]
martaswiatkowska: No, he cannot ask when the order is deactivated

app/views/services/_description.html.haml
mkasztelnik: This is still too generic. You are using helper which creates global helper methods (which can be used on other views). I propose to change it into:

```ruby
def order_service_text(service)
  service .open_access ? "Add to my services" : "Order"
end
```
This creates a much more explicit method invocation (`order_service_text(my_service)`) - no hidden state (`@service`). 

Avoiding hidden state is a general recommendation to all helper methods.

app/controllers/orders_controller.rb
mkasztelnik: This is valid for `update` method not for `create` :smile:

app/views/services/_description.html.haml
martaswiatkowska: But order model has all statuses

mkasztelnik:
@roksanaer is it valid anymore since many open access services can be added into different projects?

roksanaer:
to remember:
after introducing projects, open access service should be activated (and deactivated) in the scope of the project.
For now, before finishing the critical path, we can either:
correlate free to access services with the project concept
leave it until we finish the critical path.

Even if we decide to change it now, I would probably merge this pull request, and start enhancements after.

@martaswiatkowska, @kmarszalek, @mkasztelnik what do you think?

roksanaer:
sorry @mkasztelnik didnt see your comment before.
Did you add the functionality on adding open access service to the project?

app/services/order/deactivate.rb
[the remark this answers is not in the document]
roksanaer: true, it can be even changed to: "This service is not active anymore. You can activate it at any time from the service entry page"

after adding projects: "This service has been removed from your project (can we use here the name of the project as a $?). If you added it to other projects, you can still use it. Otherway, you can always re-activate it on the service entry page"

Issue #106, opened 2018-07-09
As a registered user I am able to rate a service so other users have a feedback about its value


roksanaer:
After few days (2-3, can be changed at any moment) after a service has been accepted,  a mail is sent to the customer (see rails active jobs: http://guides.rubyonrails.org/active_job_basics.html, delay jobs) informing about the possibility to rate the service. Link in the email should point to the connected service-order-detail view (https://v81k42.axshare.com/#g=1&p=my-services-detail).
Under the service order detail on the right, an element allowing to rate the service appears (e.g. 5 empty stars, customer can mark a given number of start by clicking on one of them (like in Skype for example), after clicking on one of the start a button appears allowing to submit the rate). Clicking on the button submits the rate. Customer is shown how many stars he has admitted to the service and under a sentence:
"thank you for giving us your feedback"

I would also allow the customer to change the evaluation. To the right of the stars there should be an icon. Putting the cursor over it gives the info: "what to change your rate? click here". Clicking the icon 'unblocks' the stars, customer can click again on one of the 5 stars, changing his rate and save it.

Submitting new grade, removes the old one from the system, the average rate of the service is re-calculated.

This tasks requires to configure the sidekiq (https://github.com/mperham/sidekiq/wiki/Getting-Started). Consultancy for this: @bwilk

kmarszalek:
Od Marka:
Some finding to investigate:
  * https://github.com/merit-gem/merit this is more reputation, badges engine, but we take a look at is during research (there is rankings support https://github.com/merit-gem/merit#rankings)
  * Simple example how to implement ranking in rails na postgresql using `ntile` window function http://naturaily.com/blog/post/ruby-on-rails-implementation-of-a-ranking-system-using-postgresql-window-functions

must ready by person implementing this issue

More advanced ranking with positives and negatives: http://www.akitaonrails.com/2016/10/31/ruby-on-rails-implementation-of-a-proper-ranking-popularity-system - very good read about good and bad ranking implementation.

abacz:
I have different proposition.
Rating in service detail in my services
https://v81k42.axshare.com/#g=1&p=my-services-detail-_active

Additionally where you are visiting service page and you have it, link after rate will appear.
https://v81k42.axshare.com/#g=1&p=service-detail-active

For now is enough. But at all we will have dedicated page for rating,

mkasztelnik:
Some inspiration from amazon:
[...]

PR #164, opened 2018-08-21
#106 Add service-opinion with submit form for order

kmarszalek:


app/views/orders/service_opinions/_form.html.haml
```diff
@@ -1,14 +1,14 @@
...
-  %i.fas.fa-star-o#first-rating{onClick: 'changeClasses(1);'}
-  %i.fas.fa-star-o#second-rating{onClick: 'changeClasses(2);'}
-  %i.fas.fa-star-o#third-rating{onClick: 'changeClasses(3);'}
-  %i.fas.fa-star-o#fourth-rating{onClick: 'changeClasses(4);'}
-  %i.fas.fa-star-o#fifth-rating{onClick: 'changeClasses(5);'}
+  %i#first-rating.fas.fa-star-o{ onClick: "changeClasses(1);" }
+  %i#second-rating.fas.fa-star-o{ onClick: "changeClasses(2);" }
+  %i#third-rating.fas.fa-star-o{ onClick: "changeClasses(3);" }
+  %i#fourth-rating.fas.fa-star-o{ onClick: "changeClasses(4);" }
+  %i#fifth-rating.fas.fa-star-o{ onClick: "changeClasses(5);" }
```
mkasztelnik: You can use stimulus for this

app/controllers/services_controller.rb
```diff
@@ -10,5 +10,7 @@ def index
 
   def show
     @service = Service.find(params[:id])
+    @service_opinions = ServiceOpinion.joins(:order).where(orders: { service: @service })
+    @service_opinions_size = @service_opinions.size
```
mkasztelnik: You can use counter cache to avoid additional db query (https://rubyplus.com/articles/3221-Counter-Cache-Column-in-Rails-5)

app/javascript/packs/application.js
```diff
@@ -38,5 +38,46 @@ document.addEventListener("turbolinks:before-render", function(event) {
  */
 document.addEventListener('DOMContentLoaded', function () {
   dom.i2svg();
+  $("#first-rating").on('click', function () {
+      document.getElementById("service_opinion_rating").value = 1;
```
mkasztelnik: Maybe something like (in pseudocode):

```
val = document.getElementById("service_opinion_rating").value
ratingContainer.querySelectorAll('raiting).each_with_index { |el, i| el.attr('data-prefix', i < val ? 'fas' : 'far') }
```

app/models/order.rb
```diff
@@ -16,12 +16,16 @@ class Order < ApplicationRecord
       jira_errored: 3
   }
 
+  RATE_PERIOD = 5.days.ago
```
mkasztelnik: Maybe this should be moved into configuration?

app/models/order.rb
```diff
@@ -16,12 +16,16 @@ class Order < ApplicationRecord
...
+  RATE_AFTER_PERIOD = 5.days
```
mkasztelnik: Similar as above

app/models/order.rb
```diff
@@ -45,4 +49,12 @@ def new_change(status: nil, message: nil, author: nil)
   def to_s
     "##{id}"
   end
+
+  def show_rating_button?
```
mkasztelnik: This is view not model specific - move into view helper.

app/models/service_opinion.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ServiceOpinion < ApplicationRecord
+  belongs_to :order
+
+  validates :rating, presence: true, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than_or_equal_to: 5 }
+  after_save :update_service_rating
+  private
+
+    def update_service_rating
```
mkasztelnik: Looks like a logic - maybe should be moved into service?

app/policies/service_opinion_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def index?
+    user
+  end
+
+  def show?
+    owner?
+  end
+  def create?
```
mkasztelnik: Enter is missing before this method.

app/policies/service_opinion_policy.rb
```diff
@@ -0,0 +1,28 @@
...
+    user
+  end
+
+  def permitted_attributes
+    [:rating, :opinion]
+  end
+  private
```
mkasztelnik: Enter is missing before this method

app/services/order/ready.rb
```diff
@@ -19,5 +19,6 @@ def update_status!
 
     def notify!
       OrderMailer.changed(@order).deliver_later
+      Order::RatingMailerJob.set(wait: Order::RATE_AFTER_PERIOD).perform_later(@order)
```
mkasztelnik: `deliver_later` has options to set when email should be sent. Take a look here: https://api.rubyonrails.org/classes/ActionMailer/MessageDelivery.html:

```ruby
OrderMailer.rating(...).deliver_later(wait: ...)
```

IMHO `Order::RatingMailerJob` is not necessary

db/migrate/20180705121340_add_tagline_to_service.rb
```diff
@@ -1,5 +1,6 @@
 class AddTaglineToService < ActiveRecord::Migration[5.2]
   def change
-    add_column :services, :tagline, :text, null: false
+    add_column :services, :tagline, :text
```
mkasztelnik: How is this change connected to this PR?

db/migrate/20180716124933_create_service_opinions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateServiceOpinions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :service_opinions do |t|
+      t.integer :rating, null: false
+      t.text :opinion
+      t.timestamp :created_at, null: false
```
mkasztelnik: Use `t.timestamps`.

db/schema.rb
```diff
@@ -78,6 +78,14 @@
     t.index ["service_id"], name: "index_service_categories_on_service_id"
   end
 
+  create_table "service_opinions", force: :cascade do |t|
+    t.integer "rating", null: false
+    t.text "opinion"
+    t.datetime "created_at", null: false
+    t.bigint "order_id"
```
mkasztelnik: Heh I see you are trying to fight with rails. Take a look at https://guides.rubyonrails.org/association_basics.html and migration definition for relations, eg.:

```ruby
t.belongs_to :author, index: true
```

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    service_opinion = create(:service_opinion, order: order, rating: "3")
+    service_opinion.save
+
+    expect(service.rating).to eq(3.0)
+  end
+
+  it "service_opinion sends after_save callback" do
+    service_opinion = create(:service_opinion, rating: "3")
+    expect(service_opinion).to receive(:update_service_rating)
+    service_opinion.save
```
mkasztelnik: This save will do nothing since no changes are done before last save (in line 29) - how this work exactly?

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  it "service_opinion updates service.rating" do
+    service = create(:service)
+    expect(service.rating).to eq(0.0)
+
+    order = create(:order, service: service)
+
+    service_opinion = create(:service_opinion, order: order, rating: "3")
+    service_opinion.save
```
mkasztelnik: factory bod `create` is saving entry in the db, thus this line is not needed.

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+RSpec.describe ServiceOpinion, type: :model do
+  it { should validate_presence_of(:rating) }
+  it { should validate_numericality_of(:rating) }
+  it { should belong_to(:order) }
+
+  it "has rating" do
+    service_opinion = create(:service_opinion, rating: "3")
+
+    expect(service_opinion.rating).to eq(3)
```
mkasztelnik: Do we really need to check db columns? Seems like dump test

spec/models/order_spec.rb
```diff
@@ -63,4 +63,37 @@
...
+      service_opinion = create(:service_opinion, rating: "3", order: order)
+      service_opinion.save
+      expect(order).to_not be_show_rating_button
+    end
+
+    it "is true when order is ready and updated_at before RATE_PERIOD and no service_opinion" do
+      order = create(:order, status: :created)
+      order.new_change(status: :ready, message: "Order ready")
+      order_change = order.order_changes.last
+      order_change.created_at = Order::RATE_PERIOD - 1.day
```
mkasztelnik: You can use time travel (https://api.rubyonrails.org/v5.1/classes/ActiveSupport/Testing/TimeHelpers.html) capability to move into the future to test this behavior (no db update will be needed than).

spec/models/order_spec.rb
```diff
@@ -63,4 +63,37 @@
...
+    it "is false when order is ready but updated_at after RATE_PERIOD" do
+      order = create(:order, status: :created)
+      order.new_change(status: :ready, message: "Order ready")
+      expect(order).to_not be_show_rating_button
+    end
+
+    it "is false when order is ready and updated_at before RATE_PERIOD but there is service_opinion" do
+      order = create(:order, status: :created)
+      order.new_change(status: :ready, message: "Order ready")
+      order.order_changes.find_by(status: :ready).created_at = Order::RATE_PERIOD - 1.day
```
mkasztelnik: You can use time travel (https://api.rubyonrails.org/v5.1/classes/ActiveSupport/Testing/TimeHelpers.html) capability to move into the future to test this behavior (no db update will be needed than).

spec/jobs/order/rating_mailer_job.rb
```diff
@@ -0,0 +1,19 @@
...
+require "rails_helper"
+
+RSpec.describe Order::RatingMailerJob do
+  let(:order_owner) { create(:user) }
+  let(:ready_service) { instance_double(Order::Ready) }
+  let(:order) {
+    order = create(:order, user: order_owner)
+    expect(Order::Ready).to receive(:new).
+        with(order).and_return(ready_service)
+    next order
```
mkasztelnik: Can you explain this line?

spec/jobs/order/rating_mailer_job.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Order::RatingMailerJob do
+  let(:order_owner) { create(:user) }
+  let(:ready_service) { instance_double(Order::Ready) }
+  let(:order) {
+    order = create(:order, user: order_owner)
+    expect(Order::Ready).to receive(:new).
```
mkasztelnik: It is strange to define expectation in the variable. Something like this should be done in a concrete test.

CHANGELOG.md
```diff
@@ -54,6 +54,8 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Change checkin service (@wziajka)
 - Add filling database with services, categories and relations between them in seeds.rb (@goreck888)
 - Styling Order Details view (@kosmidma)
+- Add service opinion form that shows 5 days after order is ready (@kmarszalek)
+- Add active job that send email to user about rating service 5 days after order is ready (@kmarszalek)
```
mkasztelnik: Seems like this is a part of the feature described in the previous line.

app/javascript/packs/application.js
kmarszalek: that was my intention, wanted to push PR and then work on this js ;)

CHANGELOG.md
kmarszalek: didnt think it through

db/migrate/20180705121340_add_tagline_to_service.rb
kmarszalek: it's not... not sure how it got here.

app/controllers/orders/service_opinions_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  before_action :authenticate_user!
+
+  def new
+    @order = Order.find(params[:order_id])
+    @service_opinion = ServiceOpinion.new(order: @order)
+    authorize(@service_opinion)
+  end
+
+  def create
+    @order = Order.find(params[:order_id])
```
mkasztelnik: joins service to avoid second db query in next line

kmarszalek:
i will squash commits after Marek's check

app/controllers/orders/service_opinions_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      redirect_to order_path(@order), notice: "Rating submitted sucessfully"
+    else
+      render :new, status: :bad_request
+    end
+  end
+
+
+  private
+    def service_opinion_template
+      ServiceOpinion.new(permitted_attributes(ServiceOpinion).merge(order: @order))
```
mkasztelnik: This method assumes that order is defined. This is ok when it is guaranteed. To make this guarantee I propose to move order find into `before_filter`

app/helpers/orders_helper.rb
```diff
@@ -10,4 +10,13 @@ def status_change(previous, current)
       "Order #{current.status}"
     end
   end
+
+  def show_rating_button
```
mkasztelnik: Maybe better name with `?` at the end? For example `ratingable?`

app/helpers/orders_helper.rb
```diff
@@ -10,4 +10,13 @@ def status_change(previous, current)
...
+    ((@order.status == "ready") && check_ready_date && @order.service_opinion.nil?)
+  end
+
+  def check_ready_date
```
mkasztelnik: Similar to the above. Methods which returns `true`/`false` should end with `?`.

app/helpers/orders_helper.rb
```diff
@@ -10,4 +10,13 @@ def status_change(previous, current)
...
+    order_change = @order.order_changes.find_by(status: :ready)
```
mkasztelnik: you can add `pluck(:created_at)` to fetch only `created_at` column values not whole record.

app/javascript/packs/application.js
```diff
@@ -38,5 +38,17 @@ document.addEventListener("turbolinks:before-render", function(event) {
...
 document.addEventListener('DOMContentLoaded', function () {
   dom.i2svg();
+  $("[data-rating-stars] > li").on('click', function () {
+      let name = $(event.currentTarget.parentElement).attr("data-rating-stars");
+      document.getElementById(name).value
+          = $(event.currentTarget.firstElementChild).attr("value");
+      let stars = event.currentTarget.parentElement.children;
+      $.each(stars, function (index){
+          let star = $(this.firstElementChild);
+          (index < document.getElementById(name).value) ?
```
mkasztelnik: You are doing doc element search multiple times. Maybe this value can be assigned to a local variable?

app/javascript/packs/application.js
```diff
@@ -38,5 +38,17 @@ document.addEventListener("turbolinks:before-render", function(event) {
...
+              star.attr('data-prefix','fas') : star.attr('data-prefix','far')
+      });
+  });
+  dom.watch();
```
mkasztelnik: Can you explain `dom.watch()`?

app/mailers/order_mailer.rb
```diff
@@ -25,6 +25,13 @@ def changed(order)
     end
   end
 
+  def rate_service(order)
+    @order = order
+    @user = order.user
+
+    mail(to: @user.email, subject: "Rate me", template_name: "rating_service")
```
mkasztelnik: Is it a good email subject?

app/models/order.rb
```diff
@@ -22,6 +22,7 @@ class Order < ApplicationRecord
   belongs_to :service
   belongs_to :user
   has_many :order_changes, dependent: :destroy
+  has_one :service_opinion
```
mkasztelnik: What is order is deleted? Maybe `dependent: :destroy` should be added here?

mkasztelnik: or `dependent: :nullify`

app/services/service_opinion/update_service.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class ServiceOpinion::UpdateService
+  def initialize(order)
+    @order = order
+  end
+
+  def call
+    @orders = Order.where(service: @order.service_id)
+
+    @service_opinions = ServiceOpinion.joins(:order).where(orders: { id: @orders })
```
mkasztelnik: Why 2 queries instead of one? Something like:

```ruby
@service_opinions = ServiceOpinion.joins(order: :services).where(orders: { services: { id: @orders } })
```

app/services/service_opinion/update_service.rb
```diff
@@ -0,0 +1,19 @@
...
+    @rating = 0
+    @service_opinions.each do |opinion|
```
mkasztelnik: You can use db count to do this. Something like:

```ruby
@service_opinions.count(:rating)
```

app/javascript/packs/application.js
kmarszalek: https://fontawesome.com/how-to-use/with-the-api/methods/dom-watch without dom.watch() i can change html but it doesnt show in view :(

app/services/service_opinion/update_service.rb
kmarszalek: ... @service_opinions.sum(:rating) ;)

app/helpers/orders_helper.rb
```diff
@@ -10,4 +10,13 @@ def status_change(previous, current)
       "Order #{current.status}"
     end
   end
+
+  def ratingable?
+    ((@order.status == "ready") && check_ready_date? && @order.service_opinion.nil?)
```
mkasztelnik: Check if this helper method is generated by `enum`: `@order.ready?`

app/helpers/orders_helper.rb
```diff
@@ -10,4 +10,13 @@ def status_change(previous, current)
...
+  end
+
+  def check_ready_date?
```
mkasztelnik: Why do you need `check` word in this method name?  Maybe `order_ready?` is better? 

Another issue with this method is that this method uses state (`@order`). Maybe a better option would be to have something like `order_ready?(order)`

app/javascript/packs/application.js
```diff
@@ -31,12 +31,30 @@ document.addEventListener("turbolinks:before-render", function(event) {
...
  */
 document.addEventListener('DOMContentLoaded', function () {
   dom.i2svg();
+  starsOnClick();
+  dom.watch();
 });
 
+function starsOnClick(node){
+    $(node || 'body').find("[data-rating-stars] > li").on('click', function () {
+        let name = $(event.currentTarget.parentElement).attr("data-rating-stars");
```
mkasztelnik: name and value are not reassigned thus `const` is much better.

app/javascript/packs/application.js
```diff
@@ -31,12 +31,30 @@ document.addEventListener("turbolinks:before-render", function(event) {
...
+        let value = document.getElementById(name).value
+            = parseInt($(event.currentTarget.firstElementChild).attr("value"));
+        let stars = event.currentTarget.parentElement.children;
+        $.each(stars, function (index){
+            let star = $(this.firstElementChild);
```
mkasztelnik: The same as above `const` is better here.

spec/support/message_delivery.rb
```diff
@@ -4,4 +4,7 @@ class ActionMailer::MessageDelivery
   def deliver_later
```
mkasztelnik: This method should be removed since other is created. In ruby you cannot have two methods with different signatures.

spec/services/service_opinion/update_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ServiceOpinion::UpdateService do
+  it "updates service rating and service_opinion_count" do
+    service = create(:service)
+    expect(service.rating).to eq(0)
+    expect(service.service_opinion_count).to eq(0)
```
mkasztelnik: 3 lines above looks like a separate test. Something like `it "initializes service with an empty rating and opinion_count"`

spec/services/order/ready_spec.rb
```diff
@@ -22,6 +22,6 @@
     order.new_change(status: :ready, message: "Order is ready")
 
     expect { described_class.new(order).call }.
-      to change { ActionMailer::Base.deliveries.count }.by(1)
+      to change { ActionMailer::Base.deliveries.count }.by(2)
```
mkasztelnik: If there are two emails maybe you should check if e.g. subjects are correct?

spec/models/service_spec.rb
```diff
@@ -49,4 +50,12 @@
 
     it { is_expected.to_not validate_presence_of(:connected_url) }
   end
+
+  it "has rating" do
+    main, other = create_list(:category, 2)
```
mkasztelnik: Why are categories important in this test?

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  it { should validate_numericality_of(:rating) }
+  it { should belong_to(:order) }
+
+  it "service_opinion updates service.rating" do
+    service = create(:service)
+    expect(service.rating).to eq(0.0)
+
+    order = create(:order, service: service)
+
+    service_opinion = create(:service_opinion, order: order, rating: "3")
```
mkasztelnik: unused variable. You can use `create` without assigning it into variable.

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+RSpec.describe ServiceOpinion, type: :model do
+  it { should validate_presence_of(:rating) }
+  it { should validate_numericality_of(:rating) }
+  it { should belong_to(:order) }
+
+  it "service_opinion updates service.rating" do
+    service = create(:service)
+    expect(service.rating).to eq(0.0)
+
+    order = create(:order, service: service)
```
mkasztelnik: The same as bellow

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ServiceOpinion, type: :model do
+  it { should validate_presence_of(:rating) }
+  it { should validate_numericality_of(:rating) }
+  it { should belong_to(:order) }
+
+  it "service_opinion updates service.rating" do
```
mkasztelnik: If you are testing single method behavior you can use `it "#method_name"` test signature. Than when rspec with description flag is used it renders nicely what is tested.

spec/models/service_opinion_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    expect(service.rating).to eq(0.0)
+
+    order = create(:order, service: service)
+
+    service_opinion = create(:service_opinion, order: order, rating: "3")
+
+    expect(service.rating).to eq(3.0)
+  end
+
+  it "service_opinion sends after_save callback" do
```
mkasztelnik: This test is testing internal implementation, not behavior. Good test for this functionality is above. IMHO this should be removed.

spec/mailers/order_spec.rb
```diff
@@ -48,4 +48,20 @@
       expect(encoded_body).to match(/#{order_url(order)}/)
     end
   end
+
+  context "Rating service" do
+    let(:order) { build(:order, id: 1) }
+    let(:mail) { described_class.created(order).deliver_now }
```
mkasztelnik: These two lets are not used anywhere. Can be removed.

app/helpers/orders_helper.rb
kmarszalek: si. https://api.rubyonrails.org/v5.0/classes/ActiveRecord/Enum.html

spec/helpers/orders_helper_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      @order.order_changes.find_by(status: :ready).created_at = RATE_PERIOD - 1.day
+      service_opinion = create(:service_opinion, rating: "3", order: @order)
+      expect(ratingable?).to be_falsy
+    end
+
+    it "is true when order is ready and updated_at before RATE_PERIOD and no service_opinion" do
+      @order = create(:order, status: :created)
+      @order.new_change(status: :ready, message: "Order ready")
+      order_change = @order.order_changes.last
+      order_change.created_at = RATE_PERIOD - 1.day
```
mkasztelnik: You can use time travel to test it instead of updating `created_at` field. I was referring to the time travel rails feature somewhere in this PR.

spec/factories/service_opinion.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :service_opinion do
+    sequence(:rating) { |n| "service #{n} rating" }
```
mkasztelnik: Isn't rating a number?

Issue #111, opened 2018-07-09
As a user I am able to see services supported by a given provider by clicking on a provider's name link in a service detail description


roksanaer:
After clicking on provider's name given in a service detail (https://v81k42.axshare.com/#g=1&p=service-detail_v3 - Onemikro) user is redirected to the provider's service list- Provider's service list should be analogical to the category/filtered service list (https://v81k42.axshare.com/#g=1&p=services-list_v3). @abacz add any additional info if necessary

roksanaer:
connected to #110

abacz:
When you click to oneMikro provider in first item you will see provider services.
https://v81k42.axshare.com/#g=1&p=services-list_provider

How do you think about special header for this kind of list with provider name? I' m thinking about it beacouse it will be unique situation where filter has own header.

roksanaer:
all in for a dedicated header

roksanaer:
IMPORTANT comment:
service provider is an entity supporting the delivery of a service; institution offering a given service; one service might have many providers, a provider can have many services, thus relation in the DB should be many to many

additionally, service owner is not equal to the service provider: a service owner is a person responsible for the service entry (real person with a unique ID in the aai) who has all provilages to edit the service entry in the MP. service provider (at least at the moment) should be an entity in the MP which is an element of a data model for the services. each service should have an attached service provider(s) but at the moment it is only a way of a categorization in a way. there is no real person or organization standing behind it, who can manage services within the service provider entity. 

our goal is to enable the management of the service offers in a scope of a service provider but it is our goal for the future

to sum up:
service provider - relation many to many with a service; entity offering a service but an entity which is not manageable in the beginning - existing only in our data model; a way of organizing the services
service owner - relation one to many (a service owner can have many services) - MP user, assigned to a service responsible for a service entry

martaswiatkowska:
Sooory that it is not ended.
All that I have been done I pushed to brunch 111-service-provider-view
Firstly I implemented that exactly like categories. 
It is needed to implement the sending parameters from check-boxes to index in controller and then filter the services ( that's what I wanted implemented )

PR #167, opened 2018-08-30
[#111] User can search services by provider

martaswiatkowska:
User can filtr services per provider by click on provider name link.
Link is available on services list and services details view.


app/models/provider.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Provider < ApplicationRecord
+  has_many :provided_services,
```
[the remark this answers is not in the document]
mkasztelnik: Than this relation can be defined simply:

```ruby
has_many :services, dependent: :nullify
```

app/models/service.rb
```diff
@@ -13,11 +13,15 @@ class Service < ApplicationRecord
   belongs_to :owner,
              class_name: "User",
              optional: true
+  belongs_to :provider,
+             class_name: "Provider",
```
mkasztelnik: This is not needed when from the relation name class can be guessed. So here something like this should be enough:

```ruby
belongs_to :provider, optional: true
```

app/models/service.rb
```diff
@@ -13,11 +13,15 @@ class Service < ApplicationRecord
...
+             optional: true
 
   validates :title, presence: true
   validates :description, presence: true
   validates :tagline, presence: true
   validates :connected_url, presence: true, url: true, if: :open_access?
+  validates :provider_id, presence: true
```
mkasztelnik:
```ruby
validates :provider, presence: true
```

app/views/backoffice/services/_form.html.haml
```diff
@@ -6,6 +6,7 @@
   = f.input :tagline
   = f.input :connected_url, label: "Service website"
   = f.input :open_access, as: :boolean
+  = f.input :provider_id, collection: get_providers_list
```
mkasztelnik: Take a look at `f.association` simple form helper.

app/views/services/_service.html.haml
```diff
@@ -6,8 +6,8 @@
       .mb-3= service.tagline
       %p.small.float-left.mr-4
         Provided by
-        %a{ href: "" }
-          SHAPE
+        = link_to get_provider_name(service), provider_path(service.provider_id)
```
mkasztelnik:
```haml
= link_to get_provider_name(service), service.provider
```

app/views/services/show.html.haml
```diff
@@ -4,6 +4,10 @@
   %p
     %b= @service.tagline
   %p= @service.description
+  .row
+    .col
+      Provided by
+      = link_to get_provider_name(@service), provider_path(@service.provider_id)
```
mkasztelnik:
```haml
= link_to get_provider_name(service), service.provider
```

config/breadcrumbs.rb
```diff
@@ -40,3 +40,8 @@
     parent :services
   end
 end
+
+crumb :provider do |provider|
+  link "Provider ##{provider.id}", provider_path(provider)
```
mkasztelnik: Do we really want to show DB id as provider name?

lib/tasks/dev.rake
```diff
@@ -18,9 +18,11 @@ if Rails.env.development?
                        owner: users.sample,
                        open_access: Faker::Boolean.boolean,
                        rating: Random.rand(5.0),
-                       connected_url: Faker::Internet.url)
+                       connected_url: Faker::Internet.url,
+                       provider_id: Provider.all.sample.id)
```
mkasztelnik:
```ruby
provider: all_providers.sample
```

lib/tasks/dev.rake
```diff
@@ -18,9 +18,11 @@ if Rails.env.development?
                        owner: users.sample,
                        open_access: Faker::Boolean.boolean,
                        rating: Random.rand(5.0),
-                       connected_url: Faker::Internet.url)
+                       connected_url: Faker::Internet.url,
```
mkasztelnik: How this is connected with this PR?

spec/features/backoffice/services_spec.rb
```diff
@@ -40,14 +40,19 @@
   end
 
   scenario "I can create new service" do
+    create(:provider)
     visit backoffice_services_path
     click_on "Create new service"
 
+    expect(Provider.all.count).to eq(1)
```
mkasztelnik: How is this relevant to this test?

spec/helpers/service_helper_spec.rb
```diff
@@ -14,4 +14,15 @@
   it "converts from decimal 0.0 value to html" do
     expect(print_rating_stars(0.0)).to match(/<i class="far fa-star"><\/i><i class="far fa-star"><\/i><i class="far fa-star"><\/i><i class="far fa-star"><\/i><i class="far fa-star"><\/i>/)
   end
+
+  it "return provider name" do
+    provider = create(:provider)
+    service = create(:service, provider_id: provider.id)
```
mkasztelnik: I propose to be explicit when possible:
```ruby
provider = create(:provider, name: "my provider")
service = create(service, provider: provider)
expect(get_provider_name(service)).to eq("my provider")
```

lib/tasks/dev.rake
martaswiatkowska: Comma is needed :)

lib/tasks/dev.rake
martaswiatkowska: Are you sure? all_providers are defined in seeds.rb

mkasztelnik: You are right the goal here is not to query db every time you are creating service. You can fetch providers list before `times` block.

Issue #161, opened 2018-08-08
ordering is not working


roksanaer:
chrome, safari
current behaviour
1. entering a MP - logged out
2. log in
3 click on a "services" tab - i enter a list of services
4. opening one of the services detailed view
5. configure the service on service attrributes tab
6. clicking on "order" button
7. i am being redirected to the main page: https://marketplace.docker-fid.grid.cyf-kr.edu.pl/; there is no new order on my order list (so probably order has not been created)

expected behaviour
1. entering a MP - logged out
2. log in
3 click on a "services" tab - i enter a list of services
4. opening one of the services detailed view
5. configure the service on service attrributes tab
6. clicking on "order" button
7. i am being redirected to the summary page (see axure: https://v81k42.axshare.com/#g=1&p=summary-step-3-end)

offtopic:
@abacz when views that reflect the axure order behaviour will be ready (so the request-step2)

mkasztelnik:
@roksanaer is it still valid?

PR #179, opened 2018-09-19
[#160 #161] Added additionaly step to order creation process 

### LINKED ISSUE #160: new behaviour of the order button for unlogged user
Chrome/safari
Current behavior:
1. entering a service - logged out
2 click on a "services" tab - i enter a list of services
3. opening one of the services detailed view
4. i can see the service description, attributes tab, opinion tab
5. i do not now what to do next and if i can do anything about the service I am looking at
(6. "order" button is missing as long as somebody loggs in)

Expected behaviour:
1. entering a service - logged out
2 click on a "services" tab - i enter a list of services
3. opening one of the services detailed view
4. i can see the service description, attributes tab, opinion tab, "order" button
5. i want to order the service, so i click the order button
6. i am not logged in so the system redirects me to the logging page (AAI)
7. after logging in I am being redirected to the order sum up page

Explanation:
not logged in user should know what are the options connected with the service no matter the loggin status. ordering button should be visible for everybody but the action of ordering (and proceeding order creation etc.) should be possible only after logging in

martaswiatkowska:
Ordering Button is visible for logged in and unlogged users. 
Added cart as page with order summary(only service  title). This solve the problem with redirection back after login in to the created order. When user click on "Order" service_id is saved in to the session and redirected to cart_path where mus be logged in.
Set Redis as session store.  
"Rack_session_access" gem is for session management in tests.

Fixed #160 #161

### DIFF
--- a/app/views/carts/_order_item.html.haml
+++ b/app/views/carts/_order_item.html.haml
@@ -1,0 +1,10 @@
+
+.row
+  - if order_item
+    = form_for(Order.new) do |f|
+      %p= order_item.title
+      = f.hidden_field :service_id, value: order_item.id
+      %button{ class: "btn btn-primary" }
+        Order
+  - else
+    %p= "Cart is empty"
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+if ENV["REDIS_URL"]
+  url = ENV["REDIS_URL"]
+else
+  url = "redis://localhost:6379/0/mp-session"
+end
+
+if Rails.env.test?
+  Rails.application.config.session_store :cookie_store,
+    key: "_mp_session"
+else
+  Mp::Application.config.session_store :redis_store,
+    servers: url,
+    expire_after: 90.minutes,
+    key: "_mp_session",
+    threadsafe: false
+end
--- a/spec/features/carts_spec.rb
+++ b/spec/features/carts_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Cart" do
+  include OmniauthHelper
+
+  let(:user) { create(:user) }
+  let(:service) { create(:service) }
+
+  before { checkin_sign_in_as(user) }
+
+  scenario "I can create Order service" do
+    visit service_path(service)
+    click_on "Order"
+
+    expect { click_on "Order" }.
+      to change { Order.count }.by(1)
+    order = Order.last
+
+    expect(order.service).to eq(service)
+    expect(order.user).to eq(user)
+    expect(order).to be_created
+    expect(page).to_not have_content("Cart is empty")
+  end
+end
--- a/spec/features/orders_spec.rb
+++ b/spec/features/orders_spec.rb
@@ -3,10 +3,9 @@
 require "rails_helper"
 
 RSpec.feature "Service order" do
-
+  include OmniauthHelper
 
   context "as logged in user" do
-    include OmniauthHelper
 
     let(:user) { create(:user) }
     let(:service) { create(:service) }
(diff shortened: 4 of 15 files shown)


roksanaer:
when testing on deployed instance on 25th, the feature was not working:
order button not visible when logged out
after login in and clicking the order button, the system redirected me to the main page with info about successful logging in

spec/features/orders_spec.rb
```diff
@@ -119,15 +115,41 @@
...
+      open_access_service =  create(:open_access_service)
+
+      visit service_path(open_access_service)
+
+      expect(page).to have_selector(:link_or_button, "Add to my services", exact: true)
+      expect(page).to have_selector(:link_or_button, "Go to the service", exact: true)
     end
 
-    scenario "I don't my services page" do
+    scenario "I cannot see my services page" do
```
mkasztelnik: Why this change? I'm not trying to enter forbidden path just checking if link to this section is not visible.

config/initializers/session_store.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+if Rails.env.test?
+  Rails.application.config.session_store :cookie_store,
+    key: "_mp_session"
+else
+  Mp::Application.config.session_store :redis_store,
+    servers: ["redis://localhost:6379/0/mp_session"],
```
mkasztelnik: This should be externalized since it needs to be configured in a specific way in docker environment.

/cc @wziajka

spec/features/carts_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  include OmniauthHelper
+
+  let(:user) { create(:user) }
+  let(:service) { create(:service) }
+
+  before { checkin_sign_in_as(user) }
+
+  scenario "I can create Order service" do
+    order_temp = { "service_id": service.id }
+    page.set_rack_session(order_item: order_temp)
```
mkasztelnik: Instead of setting specific values into rack session you can just visit two pages and clicks 2 buttons in this test :smile:

spec/features/carts_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    visit cart_path(service)
+
+    expect { click_on "Order" }.
+      to change { Order.count }.by(1)
+    order = Order.last
+
+    expect(order.service).to eq(service)
+    expect(order.user).to eq(user)
+    expect(order).to be_created
+    expect(page.get_rack_session).to_not have_key("order_item")
```
mkasztelnik: Implementation detail. If you want to check if this is clearer than you can confirm an order and next visit confirmation page to check if card is empty.

spec/features/orders_spec.rb
```diff
@@ -26,31 +25,28 @@
...
-      order = Order.last
+      click_button "Order"
 
-      expect(order.service).to eq(service)
-      expect(order.user).to eq(user)
-      expect(order).to be_created
+      expect(page).to have_current_path(cart_path)
+      expect(page).to have_text(service.title)
+      expect(page).to have_selector(:link_or_button,
+                                    "Order", exact: true)
```
mkasztelnik: Here you should also click another button and check if order is created.

app/views/carts/_order_item.html.haml
```diff
@@ -0,0 +1,10 @@
+
+.row
+  - if order_item
+    = form_for(Order.new) do |f|
+      %p= order_item.title
+      = f.hidden_field :service_id, value: order_item.id
+      %button{ class: "btn btn-primary" }
+        Order
+  - else
+    %p= "Cart are empty"
```
kmarszalek: is

spec/features/orders_spec.rb
martaswiatkowska: If I add this then it will be the same test as in the cart_spec.rb. Should I delete one of them?

mkasztelnik: Let's leave it at it is. This view will be updated according to the new ordering schema than we will improve it.

Issue #85, opened 2018-07-04
As a user I'm able to sort the results by ranking


roksanaer:

kmarszalek:
It might be helpful, from Marek:
Some finding to investigate:

https://github.com/merit-gem/merit this is more reputation, badges engine, but we take a look at is during research (there is rankings support https://github.com/merit-gem/merit#rankings)
Simple example how to implement ranking in rails na postgresql using ntile window function http://naturaily.com/blog/post/ruby-on-rails-implementation-of-a-ranking-system-using-postgresql-window-functions
must ready by person implementing this issue
More advanced ranking with positives and negatives: http://www.akitaonrails.com/2016/10/31/ruby-on-rails-implementation-of-a-proper-ranking-popularity-system - very good read about good and bad ranking implementation.

abacz:
You can find sorting on services-list with 4 types - by name and rate
https://v81k42.axshare.com/#g=1&p=services-list_v3

PR #187, opened 2018-10-03
[#84, #85] Service sort framework

michal-szostak:
# What is in this PR

This PR includes sorting via select options picking (in `/services` view).

Codevise following things has been added:

* JS support for service searching which is split into 3 forms.
* JS handler for searching on change in select 
* Fixed mini bug when all side panel (categories & filtering) was being hidden if no categories were present in database
* Included chrome webdriver gem, which was required for JS testing
* Add chrome addon to `.travis.yml`

Fixed: #84, #85


spec/features/service_spec.rb
```diff
@@ -28,3 +28,35 @@
...
+
+  after { Capybara.current_driver = :rack_test }
+
+  scenario "searching in top bar will preserve existing query params" do
+    visit services_path(sort: "title")
+
+    fill_in "q", with: "abc"
+    click_on(id: "query-submit")
+
+    expect(page).to have_current_path(services_path + "?utf8=✓&q=abc&sort=title")
```
mkasztelnik: IMHO url is not important from the feature point of view. A better test (this is for all tests bellow also) would populate DB with some services and checks if the services order is correct.

spec/rails_helper.rb
```diff
@@ -53,6 +53,13 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 
+  Capybara.register_driver :chrome do |app|
+    options = Selenium::WebDriver::Chrome::Options.new(args: %w[no-sandbox headless disable-gpu])
```
mkasztelnik: Previously I had troubles with this configuration (using `options`). The solution was to use something like this:

```ruby
Capybara.register_driver :headless_chrome do |app|
  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
    chromeOptions: { args: %w[headless disable-gpu no-sandbox] }
  )

  Capybara::Selenium::Driver.new app,
                                 browser: :chrome,
                                 desired_capabilities: capabilities
end

Capybara.javascript_driver = :headless_chrome
```

app/controllers/concerns/service/searchable.rb
```diff
@@ -4,9 +4,26 @@ module Service::Searchable
   extend ActiveSupport::Concern
 
   private
-
     def records
-      params[:q].blank? ? scope : scope.where(id: search_ids)
+      order params[:q].blank? ? scope : scope.where(id: search_ids)
```
mkasztelnik: Please use `()` for such complex lines or use local variable to assign elements and then pass them into order function

app/controllers/concerns/service/searchable.rb
```diff
@@ -4,9 +4,26 @@ module Service::Searchable
...
+    end
+
+    def order(elements)
+      unless params[:sort] && params[:sort].length > 0
```
mkasztelnik:
```ruby
if params[:sort].blank? # this is also valid nil.blank?
  elements
else
  ...
end
```

app/javascript/packs/sort_filter.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export function registerSubmitOnChange(node) {
+    $(node || 'body').find("[data-submit-on-change]").on('change', function () {
+        let id = $(this).attr('id');
+        let value = $(this).val();
+
+        let search = setSearchParams(window.location.search, {[id]: value});
+        // TODO: Someone might commemnt about how it would work with turbolinks...
+        window.location.search = search;
```
mkasztelnik: Take a look into: https://github.com/turbolinks/turbolinks-classic#triggering-a-turbolinks-visit-manually

spec/features/service_spec.rb
michal-szostak: Ok, I see the point, you might be right on this. On the other hand do you have any tested and easy way to get JS tests up and running? (that would make unit testing that JS code so much easier...)

I didn't want to setup whole testing framework for JS because:
1. It would take too much time for 2 functions
2. There is still not that much JS code

spec/features/service_spec.rb
```diff
@@ -28,3 +28,35 @@
     expect(page.body).to match(/<h1>Terms of use h1/)
   end
 end
+
+
+RSpec.feature "Service filtering and sorting" do
+  before { Capybara.current_driver = Capybara.javascript_driver }
+
+  after { Capybara.current_driver = :rack_test }
```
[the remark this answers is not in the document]
michal-szostak: O, that's nifty

michal-szostak:
On the other hand as I expected travis tests failed completly, that is mostly related to the lack of the web driver. Do you have any idea @mkasztelnik what is the cleanest way to add it to travis?

mkasztelnik:
Try to run xvfb like described here: http://quyetbui.info/chrome-headless-capybara-on-travis-ci/

app/controllers/concerns/service/searchable.rb
```diff
@@ -4,9 +4,26 @@ module Service::Searchable
   extend ActiveSupport::Concern
 
   private
-
     def records
-      params[:q].blank? ? scope : scope.where(id: search_ids)
+      order params[:q].blank? ? scope : scope.where(id: search_ids)
+    end
+
+    def order(elements)
```
mkasztelnik: I was thinking about ordering and I believe that this is not the right place to put it. `Searchable` concern is about searching records not about ordering :smile: 

Can you move ordering into separate concern? Than the api can looks as follow (e.g. in `ServicesController`):
```ruby
class ServicesController < ApplicationController
  include Service::Searchable
  include Service::Orderable

  def index
    @services = records.order(ordering).page(params[:page])
    @subcategories = Category.roots
  end
end
```

If no ordering is given then `ordering` method should return `{}` (or `nil` but `{}` is better because additional ordering can be merged).

michal-szostak: If you say so, I thought that it's pretty simple, so I thought that merging it together with Searchable would be good enaugh.

I have another question though - Is there a need (and if so how to do it in the easiest most "ROR way") - of filtering possible params for ordering?

michal-szostak:
I addressed all of the review points

app/controllers/concerns/service/sortable.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      if sort_key[0] == "-"
+        sort_key = sort_key[1..-1]
+        sort_options[sort_key] = :desc
+      else
+        sort_options[sort_key] = :asc
+      end
+
+      sort_options
+    end
```
mkasztelnik: This can be a little bit shorten:

```ruby
def ordering
  sort_key = params[:sort]
￼  sort_options = {}

  unless params[:sort].blank?
    if sort_key[0] == "-"
        sort_key = sort_key[1..-1]
        sort_options[sort_key] = :desc
      else
        sort_options[sort_key] = :asc
      end
  end

  sort_options
end
```

or 

```ruby
def ordering
￼  {}.tap do |sort_options|
    sort_key = params[:sort]
    unless params[:sort].blank?
      if sort_key[0] == "-"
          sort_key = sort_key[1..-1]
          sort_options[sort_key] = :desc
        else
          sort_options[sort_key] = :asc
        end
    end
  end
end
```

spec/rails_helper.rb
mkasztelnik: I don't really remember. The difference was CI (the other project was build on gitlab ci).

You don't need to change it, I just paste it here just to have hint when something will fail for a strange reason :)

michal-szostak: Alright ;)

Issue #188, opened 2018-10-03
Services do not work on master deployments 


wziajka:
In current master deploy (version: master 8d3332effd72a19052401a1d3852ffe00f0102c2) services do not work, server returns 500 error code. 

Logs 

I, [2018-10-03T12:40:23.850004 #1]  INFO -- : [88325369-df24-455d-9f3b-77bd0b33f2e3] Started GET "/services" for 149.156.10.86 at 2018-10-03 12:40:23 +0000
I, [2018-10-03T12:40:23.851100 #1]  INFO -- : [88325369-df24-455d-9f3b-77bd0b33f2e3] Processing by ServicesController#index as HTML
I, [2018-10-03T12:40:23.852622 #1]  INFO -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Rendering services/index.html.haml within layouts/application
D, [2018-10-03T12:40:23.854273 #1] DEBUG -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Category Exists (0.4ms)  SELECT  1 AS one FROM "categories" WHERE "categories"."ancestry" IS NULL LIMIT $1  [["LIMIT", 1]]
D, [2018-10-03T12:40:23.855081 #1] DEBUG -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Category Load (0.5ms)  SELECT "categories".* FROM "categories" WHERE "categories"."ancestry" IS NULL
I, [2018-10-03T12:40:23.858287 #1]  INFO -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Rendered collection of categories/_subcategory.html.haml [30 times] (2.5ms)
I, [2018-10-03T12:40:23.858414 #1]  INFO -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Rendered categories/_subcategories.html.haml (5.5ms)
D, [2018-10-03T12:40:23.859428 #1] DEBUG -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Service Load (0.6ms)  SELECT  "services".* FROM "services" LIMIT $1 OFFSET $2  [["LIMIT", 30], ["OFFSET", 0]]
I, [2018-10-03T12:40:23.861880 #1]  INFO -- : [88325369-df24-455d-9f3b-77bd0b33f2e3]   Rendered collection of services/_service.html.haml [18 times] (1.7ms)
[...]

PR #190, opened 2018-10-05
175 duplicated categories bugfix

goreck888:
fix problems with update and create categories and services

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -13,25 +13,26 @@ require "yaml"
 puts "Generating categories from yaml"
 yaml_hash = YAML.load_file("db/data.yml")
 
-yaml_hash["categories"].each do |category, hash|
-  parent = hash["parent"] && Category.find_by(name: hash["parent"])
-  Category.create!(name: hash["name"], description: hash["description"], parent: parent)
-end
-
-puts "Generating services from yaml"
-yaml_hash["services"].each do |service, hash|
-  current = Service.create(title: hash["title"], tagline: hash["tagline"], description: hash["description"])
-  current.categories << Category.find_by(name: hash["parent"])
-  current.set_first_category_as_main!
-
+yaml_hash["categories"].each do |_, hash|
+  Category.find_or_initialize_by(name: hash["name"]) do |category|
+    category.update!(description: hash["description"],
+                     parent: Category.find_by(name: hash["parent"]))
+  end
+  puts "Generated category #{ hash["name"] }"
 end
 
 puts "Generating providers"
-all_providers = []
-all_providers << Provider.create_with(name: "Provider 1").find_or_create_by(name: "Provider 1")
-all_providers << Provider.create_with(name: "Provider 2").find_or_create_by(name: "Provider 2")
-all_providers << Provider.create_with(name: "Provider 3").find_or_create_by(name: "Provider 3")
-all_providers << Provider.create_with(name: "Provider 4").find_or_create_by(name: "Provider 4")
+providers = (1..4).map { |i| Provider.find_or_create_by(name: "Provider #{i}") }
 
-services_size = ENV["services_size"].to_i || 0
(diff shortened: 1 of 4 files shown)


goreck888:
Fix [#188]

db/seeds.rb
```diff
@@ -33,5 +30,12 @@
 all_providers << Provider.create_with(name: "Provider 3").find_or_create_by(name: "Provider 3")
 all_providers << Provider.create_with(name: "Provider 4").find_or_create_by(name: "Provider 4")
 
-services_size = ENV["services_size"].to_i || 0
-Rake::Task["dev:prime"].invoke(services_size)
+
+puts "Generating services from yaml"
+yaml_hash["services"].each do |service, hash|
+  current = Service.create_with(title: hash["title"], tagline: hash["tagline"], description: hash["description"], provider: Provider.all.sample).find_or_create_by(title: hash["title"])
```
martaswiatkowska: use enter

db/seeds.rb
```diff
@@ -33,5 +30,12 @@
...
+  current.categories << Category.find_by(name: hash["parent"])
```
martaswiatkowska: hash["parent"]? How it's working?

db/seeds.rb
```diff
@@ -33,5 +30,12 @@
...
+  current.set_first_category_as_main!
+  puts "Generated service #{ hash["title"] }"
+
```
martaswiatkowska: I think you can delete this blank line

db/data.yml
```diff
@@ -2,7 +2,10 @@ categories:
 
     Compute:
         name: &computeParent Compute
-        description: General computing services including VM management, containers and job processing
+        description: General computing services including VM management, containers and job processing.
```
mkasztelnik: It looks like you are doing 2 things in one PR
  1. Duplicated categories fix (and duplicated services fix - so this is third thing :))
  1. Adding/Updating `yaml` with additional service information

mkasztelnik:
There is one additional thing I don't like. Every time you are doing `ModelName.create_with(...)` db query e.g.:

```
[18] pry(main)> a=Category.create_with(description: 'asdf')
  Category Load (0.4ms)  SELECT "categories".* FROM "categories"
```
is executed. This can be fixed by changing code into:

```ruby
require "yaml"
puts "Generating categories from yaml"
yaml_hash = YAML.load_file("db/data.yml")

yaml_hash["categories"].each do |_, hash|
  Category.find_or_initialize_by(name: hash["name"]) do |category|
    category.update_attributes(description: hash["description"],
                               parent: Category.find_by(name: hash["parent"]))
  end
  puts "Generated category #{ hash["name"] }"
end

puts "Generating providers"
providers = (1..4).map { |i| Provider.find_or_create_by(name: "Provider #{i}") }


puts "Generating services from yaml"
yaml_hash["services"].each do |_, hash|
  Service.find_or_create_by(title: hash["title"]) do |service|
    service.update_attributes(tagline: hash["tagline"],
                              description: hash["description"],
                              provider: providers.sample,
                              categories: [Category.find_by(name: hash["parent"])])

  end
  puts "Generated service #{ hash["title"] }"
end
```

As you can see I also removed `current.set_first_category_as_main!` line. This is not needed since Service will guard categories and will set first category into main when needed.

This solution has one additional plus: when anything is updated in the category or service except name than it will be updated. This was not a case in the previous version.

mkasztelnik:
According to https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update-21 `update_attributes!` can be changed into `update!`

db/seeds.rb
goreck888: It finds category by name included in hash parent in yaml. It works exactly like you can see in the code.

mkasztelnik:
One small remark: changelog entry is missing!

Issue #112, opened 2018-07-09
As a service owner I am able to insert an e mail dedicated for questions about the service from registered users


roksanaer:
refers to #38

mkasztelnik:
Done by @martaswiatkowska

PR #195, opened 2018-10-10
[#118] Send message about services

### LINKED ISSUE #38: As registered user I'm able to contact service provider and ask about service details
A form available from the service entry (detailed view). 
https://kcjh3g.axshare.com/#g=1&p=service-detail_v3 under the 'proceed to order' button there is another: "want to ask a question about this service?". Clicking on it opens a one field form (modal) allowing the registered user to enter his inquiry. Hitting "send" button should send a mail to a mailbox connected with that service (this email might be different from the service owner email, it should be a dedicated field connected with a service in our database).

### LINKED ISSUE #118: As a registered user I can contact service representatives

martaswiatkowska:
Added on service details view link "Want to ask a question about this service? " if user is logged in, under the 'order' button. Clicking on it opens a modal where user can write a message which is send to emails connected with that service. 
If user is not logged in, then a message "If you want ask question about service please login!" is displayed
Service owner is able to insert an email dedicated for contact on the create service view. 

Fix #118 (epik) #38 #112


app/views/backoffice/services/_form.html.haml
```diff
@@ -7,9 +7,12 @@
   = f.input :connected_url, label: "Service website"
   = f.input :open_access, as: :boolean
   = f.association :provider
+  = f.input :contact_emails, collection: User.all.collect { |u| u.email },
```
mkasztelnik: This is so terribly danger. Here you are able to get all marketplace users emails - huge security risk, rodo, etc.

Gemfile
```diff
@@ -77,6 +77,7 @@ group :test do
   gem "shoulda-matchers"
   gem "capybara"
   gem "database_cleaner"
+  gem "capybara-webkit"
```
mkasztelnik: Why this gem is needed? Can't you use chrome web driver? To turn on js support you need to add:

```ruby
it "do something with js", js: true do
  ...
end
```

app/mailers/service_mailer.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ServiceMailer < ApplicationMailer
+  def new_question(recipient_email, service_question, service)
+    @service = service
+    @message = service_question[:text]
+    @author = User.find(service_question[:author])
+    @recipient = User.find_by(email: recipient_email)
```
mkasztelnik: Is it mandatory that email given in `Service` contact list belongs to registered marketplace user? If yes then why cannot we have relation 1-n between `Service` and `User` but plain email is used?

app/mailers/service_mailer.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    mail(to: @recipient.email,
+         from: @author.email,
+         subject: "Question about #{@service.title} service",
+         template_name: "new_message")
```
mkasztelnik: Why you are not following convention here? "new_message" is already used somewhere else?

app/views/services/_description.html.haml
```diff
@@ -11,6 +11,12 @@
             = write_button_text
           - if @service.open_access?
             = link_to "Go to the service", service.connected_url, class: "btn btn-primary", target: "_blank"
+      - if current_user
```
mkasztelnik: It would be nice to move it into separate partial

app/views/services/_description.html.haml
```diff
@@ -40,3 +46,6 @@
           %a{ href: "" }SLA
         %li
           %a{ href: "" }Access policy
+
+- if current_user
```
mkasztelnik: This can be moved into the same partial mentioned above

spec/features/service_spec.rb
```diff
@@ -7,25 +7,74 @@
...
-    expect(page.body).to have_content service.tagline
+      visit service_path(service)
+
+      expect(page.body).to have_content service.title
+      expect(page.body).to have_content service.description
+      expect(page.body).to have_content service.terms_of_use
+      expect(page.body).to have_content service.tagline
+    end
+
+    scenario "terms of use is rendered from markdown to html" do
```
mkasztelnik: How this spec is connected with this PR?

spec/features/service_spec.rb
```diff
@@ -7,25 +7,74 @@
...
+    scenario "I see Ask Question" do
+      service = create(:service)
+
+      visit service_path(service)
+
+      expect(page).to have_content "Want to ask a question about this service?"
+
+    end
+
+    scenario "I can see question-modal if I click on link" do
```
mkasztelnik: `js: true` ?

spec/features/service_spec.rb
```diff
@@ -7,25 +7,74 @@
...
+      visit service_path(service)
+
+      find("#modal-show").click
+
+      within("#question-modal") do
+        fill_in("service_question_text", with: "text")
+      end
+
+      expect { click_on "SEND" }.
+        to change { sleep(1); ActionMailer::Base.deliveries.count }.by(2)
```
mkasztelnik: Why `sleep(1)`?

spec/features/service_spec.rb
```diff
@@ -7,25 +7,74 @@
...
+      expect(page).to have_content("Your message was successfully sended")
+    end
   end
 
-  scenario "terms of use is rendered from markdown to html" do
-    service = create(:service, terms_of_use: "# Terms of use h1")
+  context "as not logged in user" do
+    scenario "I see Ask Question" do
```
mkasztelnik: Better naming is needed here, maybe "I need to login to asks service question"

spec/spec_helper.rb
```diff
@@ -18,6 +19,10 @@
 # it.
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
+
+# Set javascript_driver to capybara-webkit
+Capybara.javascript_driver = :webkit
```
mkasztelnik: Please use chrome driver. It is configured in `rails_helper` already.

app/mailers/service_mailer.rb
martaswiatkowska: Yes it was in order questions but I already changed this to "new_question"

spec/features/service_spec.rb
martaswiatkowska: It is not, perhaps I added this while merging

spec/features/service_spec.rb
martaswiatkowska: Because test failed but I already deleted this

Issue #182, opened 2018-10-01
adapt 'My profile' 


roksanaer:
adapt my profile page according to the latest project:
[https://kcjh3g.axshare.com/#g=1&p=my-profile__1_](https://kcjh3g.axshare.com/#g=1&p=my-profile__1_)

mkasztelnik:
Already done by @kmarszalek

Issue #185, opened 2018-10-01
include mailchimp newsletter in 'my profile' newsletter tab


roksanaer:
according to:
https://kcjh3g.axshare.com/#g=1&p=my-profile__1_

roksanaer:
relates to #182

kmarszalek:
@mkasztelnik comment from https://github.com/cyfronet-fid/marketplace/pull/212
UI looks ok but there will be a problem with this accordion. Most probably in the second tab you will have some kind of form which will allow to turn on/off newsletter. After clicking on/off user will be redirected into /profile view (first accordion tab).

To solve this issue I propose to create dedicated controller (newsletter_controller.rb) with dedicated show method which reuses profile layout with one small modification - second accordion tab is active.

PR #212, opened 2018-10-16
[#182] Add tabs in My profile page

kmarszalek:
Adapt profile view with tabs (affiliations and newsletter)


kmarszalek:
@abacz style?

kmarszalek:
yes, and it should be done in https://github.com/cyfronet-fid/marketplace/issues/185 where someone will add newsletter and this problem will arise?

Issue #223, opened 2018-10-23
As a user I can see related services of the service (service entry page)


roksanaer:
under the service entry, boxes of related services should be visible
https://kcjh3g.axshare.com/#g=1&p=service-detail_service_with_options (element 7)
the header should change to: Services you can use with this service (from: Customers who ordered this service also viewed)
The service miniature (box) should include the same attributes as miniatures on a service list:
name
tagline 
Provided by 
Available for (targeted users)         
Area (scientific discipline)

for now, 1-n relation between services should be possible to enable this functionality

roksanaer:
soon a specific example will appear as a new ticket, related to this one

mkasztelnik:
In the scope of #241 backend and basic UI was created. UI styling is needed and additional information is required in the `Service` model. 

@roksanaer who can deal with UI?

roksanaer:
@abacz are the UI elements ready for this? If no, can you be the one preparing this or somebody else should take care of it?

abacz:
No, this view is not ready. I am going to design it after sat4envi meeting (tuesday).

PR #241, opened 2018-10-24
Related services

mkasztelnik:
Related services are shown on service show view and in the service order confirmation page. n-n relation between `Services` is modeled using `ServiceRelationship` model. It can be used to model additional
relationships between services. E.g. additional field `relation_type` can be added to `ServiceRelationship`, next `where` condition can be added into `Service` `has_many` relation between services.

This PR solves everything described in #223 except:
  * formatting (nice css which deals with multiple related services need to be created - there can be 1 or 10 related services thus some grid layout need to be applied)
  * showing elements which are not yet available in the service model (`Area (scientific discipline)`)

This is how it looks like right now:
related-services

To create relation between services following snippet can be used:

```ruby
service = Service.first
related = Service.second

ServiceRelationship.new(source: service, target: related)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Add breadcrumbs to order index and show (@mkasztelnik)
 - Tabs to My profile view (@kmarszalek)
 - 3 steps ordering (@mkasztelnik)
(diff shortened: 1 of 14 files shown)


mkasztelnik:
[flex-wrap](https://getbootstrap.com/docs/4.1/utilities/flex/) can be used to nicely show related services

PR #246, opened 2018-10-25
Skip offer selection when only one offer

mkasztelnik:
@wziajka after this PR is merged following rake task need to be invoked manually once

```
rails service_counter 
```

this will generate counter cache for existing service offers


db/migrate/20181025095520_add_offers_count.rb
```diff
@@ -0,0 +1,5 @@
+class AddOffersCount < ActiveRecord::Migration[5.2]
```
kmarszalek: strings

mkasztelnik: done

PR #256, opened 2018-10-29
Render real root categories in footer

mkasztelnik:


bwilk:
All good besides failing travis build.

PR #257, opened 2018-10-29
Remove MP version from footer

mkasztelnik:
Since MP version is shown in the administration panel it can be removed from the footer where it looks very ugly.


kmarszalek:
FYI: @roksanaer @wziajka @agpul

PR #259, opened 2018-10-29
Show offers as cards

mkasztelnik:
Preliminary version for showing service offers using bootstrap cards. Offers cards require work from UX team (cc @kosmidma, @abacz), but I believe this is a good start. 

What needs to  be done by UX team:
  * Better styling
  * small JS for adding active class into the card (than some additional border and button icon can be added) - @michal-szostak can you help with this issue? 

Sample screenshots showing how it looks like right now:

order-options

service-offers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Order steps layout (@mkasztelnik)
 - Sidekiq for delayed jobs (@mkasztelnik)
 - Admin administration panel stub with sidekiq monitoring and MP version (@mkasztelnik)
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Order steps layout (@mkasztelnik)
 - Sidekiq for delayed jobs (@mkasztelnik)
 - Admin administration panel stub with sidekiq monitoring and MP version (@mkasztelnik)
+- Use card to show service offer (@mkasztelnik)
```
kmarszalek: i would write more info - ?cart?

Issue #236, opened 2018-10-23
update the services' seeds in the application


roksanaer:

mkasztelnik:
@goreck888, @roksanaer is it valid anymore? Can we close it?

roksanaer:
yes :) thanks!

PR #261, opened 2018-10-30
[#236] Update services

goreck888:
Add new services
Add new services fields


db/seeds.rb
```diff
@@ -40,10 +45,11 @@
                     access_policies_url: hash["access_policies_url"],
                     places: hash["places"],
                     languages: hash["languages"],
-                    dedicated_for: hash["dedicated_for"],
+                    # dedicated_for: hash["dedicated_for"],
```
mkasztelnik: Why this is commented? Should this field be removed?

db/schema.rb
```diff
@@ -143,6 +143,7 @@
     t.text "dedicated_for", null: false
     t.text "terms_of_use_url", null: false
     t.text "access_policies_url", null: false
+    t.text "area", null: false
```
mkasztelnik: You need to add `area` field into `spec/factories/services.rb` factory to fix specs.

mkasztelnik: And this is strange - change in `schema.rb` but no migration. How did you lead to such a situation?

mkasztelnik:
A lot of changes not related to this PR is included. Can you fix it?

roksanaer:
and another minor comment. types of users included in the descriptions are not from our list, which goes:

Researchers
VO
Providers
Research project
Business

@agpul please map targeted users to ours so @goreck888 can introduce changes accordingly

I saw that EGI included Research organizations in their eIC description. if other services also have it, we can add it to the list

roksanaer:
>After analysis I am going to change Research project to research organisations. Research project doesnt fit here for me

OK

mkasztelnik:
Also changelog entry is missing

roksanaer:
providers need to have n-n relation model but for tomorrow's purpose it's great. thanks!

PR #262, opened 2018-10-30
Service view redesigned

mkasztelnik:
Following elements were added/improved:
  * service logo can be added to the service
  * services list redesigned
  * service header redesigned
  * related service partial prototype using bootstrap cards

Screenshots from the current version:

services-list

service-show

service-related

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Admin administration panel stub with sidekiq monitoring and MP version (@mkasztelnik)
 - Render root categories in page footer (@mkasztelnik)
 - Use bootstrap cards to show service offers (@mkasztelnik)
+- Add service logo (@mkasztelnik)
 
 ### Changed
 - Upgrade Sprockets gem to avoid CVE-2018-3760 vulnerability (@mkasztelnik)
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,9 @@ gem "will_paginate", "~> 3.1.0"
 gem "will_paginate-bootstrap4"
 gem "simple_form"
 
+gem "activestorage-validator"
+gem "image_processing", "~> 1.2"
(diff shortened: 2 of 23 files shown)


mkasztelnik:
@wziajka after this is merged storage location can be configured using `STORAGE_DIR` env variable.

roksanaer:
pr-262 or pr-263 redirects me to https://marketplace-1.docker-fid.grid.cyf-kr.edu.pl where i see no changes at all (there is even old top nav. bar)

Hovewer I have some comments relying on the screens:
- logo: OK
- service headers on the list: @abacz should they be so big?? and the font - is it the right one?
- "Available for SME within one day" is supposed to be replaced with "Dedicated for" from the service form. @abacz can you please change it in the mock up so it doesn't create confusion (right now we will not provide info how soon the service will be ready to use) - in a service entry and service box information about targeted users should be provided (field Dedicated for implemented by @kmarszalek, which is right now visible on the right side of a service entry and it should be removed from there.
- boxes for connected services: OK, great
- @mkasztelnik where is the placeholder for the service offer(s)?

mkasztelnik:
@roksanaer service offers are shown only when there are offers :) It looks like that:

offers

For the styling, I did not change anything in css, only bootstrap classes are used. Fonts, h* sizes, etc. should be set globally (and I hoped it was done long time ago by @kosmidma with @abacz guidance...).

roksanaer:
>service offers are shown only when there are offers :) 
makes sense ;)
>Fonts, h* sizes, etc. should be set globally (and I hoped it was done long time ago by @kosmidma with @abacz guidance...).
yeah, so @abacz add it for @kosmidma to fix :)

so it leaves us only to replace "Available for SME within one day" with "Dedicated for". thanks!

abacz:
I've already add quick simple guidelines -> https://sketch.cloud/s/p8ljr/oeGA5j  This file will be updated constantly.

PR #265, opened 2018-10-31
Styling of the 3. step of the ordering process

jswk:


jswk:
@roksanaer ok, so a bigger margin below the first paragraph is put.

I also incorporated the affiliation info.

About the "CHANGE" buttons. Indeed, the one above the offer should lead to the offer selection step. I changed it appropriately.
I don't think navigating to a specific part of step 2 makes sense? There aren't any parts really (at least in master).

roksanaer:
you introduced changes after probabluy merging with the master.
unfortunatelly there is a bug *https://app.zenhub.com/workspace/o/cyfronet-fid/marketplace/issues/267) so I'm unable to check the changes

jswk:
@roksanaer I have commited, but by amending the existing commit

abacz:
> About the "CHANGE" buttons. Indeed, the one above the offer should lead to the offer selection step. I changed it appropriately.
> I don't think navigating to a specific part of step 2 makes sense? There aren't any parts really (at least in master).

I agree with @jswk. After we've moved configuration to second step I propose to remove "change" link in summary.

app/javascript/stylesheets/_order.scss
```diff
@@ -1,3 +1,105 @@
...
+    }
+
+    &.active {
+      color: $white;
+      background: $color-1;
+    }
+  }
+}
+
+.order-container {
```
mkasztelnik: I don't believe this is a good idea. Such elements as `h*` should have styles defined globally. If globals are wrong then we should improve globals not to create order specific elements.

/cc @abacz

app/views/services/summaries/show.html.haml
```diff
@@ -6,16 +6,52 @@
...
+  Technical configuration
+  = link_to "Change", service_offers_path(@service)
+
+.card.offer-description
+  .card-body
+    %h4.card-title= @project_item.offer.name
+    .card-text= @project_item.offer.description
+
+    %h5 Technical parameters
+    %table.table
```
mkasztelnik: Please don't add fake data. If `attributes` should be rendered here add TODO. Otherwise, we will have a wrong assumption that something is working...

app/views/services/summaries/show.html.haml
```diff
@@ -6,16 +6,52 @@
...
+
+  %h4 My projects
+  %dl
+    %dt Project name
+    %dd= @project_item.project.name
+
+  %h4 Usage
+  %dl
+    %dt Affiliation
+    %dd= @project_item.affiliation&.organization || "No affiliation"
```
mkasztelnik: "No affiliation selected" ?

app/views/services/summaries/show.html.haml
```diff
@@ -6,16 +6,52 @@
...
+
+    %dt Customer typology
+    %dd= @project_item.customer_typology
+
+    %dt Reason to request access to #{@service.title}
+    %dd= @project_item.access_reason
+
+    %dt Additional information
+    %dd= @project_item.additional_information
```
mkasztelnik: On the "prefka" there is information about the supervisor. Such information can be taken from `affiliation` (if selected). @roksanaer do I understand correctly purpose of this affiliation field?

app/views/layouts/_order_nav.html.haml
```diff
@@ -1,12 +1,12 @@
-%ul.nav.order-nav
+%ul.nav.nav-fill.order-nav
   %li.nav-item
     = link_to "Offer selection", service_offers_path(service),
       class: "nav-link #{"active" if step == 1}"
   %li.nav-item
     = link_to_if step > 1, "Service configuration", service_configuration_path(service),
-      class: "nav-link #{"active" if step == 2} #{"disabled" if step < 2}" do
+      class: "nav-link #{"active" if step == 2}" do
```
mkasztelnik: I don't like this (cc @roksanaer, @abacz). So right now I can click second, the third tab even if an offer is not selected. This will result in error. I believe that steps should be filled in concrete order and if I'm in step 1 then I should go through step 2 to step 3 not directly to step 3. 

If the user should be able to go directly from `step 1` to `step 3` than additional logic should be added to the controller checking if this is possible, but this is something which should be done in separate PR. Here only styling :smile:

app/views/layouts/order.html.haml
```diff
@@ -10,12 +10,12 @@
...
           = yield :next_button
       = render "layouts/flash"
-    %main.container
+    %main.container.order-container
       %h1
         = @service.title
       %p.summary= @service.tagline
       = yield
-      %hr
+      %hr.bottom-hr
```
mkasztelnik: @abacz  should it be defined globally?

app/views/services/summaries/show.html.haml
```diff
@@ -6,16 +6,52 @@
...
-  %li Project: #{@project_item.project.name}
-  %li Customer Typology: #{@project_item.customer_typology}
-  %li Reason to request access to the #{@service.title}:
-  %p= @project_item.access_reason
-  %li Additional informations:
-  %p= @project_item.additional_information
-
-= link_to "Edit your service request", service_configuration_path(@service)
+%h3
+  Technical configuration
```
mkasztelnik: This is tricky since technical configuration can be connected with both `offer selection` and `configuration` (attributes can be properties or parameters).

app/views/layouts/order.html.haml
jswk: %hr is too general element to style it globally. .bottom-hr seems like a good tradeoff

app/javascript/stylesheets/_order.scss
```diff
@@ -1,3 +1,105 @@
...
+  h2 + p {
```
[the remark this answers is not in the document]
jswk: Okay, that's neat. I'll fix that

app/javascript/stylesheets/_order.scss
jswk: %h* have some global styles, but they don't match the hierarchy here. (Even here, h structure is different in card and below.) So the namespaced styles are a good idea. In this way they don't pollute other components.
When we find other places where the same hierarchy may be used, ok. But for now I think this is the correct approach

app/views/services/summaries/show.html.haml
jswk: kk

app/views/layouts/_order_nav.html.haml
jswk: You cannot click the 2nd and the 3rd when in the first step.
The handling of the disabled is done by using link_to_if. There is a condition for rendering the link (step > 1). If it isn't fulfilled the markup from the line below is rendered, where the disabled class is set.

app/views/services/summaries/show.html.haml
jswk: So what's the suggestion?

app/views/services/summaries/show.html.haml
jswk: Currently there is no such information in Affiliation model though

app/javascript/stylesheets/_order.scss
mkasztelnik: I'm not sure if global `h*` are defined correctly right now. @abacz can you check it?

app/views/services/summaries/show.html.haml
roksanaer: not sure, why we are invoking the supervisor here. it is be a part of the affiliation.

additionally, @abacz did not add it to the prefka, but:
-on the 2nd step user chooses the affiliation he wants to refer to the order
- on the 3rd step, in the summary, all fields from the chosen affiliation should be visible (I'm not sure if it works this way now as i still cannot add affiliation to the order)

has @wziajka checked what is the problem?

Issue #243, opened 2018-10-24
As a user I can see scientific disciplines that the service is supporting 


roksanaer:
according to: 
https://kcjh3g.axshare.com/#g=1&p=service-detail_service_with_options
and
https://kcjh3g.axshare.com/#g=1&p=services-list_thematic

next to the info about the provider, information about scientific discipline should be visible (Area in the mock-up) both in the service entry view and in the list view. One service can have multiple disciplines connected.
List of the connections can be found here:
https://docs.google.com/spreadsheets/d/1tQAHG-Hk8IhmSXQaR5BULiZobx1UNtfLUpK9CMqMeuE/edit?ts=5bcd5fee#gid=1123748907
(service description content)

- List of respected Areas:
- Formal Sciences 
- Physical Sciences:
   Physics 
   Chemistry
   Earth science 
   Astronomy 
- Life Sciences / Biology 
- Social Sciences 
- Applied Sciences / Engineering / Healthcare 
- Philosophy / History 
- Interdisciplinary 
- Other (Please provide a scientific disciplines that your service supports)

Additionally, the user should be able to search for services by the scientific discipline, analogously as in the provider's field case.
I am not sure if this one will be fixed with @michal-szostak implementation of filtering or we need some additional dev here.

Eventually, the ​user should be able to click on the scientific discipline (from the service entry view or the list view) and be given a list of services supporting this discipline (I assume it's the same mechanism as with the providers)

mkasztelnik:
Already done

PR #271, opened 2018-11-02
[#243] Display scientific disciplines on service view

martaswiatkowska:
Scientific disciplines are displayed on services list and service description view.
It's link which acts as a filter. After click on user are redirected to filtered services list by scientific disciplines.
For now added random areas in seed.rb and dev.rake 
It's need to add this to  @michal-szostak filter feature


app/controllers/areas_controller.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class AreasController < ApplicationController
```
mkasztelnik: Are we really want to create a dedicated path every time additional query condition appears? This had a sense for categories but I have mixed fillings that we should do something similar for every query param. What if I want to see a selected area in a selected category?

app/controllers/areas_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  include Service::Searchable
+  include Paginable
+
+  before_action :area
```
mkasztelnik: This before action method does not have sense since it is invoked only in one place. You can move line 16 into method `show` and the result will be the same.

app/models/area.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Area < ApplicationRecord
```
mkasztelnik: Maybe better naming will be `ResearchArea`?

app/models/service.rb
```diff
@@ -31,6 +34,7 @@ class Service < ApplicationRecord
              class_name: "User",
              optional: true
   belongs_to :provider, optional: true
+  belongs_to :area, optional: true
```
mkasztelnik: Why do we need two types of relations (`1-n` and `n-n`)? If the default area is needed something similar mechanism like in the service main category can be used

app/models/service_area.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ServiceArea < ApplicationRecord
+  belongs_to :service
+  belongs_to :area
+
+  validates :service, presence: true
+  validates :area, presence: true
```
mkasztelnik: Couple `service` and `area` should be unique.

app/views/backoffice/services/_form.html.haml
```diff
@@ -21,7 +21,10 @@
   = f.input :activate_message
   = f.input :logo
   = f.association :provider
-  = f.input :contact_emails, as: :array, multiple: true
+  = f.input :contact_emails, multiple: true, as: :array
+  = f.association :areas, include_blank: false,
+            input_html: { multiple: true }, include_hidden: false
```
mkasztelnik: What `include_hidden` does?

app/views/services/_header.html.haml
```diff
@@ -14,9 +14,11 @@
             Provided by
             = link_to service.provider.name, service.provider
           .col
-            %code TODO
             Area
-            %strong Biology
+            - service.areas.each do |area|
```
mkasztelnik: Since areas display is the same in 2 places it can be extracted into partial and use in two places instead of duplicating code (see `app/views/services_service.html.haml`)

db/migrate/20181029141125_create_service_areas.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateServiceAreas < ActiveRecord::Migration[5.2]
+  def change
+    create_table :service_areas do |t|
+      t.belongs_to :service, index: true
+      t.belongs_to :area, index: true
```
mkasztelnik: You should add a unique constraint to `service` and `area`. Take a look [here](https://github.com/cyfronet-fid/marketplace/blob/master/db/migrate/20181024093203_create_service_relationships.rb#L12) how to define such constraint.

db/seeds.rb
```diff
@@ -21,10 +21,16 @@
   puts "Generated category #{ hash["name"] }"
 end
 
+yaml_hash["areas"].each do |_, hash|
+  Area.find_or_create_by(name: hash["name"])
+  puts "Generated  areass #{ hash["name"] }"
```
mkasztelnik: `#{hash["name"]} area generated` ?

db/seeds.rb
```diff
@@ -43,8 +49,10 @@
                     dedicated_for: hash["dedicated_for"],
                     restrictions: hash["restrictions"],
                     phase: hash["phase"],
-                    categories: [Category.find_by(name: hash["parent"])])
+                    categories: [Category.find_by(name: hash["parent"])],
+                    areas: [Area.all.sample])
```
mkasztelnik: Load all areas in line 14 and assign it into a variable. Then use this variable with `sample` to get sample element from the preloaded array. Using such a solution will eliminate additional not needed db queries.

lib/tasks/dev.rake
```diff
@@ -31,11 +31,14 @@ if Rails.env.development?
                        helpdesk_url: Faker::Internet.url,
                        tutorial_url: Faker::Internet.url,
                        restrictions: Faker::Lorem.sentence,
-                       phase: Faker::Lorem.sentence)
+                       phase: Faker::Lorem.sentence,
+                       areas: [Area.all.sample])
+        service.areas << Area.all.sample
```
mkasztelnik: Similar like above

spec/factories/services.rb
```diff
@@ -26,5 +26,6 @@
     sequence(:dedicated_for) { |n| "service #{n} dedicated for" }
     sequence(:restrictions) { |n| "service #{n} restrictions" }
     sequence(:phase) { |n| "service #{n} phase" }
+    sequence(:areas) { |n| [create(:area)] }
```
mkasztelnik: use simple `area` to create a related area. Something like [here](https://github.com/cyfronet-fid/marketplace/blob/master/spec/factories/affiliations.rb#L11)

mkasztelnik:
Also changelog entry is missing

db/seeds.rb
martaswiatkowska: no it's from db/data.yml

spec/factories/services.rb
martaswiatkowska: ok but why I get error  NoMethodError: undefined method `research_area=' ?

mkasztelnik: Ok I see now, there is `[]` inside. This changes a lot. So the construction is ok but there is another issue: does this related object is always necessary for all tests? Please remember that this line will add additional 2 db inserts (one for are, second for `service_research_ares` table).

martaswiatkowska: This attribute is required so yes I think that

app/views/backoffice/services/_form.html.haml
martaswiatkowska: It's eliminate empty element "".

app/controllers/services_controller.rb
```diff
@@ -6,7 +6,12 @@ class ServicesController < ApplicationController
   include Paginable
 
   def index
-    @services = paginate(records.order(ordering))
+    if params[:research_area_id]
+      @services = paginate(Service.joins(:service_research_areas).
+              where(service_research_areas: { research_area_id: params[:research_area_id] }))
```
mkasztelnik: Additional query params should be added to [`Service::Searchable` concern](https://github.com/cyfronet-fid/marketplace/blob/master/app/controllers/concerns/service/searchable.rb). All search logic connected with service should be placed there. This concert in used in `ServicesController` and `CategoriesController`

db/migrate/20181029141125_create_service_research_areas.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateServiceResearchAreas < ActiveRecord::Migration[5.2]
+  def change
+    create_table :service_research_areas do |t|
+      t.belongs_to :service, index: true
+      t.belongs_to :research_area, index: true
```
mkasztelnik: Please add also `foreign_key: true` into `servcie` and `research_area` `belongs_to` relation

lib/tasks/dev.rake
```diff
@@ -8,9 +8,10 @@ if Rails.env.development?
     task :prime, [:services_size] => "db:setup" do |task, args|
       users = User.all + [nil]
       services_size = args.fetch(:services_size, 100).to_i
+      areas = ResearchArea.all
       puts "Generating #{services_size} new services"
       services_size.times do
-        Service.create(title: Faker::Lorem.sentence,
+        service = Service.create(title: Faker::Lorem.sentence,
```
mkasztelnik: `service` variable is not used.

app/models/service.rb
```diff
@@ -12,6 +12,9 @@ class Service < ApplicationRecord
   has_many :service_categories, dependent: :destroy
   has_many :categories, through: :service_categories
   has_many :service_opinions, through: :project_items
+  has_many :service_research_areas, dependent: :destroy
+  has_many :research_areas, through: :service_research_areas
+  accepts_nested_attributes_for :research_areas
```
mkasztelnik: Why this is needed?

app/models/research_area.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ResearchArea < ApplicationRecord
+  has_many :service_research_areas, autosave: true, dependent: :destroy
+  has_many :services, through: :service_research_areas
+  accepts_nested_attributes_for :services
```
mkasztelnik: Why this is needed?

mkasztelnik:
Bug:
  * open `Galaxy` service
  * click on `Life Science` research area

As a conclusion, you are redirected into: `http://localhost:5000/services?research_area_id=5`, it does not trigger `filter_research_area` invocation, because it is registered as `research_area`. Unfortunately, after changing link to `http://localhost:5000/services?research_area=5` following error is returned:

bug

This is not your bug but bug introduced by @michal-szostak in #189 (and replicated in line 33 :smile:). I propose to fix it and improve code by changing `services_filtered_by_fields` into:

```ruby
def services_filtered_by_fields
  @@searchable_fields.
    select { |field| params[field].present? }.
    inject(scope) { |filtered, field| filter_by_field(filtered, field) }
end
```

Thanks to this simple change only filter methods with existing query param value will be invoked. This will allow to remove all:

```ruby
if search_value.empty?
  return services
end
```

mkasztelnik:
What is more I propose to use area name (maybe `slug`?) instead of id. This is more usefriendly than `id`. This can be done in separate PR.

PR #282, opened 2018-11-05
[#280] Add search by "dedicated_for"

michal-szostak:
# What is in this PR

Added filter for dedicated_for (backend & UI)

Fixes: #280

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -15,6 +15,10 @@ module Service::Searchable
       services
     end
 
+    def filter_dedicated_for(services, search_value)
+      services.where(dedicated_for: [search_value])
+    end
+
     def filter_rating(services, search_value)
       services.where("rating >= ?", search_value)
     end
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/service/searchable.rb
```diff
@@ -56,6 +54,9 @@ def records
     end
 
     def filter_by_field(elements, field)
+      if params[field].blank?
```
mkasztelnik: Take a look here https://github.com/cyfronet-fid/marketplace/pull/271#issuecomment-436154367 for a little bit different approach to fix this issue.

app/views/services/_filters.html.haml
```diff
@@ -16,6 +16,20 @@
...
 
+  %p.font-weight-bold.mt-3 Dedicated for
+  .row
+    .col-md-10.offset-md-1
+      .form-group
+        %select#dedicated-for.form-control.form-control-sm{ name: "dedicated_for" }
+          %option{ value: "", selected: "selected" }
+            Any
+          = options_for_select([["Researchers", "Researchers"],
+                                ["VO", "VO"],
```
mkasztelnik: Please move values into the controller (next it will be easier to extract it into a separate model).

db/schema.rb
```diff
@@ -79,6 +79,29 @@
     t.index ["service_id"], name: "index_offers_on_service_id"
   end
 
+  create_table "order_changes", force: :cascade do |t|
```
mkasztelnik: How this is related to this PR?

michal-szostak: I didn't add it, it might be something after rebasing

mkasztelnik:
And of course changelog entry is missing :)

PR #283, opened 2018-11-05
Fix navbar in smaller browser windows

jswk:
The dropdowns are hidden when toggler is clicked and expanded on header
click (in screen widths < lg). "Services & Resources" dropdown is
always expanded.

### DIFF
--- a/app/javascript/stylesheets/_header.scss
+++ b/app/javascript/stylesheets/_header.scss
@@ -35,43 +35,61 @@ header {
     display: none;
   }
 
-  .nav-link {
-    position: relative;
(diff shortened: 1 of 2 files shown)


jswk:
I didn't implement the underline for root categories as it didn't seem necessary in the hamburger view. @abacz what do you think?

abacz:
It is good to show user where he is in navigation even it is root category not specific site.

jswk:
I added chevrons which indicate root categories. Moreover, expanded submenu is indented to align more or less with with the chevron.

PR #289, opened 2018-11-07
Clean filters views

mkasztelnik:
Service categorization (such as provided by, research areas, dedicated for) was extracted into separate partial and reused in services `index` and `show` views. `/providers/:id` was removed because it duplicated
functionality of `/services?provider=:id`.


kmarszalek:
approve :D

PR #291, opened 2018-11-07
Bootstrap improvements towards compliance with guidelines

jswk:
- Instead of overriding classes like h* use bootstrap variables.
- Add styles for .btn-primary/secondary/error/info styled after
  guidelines.
- Fix spacing around "Next" button in header when ordering a service.
- Correct capitalization of .btn-primary in services sidebar filters.


jswk:
I have used bootstrap provided variables where possible, cleaned up the font-weight issues, switched to using Ubuntu as well. (Teko files are pruned.)

@mkasztelnik The input-group in the second step has been corrected. Other selects should also have added class 'custom-select' though. But it can be done in another PR.

If nothing major remains to be fixed, I'd like to merge this as soon as possible for other to be able to continue using the new components (buttons, mostly).

PR #294, opened 2018-11-07
Show alert for logged in user when no active affiliation

mkasztelnik:
Alert is shown for logged in user when there is no active affiliation

alert

After this is deployed and there are active affiliations count cache should be updated by invoking this command (/cc @wziajka):

```
rails active_affiliation_counter
```

### DIFF
--- a/spec/features/alerts_spec.rb
+++ b/spec/features/alerts_spec.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Affiliations" do
+  include OmniauthHelper
+
+  let(:user) { create(:user) }
+
+
+  context "as logged in user" do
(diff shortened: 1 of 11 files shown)


spec/features/alerts_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+require "rails_helper"
+
+RSpec.feature "Affiliations" do
+  include OmniauthHelper
+
+  let(:user) { create(:user) }
+
+
+  context "as logged in user" do
```
kmarszalek: what about user with affiliation?

kmarszalek: ok, you check active_affiliation? not the whole condition for this scenario. its enough

mkasztelnik: I added additional spec which checks that alert is not shown for logged in user.

PR #295, opened 2018-11-07
Order congratulations view styles

### LINKED ISSUE #235: As a user, after issuing an order I can see the congratulation panel (after order)
https://kcjh3g.axshare.com/#g=1&p=after_order

elements:
- search bar
- congratulation info
- link to 'my services'
- related services (see #223)

Clicking on one of the related services navigated to the service entry page

mkasztelnik:
gucio

Fixes #235

### DIFF
--- a/app/views/services/summaries/confirmation.html.haml
+++ b/app/views/services/summaries/confirmation.html.haml
@@ -1,5 +1,16 @@
-Huray, you ordered this service!!!
+- breadcrumb :congratulations, @project_item
 
-= link_to "Go to your service request", project_item_path(@project_item)
+.container
+  %h1 Congratulations!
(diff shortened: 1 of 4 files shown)


app/views/services/summaries/confirmation.html.haml
```diff
@@ -1,5 +1,16 @@
-Huray, you ordered this service!!!
+- breadcrumb :congratulations, @project_item
 
-= link_to "Go to your service request", project_item_path(@project_item)
```
kmarszalek: you didnt run tests... there is test that check if there is "go to you service request"... ;(

mkasztelnik: Fixed

PR #296, opened 2018-11-08
Added consent view after affiliation creation

martaswiatkowska:
Added last step with consent affiliation creation


app/views/profiles/affiliations/consent.html.haml
```diff
@@ -0,0 +1,10 @@
+.container
+  %h1 New affiliation - last step
+  %p
+    We've already sent confirmation link to #{@affiliation.email} please
+    stand up and make coffee an confirmation linksit amet, consectetur
```
kmarszalek: what is this text?

PR #299, opened 2018-11-08
Improve affiliations and projects views

jswk:
- Remove placeholders from affiliation list view and replace them with
  actual field values.
  Correct affiliations view in line with prefkas.
  Don't use .btn-group in affiliation edit view's footer.
- Redefine .display-* with variables rather than selector redefinitions.
- Try to switch to using Bootstrap for breadcrumbs, but gretel doesn't
  support Bootstrap4 yet.


mkasztelnik:
> @jswk it's a minor but can we add some info that in order to have an active affiliation the users needs to open an email and click the link in it? @abacz any suggestions where it should appear?

This is work in progress by @martaswiatkowska in the scope of https://github.com/cyfronet-fid/marketplace/pull/296

PR #304, opened 2018-11-09
Hierarchic categories count

mkasztelnik:
Previously only services directly assigned to the category was taken while calculating services count. After this change services directly assigned to the category and all its subcategories are taken into account.

### DIFF
--- a/app/models/service_category.rb
+++ b/app/models/service_category.rb
@@ -2,7 +2,12 @@
 
 class ServiceCategory < ApplicationRecord
   belongs_to :service
-  belongs_to :category, counter_cache: :services_count
+  belongs_to :category
+  counter_culture :category,
+                  column_name: "services_count",
+                  foreign_key_values: ->(category_id) do
+                    [category_id, *Category.find(category_id).ancestor_ids]
+                  end
 
   validates :service, presence: true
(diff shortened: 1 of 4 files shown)


lib/tasks/counter_cache.rake
```diff
@@ -1,13 +1,5 @@
 # frozen_string_literal: true
 
-desc "Counter cache for categories has many services"
```
jswk: Why remove this?

app/models/service_category.rb
```diff
@@ -16,4 +21,8 @@ def guarantee_only_one_main
                       where.not(id: id).
                       update(main: false)
     end
+
+    def affected_categories(category_id)
```
jswk: Is this used anywhere?

mkasztelnik: This was required only once when categories counter cache were added for existing categories and services. Since we are (and will till 23th) populating db with seeds we don't need it.

lib/tasks/counter_cache.rake
mkasztelnik: This was required only once when categories counter cache were added for existing categories and services. Since we are (and will till 23th) populating db with seeds we don't need it.

PR #307, opened 2018-11-09
Fix affiliation fields in profile view

kmarszalek:


app/models/affiliation.rb
```diff
@@ -37,6 +37,16 @@ def to_param
     iid.to_s
   end
 
+  def webpage
+    webpage = read_attribute(:webpage).to_s
+    url_starts_with_protocol?(webpage) ? webpage : "http://#{webpage}"
```
mkasztelnik: you can extract this whole line into method which will take attribute name as param and return full path. Something like:

```ruby
def full_path(url_param_name)
  value = read_attribute(:url_param_name).to_s
  url_starts_with_protocol?(value) ? value : "http://#{value}"
```

Additionally, what will happen if in line 41 attribute value will be `nil`? This should be taken into account as well.

app/models/affiliation.rb
```diff
@@ -57,4 +67,8 @@ def email_in_webpage_domain?
     def email_domain_regexp
       email.downcase.split("@").last.sub(".", "\.") unless email.blank?
     end
+
+    def url_starts_with_protocol?(url)
+      (url.blank? || url.start_with?("http" || "https"))
```
mkasztelnik: I see you hide `nil` checking here. For me, it is a wrong place to check such condition.

app/views/profiles/affiliations/_affiliation.html.haml
```diff
@@ -43,8 +43,4 @@
         .col-md-3
           %p.small.mb-0
             Departmental web page:
-          %a{ href: affiliation.webpage }
-            #{affiliation.webpage}
-
-
-
+          = link_to(affiliation.webpage, affiliation.webpage)
```
mkasztelnik: It seems like supervisor profile link is not shown. Should we remove this field?

kmarszalek: there is new view. supervisor_profile link is shown by making link from his name

PR #311, opened 2018-11-09
Attributes impl

bwilk:
### General idea 
Each attribute type is represented by a class inherited from `Attribute`. This class have two roles - it allows the user (someone who creates an offer) to specify configuration for attribute (e.i. how the attribute is presented and how it should act after specifying a value by the user that configures a service during order process), as well as to store the attribute value. Each attribute type have it's own methods of validating values against this configuration. By this design, an attribute can be a representation of both `parameter` and `property` at the same time.

As an example select attribute needs to check whether the value is one of the values specified in configuration.

Formats of `config` and `value`, are validated by respective JSON schemas in `config_valid?`and `value_valid?` methods. In order to handle a new attribute type you may choose to create your own implementation of validation methods for `config` or `value` validations or simply overload `config_schema` or `value_schema` (if it's enough for your case). 

Please review sample implementation of first attribute types and as well as `attributes_spec.rb` in order to get familiar with this design.

Each attribute type needs to have it's own `TYPE` constant value in order to handle the process of parsing them from JSON. Why JSON? Have a look into database schema for `Offer` and `ProjectItem` to find out. Both models have fields that represent an array of attributes stored by PostgreSQL in `JSONB` type. This allows us to have all the different information about attributes connected with a single `Offer` or `ProjectItem` stored in one place, and handle the storage formats variation between different attribute types. `JSONB` type allows for efficient searching within the JSON structure of the entities stored in the database - that's the second reason of this choice.

Each attribute type is handled differently by frontend. 
- for form input check example in `app/views/services/configurations/attributes/_multiselect.html.haml`
- for order summary display check example  `app/views/project_items/attributes/_multiselect.html.haml`

### General state 
The implementation is not perfect and will be enhanced and refactored as soon as possible where it's needed. Technical things:  JSON schema on backend, Simple Form on frontend (this is actually pretty difficult to work with deep customization of inputs in this gem so writing custom front components from scratch is a viable path in the long term).

This PR contains backend and frontend for the user that specifies an order and nothing more. There is a different story for creating attributes while specifying an `Offer` (the back office component) which is not included here. Although, the code is designed with this in mind. 

#### Things that need to be improved in short term:
- Tests, tests, tests! For now only attribute backend is more less tested. There are views (forms), controllers, policies, and models that need to be tested
- Generation of new attributes for an `Offer`, and id uniqueness. Currently there is no such logic. Each id is given by hand and they can overlap.   
- Handling detailed validation errors (for now we just know that attribute is not valid) 
- Extending the implementation of each attribute type (more configuration options) both on front end and backend
- New attribute types 

#### Long term plans:
- back office frontend/backend
- global representation of attributes, comparison/link between same attribute of one service and and another (maybe same ids, same types will be imposed on comparable attributes) tbd. 
- connection with categories (which is also tight with the former - global representation of attributes), attributes that are required in a given category 
- reflecting properties on offers panel  

### Known issues 
- Use correct types for configs and values. That is `17` is different than `"17"`, and if you want integer use `value_type` of integer not string. There is a test for this - currently wrapped with comment tag. Wrong types can cause failure as Rails automatically coerces types and validation of string type is performed on already coerced integer value in a given case. This needs to be fixed but if we are careful with specification of attributes nothing can go wrong 😝   

### Frontend - Configuration

zrzut ekranu 2018-11-09 o 12 35 57
### Frontend - Summary  
zrzut ekranu 2018-11-09 o 12 46 25


mkasztelnik:
@bwilk can you fix tests?

bwilk:
@mkasztelnik I wish I could...

PR #315, opened 2018-11-09
Deny to destroy an affiliation which has project item

jswk:


app/policies/affiliation_policy.rb
```diff
@@ -45,4 +45,8 @@ def confirmed?
     def owner?
       record.user == user
     end
+
+    def has_project_item?
+      !record.project_items.empty?
```
mkasztelnik: Maybe better use db count instead loading all `project_items`?

```ruby
record.project_items.count.positive?
```

jswk: Makes sense

PR #320, opened 2018-11-10
[#317] styling order steps

jarekzet:
Zmiany dot. checkoutu zgodnie z prewkami.


roksanaer:
cannot fully review it due to the problem with emails (cannot activater my affiliation as the mails are not being sent)
i can get to the second step and so far so good. But! @jarekzet, @abacz  can we make the select button look better?

michal-szostak:
I rebased, added CHANGELOG.md's entry, squasched and fixed commit message

Issue #287, opened 2018-11-06
Refine filtering and categories UI


michal-szostak:
UI Filter widgets should be changed to conform to the following specification:

* Provider should be a multi select
* Dedicated for should be a multi select via checkboxes
* Categories should conform to this preview: https://v81k42.axshare.com/#g=1&p=services-list_v3
   (one above and one below category should be visible, selected category should be `bold`)

Categories view should always display general filters

PR #322, opened 2018-11-11
[#287] Add Multiselect

michal-szostak:
# What is in this PR

* Style `_filters.html.haml` (sizing and positioning)
* Add multicheckbox widget
* Change providers and dedicated for to be multicheckbox widget
* Modify `searchable.rb` to work with multiselects

All changes are connected to filter sidebar which can be accessed via `/services` path

# How it looks like

screen shot 2018-11-11 at 13 01 47

Fixes: #287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Simple form checks validation if conditions (@mkasztelnik)
 - Clickable next order nav step (@mkasztelnik)
 - Affiliation last step - consent (@martaswiatkowska)
+- Multicheckbox widget to filters (providers & dedicated for) (@michal-szostak)
 
 ### Changed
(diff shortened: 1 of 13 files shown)


michal-szostak:
There is one thing which I would like to advise me on @mkasztelnik 

This piece of code works, but it calls too many sql requests - how should I aggregate it to a single query (it is possible in SQL, but I would like to keep it in Ruby)
```
  def provider_options
    Provider.joins(:services).group("providers.id").order(:name).map do |provider|
      [provider.name, provider.id, provider.services.count]
    end
  end
```

Additionally @roksanaer, @abacz - there's one question for you - when provider's name is too long it wraps, which destroys layout - what should we do about it. Personally I would hide excessive part with `...` but there might be other considerations. Any thoughts?

michal-szostak:
Alright - I've found solution to reducing number of SQL requests to one.
As for styling if @abacz is alright with this I'm giving a green light for merge.

PR #366, opened 2018-11-13
Uncomment offer attribute section

mkasztelnik:
We should be aware that there is TODO, otherwise, we will forget about it.


kmarszalek:
ticket for that TODO? what this TODO is about?

Issue #351, opened 2018-11-12
Add default usage section attribute to project


martaswiatkowska:
Project should have default: 
- reason for access
- Customer typology

which are copied into configuration step after selecting the project

mkasztelnik:
Already done

PR #380, opened 2018-11-13
[#351] Additional project attributes

martaswiatkowska:
New 2 optional fields were added into `Project` model: `reason_for_access` and `customer_typology`. When the user selects project during ordering process these fields are used to populate initial `project_item#reason_for_access` and `project_item#customer_typology` values. 

I know this implementation is not perfect and additional logic should be added to check if the user makes some modification to these fields while switching project but according to @roksanaer filling these fields always when the project is switched is good enough for now.


app/views/projects/_form.html.haml
```diff
@@ -1,4 +1,5 @@
 = simple_form_for @project, remote: true,
   html: { role: "form", id: "modal-form" } do |f|
   = f.input :name
-
+  = f.input :reason_for_access
+  = f.input :customer_typology, collection: ProjectItem.customer_typologies
```
mkasztelnik: This should also be defined in `Project` as enum similar like in `ProjectItem`

mkasztelnik: I will fix it

mkasztelnik:
@kmarszalek please give it another try :)

Issue #370, opened 2018-11-13
Error in affilation registration


agpul:
**Describe the bug**
Tiziana is reporting that EOSC-hub marketplace is experiencing issues when registering affiliations

**To Reproduce**
When she fills the affiliation and she puts her emial adress: tiziana.ferrari@egi.eu as email contact, she gets the following error:

Email does not belong to webpage domain

**Expected behavior**
I think that the error shouldnt occure, or maybe  the returned message should be different because users dont know what to do

**Screenshots**
attached![Screen Shot 2018-11-13 at 09.35.27.png](https://images.zenhubusercontent.com/5b8fbb317859ca34f6abbcb2/07c20940-2cb3-4c5c-8fae-d357fa907175)

**Additional context**

PR #384, opened 2018-11-13
[#370] Disable affiliation email domain validation

jswk:
The domain validation doesn't properly handle edge cases such as
inputting "domain.org/" (with slash), and properly others.
So disable it altogether for now.

Fixes: #370


mkasztelnik:
I propose to fix it

Marek Kasztelnik

wt., 13 lis 2018, 14:22: Jakub Sawicki <notifications@github.com>
napisał(a):

jswk:
> I propose to fix it
It was discussed during the meeting today and it will be fixed once more pressing matters are addressed.

PR #391, opened 2018-11-13
Clean services list views

mkasztelnik:
Common layout is used when browsing all services and categories.


app/views/categories/nav/_categories.html.haml
```diff
@@ -0,0 +1,8 @@
+%h5.text-uppercase.underline-light.mb-2.pb-2 Categories
```
kmarszalek: itnt it the same code that is in  app/views/categories/_categories.html.haml? maybe externalize it to one file? ;)

mkasztelnik: Good catch - `app/views/categories/_categories.html.haml` committed by mistake

PR #392, opened 2018-11-13
Revert "Uncomment offer attribute section"

goreck888:
we decided to remove for now (needed for tomorrow's meeting) the parameters from the offers on the service entry and 1st order step


goreck888:
zrobiła to Roksana. elo

Issue #376, opened 2018-11-13
Add category select in text search


michal-szostak:
Text search should have category (and subcategory) select, which automatically sets itself to selected category / subcategory. Additionally there should be "All" option.

Visually it should look as follows:

![Screen Shot 2018-11-13 at 11.23.19.png](https://images.zenhubusercontent.com/5b151ec376f98d2620745f31/42bd8167-028a-413d-9b44-1321877968a7)

PR #404, opened 2018-11-14
[#376] Add category select in text search

michal-szostak:
# What is in this PR?

This PR extends search bar to include category select: 

image

* Category select box will automatically set itself based on URL's
  category id (or lack therof)
* Changing category will change form action attribute
* All of this is achieved by some JS magic (via stimulus)

Fixes: #376

### DIFF
--- a/app/javascript/controllers/search_controller.js
+++ b/app/javascript/controllers/search_controller.js
@@ -2,15 +2,13 @@ import {Controller} from 'stimulus'
 
 export default class extends Controller {
     static targets = ['categorySelect', 'form'];
-    SERVICES_URL = "/services";
-    CATEGORIES_URL = "/categories";
-
-    initialize() {
-    }
 
     connect() {
+        this.SERVICES_URL = this.data.get("servicesPath");
+        this.CATEGORIES_URL = this.data.get("categoriesPath");
+
         this.categorySelectTarget.value = "";
-        let match = window.location.pathname.match(new RegExp(`^${this.CATEGORIES_URL}/(\\d+$)`));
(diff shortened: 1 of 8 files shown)


app/views/layouts/application.html.haml
```diff
@@ -5,7 +5,7 @@
     = render "layouts/flash"
     %header.pb-2
       = render "layouts/navbar"
-      = render "services/search", search_submit_path: @search_submit_path
+      = render "services/search", search_submit_path: @search_submit_path, categories: @root_categories
```
mkasztelnik: This does not work with subcategory. 

sub


It can be easily solved by changing this line into:

```haml
= render "services/search", search_submit_path: @search_submit_path, categories: @siblings || @root_categories
```

app/javascript/controllers/search_controller.js
```diff
@@ -0,0 +1,28 @@
+import {Controller} from 'stimulus'
+
+export default class extends Controller {
+    static targets = ['categorySelect', 'form'];
+    SERVICES_URL = "/services";
+    CATEGORIES_URL = "/categories";
```
mkasztelnik: This is minor for now: paths can be externalized into data parameter similar as shown [here](https://stimulusjs.org/handbook/working-with-external-resources):

```js
this.data.get("services_path");
this.data.get("categories_path");
```

and in haml:

```haml
= form_tag "#{search_submit_path}", method: :get, role: "search", class: "", "data-target": "search.form",
  "data-controller": "search", "data-search-services-path" => services_path,  "data-search-services-path" => categories_path, 
  "data-sync-query-form": "data-sync-query-form" do
```

agpul:
looks and works good, we can merge it

PR #409, opened 2018-11-15
[#358] Terms and condition hint have dynamic url

### LINKED ISSUE #358: on the 3rd step of an ordering path Terms and conditions of a chosen service should be underpinned under the "I agree on the terms" checkbox
the link should be taken from Terms of use attribute of a service

### LINKED ISSUE #385: documents and service restrictions are not mandatory attributes of a service
just know we got the message that there are no slas for services. so we need to make:
corporate sla
terms of use
access policies

and

service restrictions

a volountary attributes that are not shown if empty. also, if all 3 are empty all Documents section should be hidden

@martaswiatkowska it has also meaning for your task. if a service does not have terms of use, please, on the 3rd step show the same message with the same static link as it is now. 

@kmarszalek as this should be quick and have a big impact factor, please take care of it as soon as possible. thanks!

martaswiatkowska:
Terms and condition are taken from service.terms_of_use_url. 

Fix #358

### DIFF
--- a/app/controllers/services/summaries_controller.rb
+++ b/app/controllers/services/summaries_controller.rb
@@ -11,7 +11,9 @@ class Services::SummariesController < Services::ApplicationController
   end
 
   def create
-    @confirmation = Confirmation.new(confirmation_params)
+    @confirmation = Confirmation.new(confirmation_params.
+                                       merge(required: @service.terms_of_use_url))
+
     if @confirmation.valid?
       authorize(project_item_template)
 
@@ -45,6 +47,10 @@ class Services::SummariesController < Services::ApplicationController
     end
 
     def confirmation_params
-      params.fetch(:confirmation).permit(:terms_and_conditions)
+      if @service.terms_of_use_url.present?
+        params.fetch(:confirmation).permit(:terms_and_conditions)
+      else
+        {}
+      end
     end
 end
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -25,4 +25,9 @@ module ServiceHelper
   def any_present?(record, *fields)
     fields.map { |f| record.send(f) }.any? { |v| v.present? }
   end
+
+  def get_terms_and_condition_hint_text(service)
+    "You are about to order an #{service.title}. Please accept " \
+      "#{link_to service.title, service.terms_of_use_url} to preceed.".html_safe
+  end
 end
--- a/app/views/services/summaries/show_normal.html.haml
+++ b/app/views/services/summaries/show_normal.html.haml
@@ -7,9 +7,9 @@
   You can change anything now if you need.
   After clicking the order button your request will be sent to service provider.
 
-
   = simple_form_for @confirmation, url: service_summary_path(@service), html: { id: "order-form"  } do |f|
-    = f.input :terms_and_conditions, as: :boolean
+    - if @service.terms_of_use_url.present?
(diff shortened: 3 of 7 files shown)


app/helpers/service_helper.rb
```diff
@@ -29,4 +29,14 @@ def get_providers_list
   def any_present?(record, *fields)
     fields.map { |f| record.send(f) }.any? { |v| v.present? }
   end
+
+  def get_terms_and_condition_hint_text
+    "You are about to order an " + @service.title + ". Please accept ".html_safe +
```
kmarszalek: remove global

app/helpers/service_helper.rb
```diff
@@ -29,4 +29,14 @@ def get_providers_list
...
+    get_terms_and_condition_link + " to preceed."
+  end
+
+  def get_terms_and_condition_link
+    @service.terms_of_use_url ? link_to(@service.title, @service.terms_of_use_url, target: :_blank) :
```
kmarszalek: add present?

app/helpers/service_helper.rb
```diff
@@ -29,4 +29,14 @@ def get_providers_list
...
   end
+
+  def get_terms_and_condition_hint_text(service)
+    "You are about to order an " + service.title + ". Please accept ".html_safe +
+    get_terms_and_condition_link(service) + " to preceed."
+  end
+
+  def get_terms_and_condition_link(service)
+    service.terms_of_use_url.present? ? link_to(service.title, service.terms_of_use_url, target: :_blank) :
+      link_to("EOSC Terms and Conditions", "http://eoscportal.trust-itservices.com/governance/rules-participation", target: :_blank)
```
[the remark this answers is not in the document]
roksanaer: sorry @martaswiatkowska for not passing the newest version of this feature. Terms and conditions should now be shown here if theye are not specified in the service entry

app/views/services/summaries/show_normal.html.haml
```diff
@@ -7,9 +7,11 @@
   You can change anything now if you need.
   After clicking the order button your request will be sent to service provider.
 
-
   = simple_form_for @confirmation, url: service_summary_path(@service), html: { id: "order-form"  } do |f|
-    = f.input :terms_and_conditions, as: :boolean
+    - if @service.terms_of_use_url.present?
```
mkasztelnik: I propose to not render this input field at all when there is no need to accept terms and condition and deal with such case in controller.

app/controllers/services/summaries_controller.rb
```diff
@@ -11,7 +11,9 @@ def show
   end
 
   def create
-    @confirmation = Confirmation.new(confirmation_params)
+    @confirmation = Confirmation.new(confirmation_params.
+                                       merge(service_terms_of_use_url: @service.terms_of_use_url))
```
mkasztelnik: Maybe we can generalize this (since `Confirmation` is a general purpose class). I propose to rename key into `required`.

app/helpers/service_helper.rb
```diff
@@ -29,4 +29,14 @@ def get_providers_list
   def any_present?(record, *fields)
     fields.map { |f| record.send(f) }.any? { |v| v.present? }
   end
+
+  def get_terms_and_condition_hint_text(service)
+    "You are about to order an " + service.title + ". Please accept ".html_safe +
```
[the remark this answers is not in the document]
mkasztelnik: Can you also remove `+` in favor of `"string with #{variable}"?

PR #422, opened 2018-11-16
Dedicated order views for open access services

mkasztelnik:
The open access service ordering process was changed a bit from the user point of view (from the backend point of view everything stays the same):
  * User is forced to select a project (no more default project selection). As a conclusion use is redirected into configuration view at the beginning 
  * Summary and configuration views are specific for open access (limited information is shown, different labels, etc.).
  * Navigation configuration tab has a different name for open access services

open-access-config

open-access-summary

configuration-open-access


mkasztelnik:
@roksanaer the second problem is not connected with this PR. The problem should be identified and fixed in the scope of different PR.

mkasztelnik:
@roksanaer I added dedicated cancel button title for the open access service order. After the build is complete give it another try.

mkasztelnik:
@wziajka after this is merged db should be cleaned and seeds generated.

Issue #411, opened 2018-11-15
add Related infrastructures and platforms filter to the global filters


roksanaer:
visualisation and functionality should be the same as for Providers

examples of the values:

EUDAT
EGI
WeNMR
CLARIN
D4Science
EOPillar
EC3

PR #426, opened 2018-11-17
410 related infrastructures and platforms

### LINKED ISSUE #410: add "Related infrastructures and platforms" attribute to the service
This attribute will not be shown on the entry page but used in filtering functionality. it should be a list of elements (strings). relation many - many (like in providers)

Examples:
EUDAT
EGI
WeNMR
CLARIN
D4Science
EOPillar
EC3

martaswiatkowska:
Service can have many related platforms. 
Its possible to filter services by platform. 
Platform are not displayed anywhere for now.

Fix #410, #411

### DIFF
--- a/app/views/services/_about.html.haml
+++ b/app/views/services/_about.html.haml
@@ -19,13 +19,13 @@
               %span
                 = service.languages
 
-      - if any_present?(service, :dedicated_for)
-        %h5 Dedicated for
+      - if any_present?(service, :platforms)
+        %h5 Related Infrastructures and Platforms
         %ul.list-group
-          - service.dedicated_for.each do |dedication|
+          - service.platforms.each do |platform|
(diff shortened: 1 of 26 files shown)


app/views/services/_about.html.haml
```diff
@@ -19,13 +19,13 @@
               %span
                 = service.languages
 
-      - if any_present?(service, :dedicated_for)
-        %h5 Dedicated for
+      - if any_present?(service, :platforms)
+        %h5 Related Infrastructures and Platforms
         %ul.list-group
-          - service.dedicated_for.each do |dedication|
+          - service.platforms.each do |platform|
```
mkasztelnik: Please add `Service.joins(:platforms)` in controller show view to avoid `n+1` queries.

PR #437, opened 2018-11-19
Home new layout

jarekzet:
New layout for MP home page.


roksanaer:
@abacz can we tone down a little bit this blue colour?

roksanaer:
links on the left nav bar
image
do not lead to the cateogories URLs. only all services works

abacz:
@roksanaer No, we can't. We use official color tones from CI.

PR #438, opened 2018-11-19
Header new version

abacz:


roksanaer:
under eosc portal there has to be https://eosc-portal.eu URL

jswk:
Btw, the Login link is visible twice in xs viewport width
image

PR #439, opened 2018-11-19
Research area hierarchy

### LINKED ISSUE #378: category tree for scientific areas

mkasztelnik:
A hierarchy was added to `ResearchArea` using [ancestry](https://github.com/stefankroes/ancestry) (the same used to create categories hierarchy). On the front, no breaking space was used to add name indention ( `&nbsp;`). The option group was not good enough, because you cannot have parent clickable and this is a flat structure. 

This PR is ready from the technical point of view and can be reviewed, but before merging someone (most probably @roksanaer, @agpul, @goreck888) need to introduce correct research area hierarchy. Here you can find example (which need to be removed before merge) how to do this: https://github.com/cyfronet-fid/marketplace/blob/378-research-areas-tree/db/data.yml#L150-L152

sub

TODOs:
  - [x] @mkasztelnik: implement hierarchy in `ResearchArea`
  - [ ] @roksanaer, @agpul, @goreck888 create real research area hierarchy in yaml

Fixes #378

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Show affiliation info for project item (@jswk)
 - Add a landing page for when affiliation is activated (@jswk)
 - Favicon (@michal-szostak)
+- Research area hierarchy (@mkasztelnik)
 
 ### Changed
 - Upgrade Sprockets gem to avoid CVE-2018-3760 vulnerability (@mkasztelnik)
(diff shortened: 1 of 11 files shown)


mkasztelnik:
@roksanaer, @abacz  spaces are enough or I should also add `-`? It looks like this:

subwithminus

mkasztelnik:
@roksanaer I would like to merge it sooner than later because later there will be a lot o merge conflicts. Are you planning to add hierarchy? If no then I will comment fake data and merge it without hierarchy defined in yaml (but with hierarchy support). 

What is more @abacz, @roksanaer please answer question above.

roksanaer:
> @roksanaer, @abacz spaces are enough or I should also add `-`? It looks like this:
> 
> subwithminus

i would go with a pause as in given example

roksanaer:
> @roksanaer I would like to merge it sooner than later because later there will be a lot o merge conflicts. Are you planning to add hierarchy? If no then I will comment fake data and merge it without hierarchy defined in yaml (but with hierarchy support).
> 
> What is more @abacz, @roksanaer please answer question above.

is it possible to do so, without disrupting current research area mapping?

Issue #368, opened 2018-11-13
Adjust counter in the filters to match real number of services within the category


roksanaer:
after entering a service category, filter counters should show the amount of services available in chosen category

in this example, i chose egi services in data&storage. result shows me 4 services, counter says 6.

![image.png](https://images.zenhubusercontent.com/5b0fef1989d5dc41604fb4c3/6b48889d-3568-4559-84a2-f69d522b16ac)

PR #440, opened 2018-11-19
[#368] Fix service count in filter multiselects

michal-szostak:
# What is in this PR?

* Fix `dedicated_for_options`, `provider_options`,
  `related_platform_options` to count number of services with respect
  to selected category

* create new unit test spec for `Searchable`

Fixes: #368

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -12,7 +12,7 @@ module Service::Searchable
 
     def filter_related_platforms(services, search_value)
       services.joins(:service_related_platforms).group("services.id").
-        where("service_related_platforms.platform_id IN (?)", search_value)
+          where("service_related_platforms.platform_id IN (?)", search_value)
     end
 
     def filter_location(services, search_value)
@@ -35,7 +35,7 @@ module Service::Searchable
 
     def filter_research_area(services, search_value)
       services.joins(:service_research_areas).
-              where(service_research_areas: { research_area_id: search_value })
+          where(service_research_areas: { research_area_id: search_value })
     end
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/concerns/service/searchable.rb
```diff
@@ -35,37 +35,46 @@ def filter_providers(services, search_value)
...
      ["Providers", "Providers"],
      ["Research organisations", "Research organisations"],
      ["Business", "Business"],
      ["Other", "Other"]].map do |field|
-      [field[0], field[1], Service.where("? = ANY(dedicated_for)", field[1]).count]
-    end .select do |element|
+      query = Service.joins(:categories).where("? = ANY(dedicated_for)", field[1])
+
+      unless category.nil?
+        query = query.where("categories.slug = ?", category.slug)
```
mkasztelnik: Better use category `id` instead of `slug` for join. You can also:

```ruby
query.where(categories: category)
```

app/controllers/concerns/service/searchable.rb
```diff
@@ -35,37 +35,46 @@ def filter_providers(services, search_value)
...
      ["VO", "VO"],
      ["Providers", "Providers"],
      ["Research organisations", "Research organisations"],
      ["Business", "Business"],
      ["Other", "Other"]].map do |field|
-      [field[0], field[1], Service.where("? = ANY(dedicated_for)", field[1]).count]
-    end .select do |element|
+      query = Service.joins(:categories).where("? = ANY(dedicated_for)", field[1])
+
+      unless category.nil?
```
mkasztelnik: `if category` ?

app/controllers/concerns/service/searchable.rb
```diff
@@ -35,37 +35,46 @@ def filter_providers(services, search_value)
...
 private
 
-  def dedicated_for_options
+  def dedicated_for_options(category = nil)
     [["Researchers", "Researchers"],
      ["VO", "VO"],
      ["Providers", "Providers"],
      ["Research organisations", "Research organisations"],
      ["Business", "Business"],
      ["Other", "Other"]].map do |field|
```
mkasztelnik: You can improve it by changing to

```ruby
["Other", "Other"]].map do |name, value|
```

app/controllers/concerns/service/searchable.rb
```diff
@@ -74,21 +83,29 @@ def rating_options
...
+    query.group("platforms.id")
+         .order(:name)
+         .map { |provider| [provider.name, provider.id, provider.service_count] }
   end
 
   def records
     searchable_fields.
-      select { |field| params[field].present? }.
-      inject(search_scope) { |filtered_services, field| filter_by_field(filtered_services, field) }
+        select { |field| params[field].present? }.
```
mkasztelnik: by default, all ruby code is formatted using 2 spaces, not 4.

app/controllers/concerns/service/searchable.rb
michal-szostak: I changed to `query.where("categories.id = ?", category.id)`

`query.where(categories: category)` does not work

Issue #435, opened 2018-11-19
Add element with all services in the category tree


roksanaer:
![image.png](https://images.zenhubusercontent.com/5b0fef1989d5dc41604fb4c3/fcd9da0a-4df9-4b05-81c6-c2fa5ffeaac5)

above current category tree, there should be a new element: 'All services'.
it should be visually separated from the service categories

PR #441, opened 2018-11-19
[#435] Add "All Services" link

michal-szostak:
# What is in this PR?

* Add "All Services" link with services count above categories
  navigation in sidebar

# How it looks like?

image

Fixes: #435


app/views/categories/nav/_categories.html.haml
```diff
@@ -1,3 +1,8 @@
+- active = "font-weight-bolder" unless local_assigns[:current]
+#all-services-link{ class: "mb-3 d-flex justify-content-between align-items-center #{active}" }
+  = link_to "All Services", services_path
+  %span= Service.all.count
```
mkasztelnik: We can avoid additional db query by invoking:

```ruby
category.map(&:services_count).reduce(:+)
```

michal-szostak: How would that work with subcategories and so on? I think it might break...

PR #454, opened 2018-11-19
Remove puts

mkasztelnik:
@michal-szostak please, please, please don't commit puts.


michal-szostak:
I really try, but sometimes I do overlook them.
On the other hand if ruby debugger would work anywhere as good as any other language counterpart there would be no puts at all though...

PR #456, opened 2018-11-19
platforms update

roksanaer:
updating WeNMR services entries
adding platforms to MP services

TODO: palatforms should not be mandatory. when done, remove all lines
 with *ec3, *eopillar platforms in one line (no such
 combination takes place)

### DIFF
--- a/db/data.yml
+++ b/db/data.yml
@@ -1,19 +1,45 @@
 platforms:
 
     EGI:
-        name: &egi "EGI"
+        name: &egiPlatform "EGI"
     EUDAT:
-        name: &eudat "EUDAT"
+        name: &eudatPlatform "EUDAT"
     WENMR:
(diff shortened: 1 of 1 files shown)


db/data.yml
```diff
@@ -4508,6 +4543,19 @@ relations:
...
+#    target: *b2drop
+#  powerFitToB2drop:
+#    both: true
+#    source: *powerFit
+#    target: *b2drop
+#  disVisToB2drop:
+#    both: true
+#    source: *disVis
+#    target: *b2drop
+
```
michal-szostak: I hope there is a good reason to add commented out code

roksanaer: yes! I've started adding it, but then i saw that maybe this relation will not be required. i will ask about it tomorrow. in case it should be there, this will not be lost ;)

PR #457, opened 2018-11-19
sla links fixes

roksanaer:


roksanaer:
yep... theoretically we still demand an sla from the services wanting to join eosc and for those who do not have one we should offer the eosc corporate one. but it has not be produce yet and not all services have their own so... that's what happened

Issue #442, opened 2018-11-19
Default sort order for services


michal-szostak:
Services should be ordered by name by default (both in `services_controller` and `categories_controller`

PR #469, opened 2018-11-20
[#442] Default sort order of services

michal-szostak:
# What is in this PR?

* changed default services ordering to :title :asc

Fixes: #442

### DIFF
--- a/spec/controllers/concerns/sortable_spec.rb
+++ b/spec/controllers/concerns/sortable_spec.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+class SortableFakeController < ApplicationController
+  attr_accessor :params
+
+  def initialize
+    @params = {}
+  end
+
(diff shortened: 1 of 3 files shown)


spec/controllers/concerns/sortable_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    @params = {}
+  end
+
+  include Service::Sortable
+end
+
+RSpec.describe SortableFakeController do
+  let!(:controller) { SortableFakeController.new }
+
+  context "ordering" do
```
mkasztelnik: Can be for now but it tests internal controller implementation - which I don't like. Why not create a feature test?

michal-szostak: It tests concerns, it's different.

I agree that additional feature test (more e2e) approach is good, but this makes it easy to test concerns with minimal effort.

I'll add feature test on top of it.

Issue #470, opened 2018-11-20
dedicated page after accepting the affiliation with email link


roksanaer:
dedicated page analogical to /profile/affiliations

Change header and description

Affiliation Activate (header)
You have successfully activated your $affiliationName affiliation. You can view it in your profile or go to the service list and ask for the access to EOSC services.

Buttony:
Take me to my profile (without I  understand)
Go to services list

PR #478, opened 2018-11-20
[#470] Affiliation activation landing page

jswk:
A landing page is shown to the user when she activates an affiliation.
Remove a duplicated Affiliation::Confirm test.

Fixes: #470


app/services/affiliation/confirm.rb
```diff
@@ -10,16 +10,15 @@ def initialize(user, token)
 
   def call
     @error = nil
-    affiliation = Affiliation.find_by_token(@token)
+    @affiliation = Affiliation.find_by_token(@token)
```
mkasztelnik: I really don't like this change. It changes stateless service into stateful service, what is more, internal implementation is exposed to outside word.

app/controllers/affiliation_confirmations_controller.rb
```diff
@@ -4,14 +4,6 @@ class AffiliationConfirmationsController < ApplicationController
...
-    confirmator = Affiliation::Confirm.new(current_user, params[:at])
-
-    if confirmator.call
-      redirect_to profile_path,
-                  notice: "Affiliation confirmed sucessfully"
-    else
-      redirect_to root_path,
-                  alert: confirmator.error || "Unable to confirm your afiliation"
-    end
+    @confirmator = Affiliation::Confirm.new(current_user, params[:at]).call
```
mkasztelnik: Can't we only change `root_path` into view which shows problems with affiliation activation?

app/services/affiliation/confirm.rb
mkasztelnik: If you need more information about failure than please return error status code from the service (e.g. `:ok`, `:not_found`, ...) and in the view render appropriate message (using e.g. I18n support).

jswk: OK, it makes sense. I'll change it to a stateless version

PR #479, opened 2018-11-20
[#442] Add feature test for default sort order

michal-szostak:
I have no idea how it didn't get selected for commit in PR #469 

* Add feature test for #442 (default sort order of services)

Fixes: #442

### DIFF
--- a/spec/features/service_spec.rb
+++ b/spec/features/service_spec.rb
@@ -336,3 +336,17 @@ RSpec.feature "Service filtering and sorting" do
     `pending "add test after implementing location to filtering #{__FILE__}"`
   end
 end
+
+
+RSpec.feature "Service view" do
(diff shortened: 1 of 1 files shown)


spec/features/service_spec.rb
```diff
@@ -336,3 +336,21 @@
     `pending "add test after implementing location to filtering #{__FILE__}"`
   end
 end
+
+
+RSpec.feature "Service view" do
+  scenario "should by default sort services by name, ascending" do
+    create(:service, title: "Service e")
+    create(:service, title: "Service d")
+    create(:service, title: "Service c")
```
mkasztelnik: Isn't 2 or max 3 services enough to test default order?

michal-szostak: Well, you're right, I'll change it during branch update

Issue #485, opened 2018-11-20
add Research groups on the left bar with filtering (dedicated for)


roksanaer:
in yml VO (dedicated for) has been replaced with Research groups. please add adequate element to the filtering bar

![image.png](https://images.zenhubusercontent.com/5b0fef1989d5dc41604fb4c3/55c183c9-fff4-4543-8ba0-3816e5737fbf)

PR #486, opened 2018-11-20
[#485] Replace dedicated for with Research groups

michal-szostak:
# What is in this PR?

* Replaced label 'Dedicated For' in `_filters.html.haml` with
  'Research groups'

# How it looks like?

image

Fixes: #485


michal-szostak:
Then this is not a minor thing, this requires a whole migration

PR #491, opened 2018-11-21
mailer config for images

kmarszalek:


README.md
```diff
@@ -180,6 +180,7 @@ ENV variables:
     script is added into head section)
   * `PORTAL_BASE_URL` - portal base URL used to generate footer and other static
     links to EOSC portal
+  * `NO_PROTOCOL_ROOT_URL` and `PROTOCOL_ROOT_URL` - mailer config hosts
```
mkasztelnik: If these variables should be used only for assets (which makes sense because assets can be moved into CDN) than names should be improved, e.g. `ASSET_HOST` and `ASSET_PROTOCOL`. If not then what is the difference between `ROOT_URL` and `PROTOCOL_ROOT_URL`?

mkasztelnik:
And changelog entry is missing.

config/environments/production.rb
```diff
@@ -100,6 +100,9 @@
   config.action_mailer.default_url_options = { host: ENV["ROOT_URL"] }
   config.action_mailer.delivery_method = :smtp
 
+  config.action_controller.asset_host = "#{ENV["ASSET_HOST"]}"
+  config.action_mailer.asset_host = "#{ENV["ASSET_PROTOCOL"]}#{ENV["ASSET_HOST"]}"
```
mkasztelnik: Maybe better:

```ruby
"#{ENV["ASSET_PROTOCOL"]}://#{ENV["ASSET_HOST"]}"
```

README.md
```diff
@@ -180,6 +180,7 @@ ENV variables:
     script is added into head section)
   * `PORTAL_BASE_URL` - portal base URL used to generate footer and other static
     links to EOSC portal
+  * `ASSET_HOST` and `ASSET_PROTOCOL` - assets mailer config
```
mkasztelnik: Please give some example of variables values and mention if these variables are mandatory or optional.

config/environments/production.rb
```diff
@@ -100,6 +100,9 @@
   config.action_mailer.default_url_options = { host: ENV["ROOT_URL"] }
   config.action_mailer.delivery_method = :smtp
 
+  config.action_controller.asset_host = "#{ENV["ASSET_HOST"]}"
```
mkasztelnik: Please use fallback solution - or throw errors when these variables are not set. Otherwise, this can lead to weird bugs, hard to hunt.

PR #505, opened 2018-11-21
[#277] dedicated for is saved in db

### LINKED ISSUE #277: Change field dedicated_for to DB table
Actually, field dedicated_for is changed by me to an array of strings.
But in fact, it's an enum and the values of it are in https://docs.google.com/spreadsheets/d/1tQAHG-Hk8IhmSXQaR5BULiZobx1UNtfLUpK9CMqMeuE/edit?ts=5bd0b4c6#gid=1731955460
in the tab, "service description content" in cells B5:B10
For one service it should be possible to choose multiple receivers of the service.

## Solution
New DB table and many to many relation should be created connected to the `Service` Active record.
Additionally changes following things need to be updated:

* `db/data.yml` and `db/seeds.rb`
* `searchable.rb` concern - filtering dedicated for should take into account new schema

**EDIT**: (@michal-szostak) - I changed ticket description to match our current conclusion.

martaswiatkowska:
Fix #277

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -20,8 +20,9 @@ module Service::Searchable
       services
     end
 
-    def filter_dedicated_for(services, search_value)
-      services.where("ARRAY[?]::varchar[] && dedicated_for", search_value)
+    def filter_target_groups(services, search_value)
+      services.joins(:service_target_groups).
+          where("service_target_groups.target_group_id IN (?)", search_value)
     end
 
     def filter_rating(services, search_value)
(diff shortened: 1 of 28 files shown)


app/controllers/concerns/service/searchable.rb
```diff
@@ -66,26 +67,21 @@ def indented_name(name, level)
...
-      end
-      [field[0], field[1], query.count]
-    end.select do |element|
-      element[2] != 0
+  def target_groups_options(category = nil)
+    query = TargetGroup.select("target_groups.name, target_groups.id, count(service_target_groups.service_id) as service_count")
+                .joins(:categories)
+
+    unless category.nil?
+      query = query.where("categories.id = ?", category.id)
```
mkasztelnik: You can use short version:

```ruby
query = query.where("categories.id = ?", category.id) if category
```

or even:

```ruby
query = query.where(categories: category) if category
```

app/helpers/service_helper.rb
```diff
@@ -32,7 +32,7 @@ def get_terms_and_condition_hint_text(service)
   end
 
   def dedicated_for_links(service)
-    service.dedicated_for.map { |target| link_to(target, services_path(dedicated_for: target)) }
+    service.target_groups.map { |target| link_to(target.name, services_path(target_groups: target)) }
```
kmarszalek: after rebase to master you will have second method in this helper that uses dedicated_for

app/controllers/concerns/service/searchable.rb
martaswiatkowska: I think not

mkasztelnik: Yeap you can, I checked it.

PR #509, opened 2018-11-21
Add to project button now is a link

abacz:


michal-szostak:
Just one thing - in PRs put a screenshot of the feature at hand - especially if the PR / feature is connected to some UI element. It makes it much easier to understand what it's all about.

PR #510, opened 2018-11-22
Flash fade out after 5 seconds

### LINKED ISSUE #196: Flash notification appearance
Currently, we are showing notification in this place:

flash

@abacz is this a right place and styling according to you UX vision? The second question is connected to how long we should show it. Right not it is shown until user enter another page (clicks any link on the page). Here I propose to add some timeout (e.g. 5, 10 seconds).

mkasztelnik:
Fixes #196


kmarszalek:
alerts

Issue #483, opened 2018-11-20
Open access service not activated after the "order" is submitted


roksanaer:
**EDIT**: ( @michal-szostak ) This issue is directly connected to misconfigured WebHook in JIRA instance.

**ORIGINAL CONTENT**:
@wziajka managed do integrate production with JIRA - tickets are being created. however, for open access services as Haddock, project item is stil in state "created" after it is issued.

please check it is a major bug

![image.png](https://images.zenhubusercontent.com/5b0fef1989d5dc41604fb4c3/102b632a-451c-47c9-aa69-51d72182785b)

@wziajka we need to integrate dev instance as well....

mkasztelnik:
This is most probably jira configuration problem. @wziajka can you confirm that there are delayed jobs failures?

mkasztelnik:
Or maybe the problem with changing transition from created to done in jira.

michal-szostak:
@mkasztelnik - This is problem on our side, open access services should explicitly change status after creation. I'm reopening this ticket.

michal-szostak:
Ok, so it seems that this issue was a combination of misconfiguration of our application, target JIRA instance AND issues in code.

PR #525, opened 2018-11-22
[#483] Ready service JIRA Issue transition

michal-szostak:
# What is in this PR?

* Transition in ready service is now based on configured
  workflow done state id, instead of hardcoded state.

* Add check to `jira:check` rake task which checks whether jira's
  workflow has direct transition from `todo` to `done` state

Fixes: #483

### DIFF
--- a/app/models/jira/checker.rb
+++ b/app/models/jira/checker.rb
@@ -44,6 +44,7 @@ class Jira::Checker
   block_error_handling :check_add_comment!
   block_error_handling :check_delete_issue!
   block_error_handling :check_workflow!
+  block_error_handling :check_workflow_transitions!
   block_error_handling :check_webhook!
 
   def check_connection!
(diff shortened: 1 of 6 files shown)


app/models/jira/checker.rb
```diff
@@ -130,6 +131,16 @@ def check_workflow!(id)
     end
   end
 
+  def check_workflow_transitions!(issue)
+    begin
+      trs = issue.transitions.all.select { |tr| tr.to.id.to_i == client.wf_done_id }
```
mkasztelnik: :heart:

app/services/project_item/ready.rb
```diff
@@ -35,7 +35,7 @@ def ready_in_jira!
                     project: { key: client.jira_project_key },
                     issuetype: { id: client.jira_issue_type_id } })
 
-        trs = issue.transitions.all.select { |tr| tr.name == "Done" }
+        trs = issue.transitions.all.select { |tr| tr.to.id.to_i == client.wf_done_id }
```
mkasztelnik: It always will be int? Maybe other way around `client.wf_done_id.to_s`?

michal-szostak: Jira API will always give ID as a string, I guess it is so they won't go number overflow.

Issue #466, opened 2018-11-20
mapping order (project item) data to JIRA ticket attributes


roksanaer:
![image.png](https://images.zenhubusercontent.com/5b0fef1989d5dc41604fb4c3/abbb12eb-0643-486a-a2d0-f8baf7c8c5cf)

order reference should be a link to the order.

rest please for now implement as in attached screenshot. i will contact ciril if we can change the attributes in the ticket but i'm afraid now

agpul:
I enclose some additional comments from GGUS. Please take it into consideration if they are relevant or skip it if they are already done.

"The marketplace is creating not completed JIRA tickets when an order is submitted via the web interface.

For example: https://jira.eosc-hub.eu/browse/EOSCSO-167

In this ticket, there is no information about who submitted the ticket and about selected service options and attributes.

In addition, name of the ticket is not compliant with the specification available at https://docs.google.com/document/d/1qJ0Af60qCv9cF7Cm-uKklLnkn8MwHXr6iy2Hp76Sv0I/ "

PR #531, opened 2018-11-23
[#466] Mapping order (project item) data to JIRA issue

michal-szostak:
# What is in this PR?

* Add custom fields mapping for JIRA
* Add custom field checks and suggestions to checked (and
  `rake jira:check` task)

Fixes: #466

### DIFF
--- a/app/models/jira/checker.rb
+++ b/app/models/jira/checker.rb
@@ -46,6 +46,7 @@ class Jira::Checker
   block_error_handling :check_workflow!
   block_error_handling :check_workflow_transitions!
   block_error_handling :check_webhook!
+  block_error_handling :check_custom_fields!
 
   def check_connection!
     begin
@@ -141,10 +142,23 @@ class Jira::Checker
     end
   end
 
+  def check_custom_fields!
+    fields = client.Field.all
+
+    statuses = self.client.custom_fields.map do |field_name, field_id|
+      [field_name, fields.any? { |f| f.id == field_id && f.name.to_sym == field_name }]
+    end .to_h
+
+    if statuses.any? { |k, v| !v  }
+      raise CheckerCompositeError.new("CUSTOM FIELD mapping have some problems",
+                                      statuses)
+    end
+  end
+
   def check_webhook!(host)
     webhook = nil
     # noinspection RubyDeadCode
-    self.client.Webhook.all.each do |wh|
+    client.Webhook.all.each do |wh|
       if wh.attrs["url"] == (host + api_webhooks_jira_path + "?issue_id=${issue.id}")
         unless wh.enabled
           raise CheckerWarning.new("Webhook \"#{wh.name}\" is not enabled")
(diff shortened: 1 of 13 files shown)


app/models/jira/checker.rb
```diff
@@ -141,6 +142,19 @@ def check_workflow_transitions!(issue)
...
   end
 
+  def check_custom_fields!
+    fields = self.client.Field.all
+
+    statuses = self.client.custom_fields.to_a.map do |field_name, field_id|
+      [field_name, !fields.find { |f| f.id == field_id && f.name.to_sym == field_name }.nil?]
+    end .to_h
+
+    unless statuses.find { |k, v| !v  }.nil?
```
mkasztelnik: you can use `all?` or `any?` to return boolean value.

app/models/jira/client.rb
```diff
@@ -30,14 +33,96 @@ def initialize
...
+    issue = self.Issue.build
+
+    fields = { summary: "Service order, #{project_item.project.user.first_name} " +
+                       "#{project_item.project.user.last_name}, " +
+                       "#{project_item.service.title}",
+              project: { key: @jira_project_key },
+              issuetype: { id: @jira_issue_type_id },
+    }
+
+    @custom_fields.reject { |k, v| !v || v.empty? }.each do |field_name, field_id|
```
mkasztelnik: You can use `blank?` method:

```ruby
nil.blank? # true
"".blank?  # true
[].blank? # true
```

app/models/jira/client.rb
```diff
@@ -30,14 +33,96 @@ def initialize
...
+    when "CI-Surname"
+      project_item.project.user.last_name
+    when "CI-Email"
+      project_item.affiliation ? project_item.affiliation.email : ""
+    when "CI-DisplayName"
+      "#{project_item.project.user.first_name} #{project_item.project.user.last_name}"
+    when "CI-EOSC-UniqueID"
+      project_item.project.user.uid
+    when "CI-Institution"
+      project_item.affiliation ? project_item.affiliation.organization : ""
```
mkasztelnik: You can simplify it a bit:

```ruby
project_item.affiliation&.organization || ""
```

lib/jira/console_checker.rb
```diff
@@ -63,12 +63,33 @@ def show_available_issue_states
       end
     end
 
+    def show_suggested_fields_mapping(mismatched_fields)
+      fields = @checker.client.Field.all.select { |f| f.custom }
+
+      puts "SUGGESTED MAPPINGS"
+      mismatched_fields.each do |field_name|
+        suggested_field_description = if (f = fields.find { |f| f.name == field_name.to_s })
```
mkasztelnik: shadowing outer local variable - `f`

app/models/jira/client.rb
michal-szostak: that's handy

Issue #546, opened 2018-11-28
Support for SELECT options for custom fields


michal-szostak:
The `CP-CustomerTypology` in JIRA is specified as a select type field, with a few available options, these options must be provided as numerical ids.

Currently, customer typology is a string in db which is passed directly to JIRA api "as is" which causes errors. It must be remapped to the aproperiate id, which should be defined in configuration

PR #547, opened 2018-11-28
[#546] Support for SELECT options for custom fields

michal-szostak:
# What is in this PR?

* `ProjectItem::RegisterJob` did not recast exceptions, so if something went wrong sidekiq did not try to restart job, and no errors were logged, exception handles were retained, but error reraising has been added.
* It turned out that `CP-CustomerTypology` field in production JIRA's custom field scheme has "select" type, which requires very specific value handling. It has been added, with appropriate tests.
* Jobs and Services connected to JIRA issues had their tests expanded with failure scenarios

## Why there is no new check in `rake jira:check`?

As far as I know accessing field available option metadata is not possible, at least if you do not
have JIRA admin priviledges.

The issue for that has been created separately here #548 

Fixes: #546

### DIFF
--- a/app/jobs/project_item/register_job.rb
+++ b/app/jobs/project_item/register_job.rb
@@ -3,15 +3,17 @@
 class ProjectItem::RegisterJob < ApplicationJob
   queue_as :orders
 
-  rescue_from(ProjectItem::Register::JIRAIssueCreateError) do |exception|
+  rescue_from(Jira::Client::JIRAIssueCreateError) do |exception|
     # TODO: we need to define what to do when question registration in e.g.
     #       JIRA fails. Maybe we should report this problem to Sentry and
     #       do some manual intervantion?
+    raise exception
   end
 
   rescue_from(StandardError) do |exception|
     # This is general error, which should not occur, but should be
     # caught just in case
+    raise exception
   end
 
   def perform(project_item)
--- a/config/jira.yml
+++ b/config/jira.yml
@@ -31,6 +31,12 @@ default: &default
     CP-ReasonForAccess: <%= "'" + (ENV["MP_JIRA_FIELD_CP_ReasonForAccess"] || "") + "'" %>
     SO-1: <%= "'" + (ENV["MP_JIRA_FIELD_SO_1"] || "") + "'" %>
 
+    select_values:
+      CP-CustomerTypology:
+        single_user: <%= "'" + (ENV["MP_JIRA_FIELD_SELECT_VALUES_CP_CustomerTypology_single_user"] || "") + "'" %>
+        research: <%= "'" + (ENV["MP_JIRA_FIELD_SELECT_VALUES_CP_CustomerTypology_research"] || "") + "'" %>
+        private_company: <%= "'" + (ENV["MP_JIRA_FIELD_SELECT_VALUES_CP_CustomerTypology_private_company"] || "") + "'" %>
+
 development:
   <<: *default
 
@@ -63,6 +69,12 @@ test:
     CP-CustomerTypology: "CP-CustomerTypology-1"
     CP-ReasonForAccess: "CP-ReasonForAccess-1"
     SO-1: "SO-1-1"
+
+    select_values:
+      CP-CustomerTypology:
+        single_user: '20000'
+        research: '20001'
+        private_company: '20002'
 
 production:
   <<: *default
--- a/spec/jobs/project_item/ready_job_spec.rb
+++ b/spec/jobs/project_item/ready_job_spec.rb
@@ -2,26 +2,26 @@
 
 require "rails_helper"
 
-RSpec.describe ProjectItem::RegisterJob do
+RSpec.describe ProjectItem::ReadyJob do
   let(:project_item_owner) { create(:user) }
   let(:project) { create(:project, user: project_item_owner) }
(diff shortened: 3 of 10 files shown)


app/jobs/project_item/register_job.rb
```diff
@@ -3,15 +3,17 @@
 class ProjectItem::RegisterJob < ApplicationJob
   queue_as :orders
 
-  rescue_from(ProjectItem::Register::JIRAIssueCreateError) do |exception|
+  rescue_from(Jira::Client::JIRAIssueCreateError) do |exception|
     # TODO: we need to define what to do when question registration in e.g.
     #       JIRA fails. Maybe we should report this problem to Sentry and
     #       do some manual intervantion?
+    raise exception
```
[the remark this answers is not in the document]
mkasztelnik: For exceptions which will fail each time (such as misconfiguration in MP) I propose to create sentry entry with dedicated tag (e.g. `manul-intervention-needed`). Here you can find documentation how to submit such sentry message: https://docs.sentry.io/clients/ruby/. 

To use is in code you need to move `sentry-raven` gem from production to default group.

config/jira.yml
```diff
@@ -31,6 +31,12 @@ default: &default
     CP-ReasonForAccess: <%= "'" + (ENV["MP_JIRA_FIELD_CP_ReasonForAccess"] || "") + "'" %>
     SO-1: <%= "'" + (ENV["MP_JIRA_FIELD_SO_1"] || "") + "'" %>
 
+    select_values:
+      CP-CustomerTypology:
+        single_user: <%= "'" + (ENV["MP_JIRA_FIELD_SELECT_VALUES_CP_CustomerTypology_single_user"] || "") + "'" %>
+        research: <%= "'" + (ENV["MP_JIRA_FIELD_SELECT_VALUES_CP_CustomerTypology_research"] || "") + "'" %>
+        private_company: <%= "'" + (ENV["MP_JIRA_FIELD_SELECT_VALUES_CP_CustomerTypology_private_company"] || "") + "'" %>
```
mkasztelnik: This configuration becomes scary. For sure we need to refactor it and simplify it a bit in the feature.

spec/jobs/project_item/ready_job_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+require "rails_helper"
+
+RSpec.describe ProjectItem::ReadyJob do
+  let(:project_item_owner) { create(:user) }
+  let(:project) { create(:project, user: project_item_owner) }
+  let(:ready_service) { instance_double(ProjectItem::Ready) }
+  let(:project_item) {
+    project_item = create(:project_item, project: project)
+    expect(ProjectItem::Ready).to receive(:new).
```
mkasztelnik: Please consider not using `expect` while creating a variable. The expectation should be defined in test, not in variable creation.

spec/services/project_item/ready_spec.rb
```diff
@@ -7,95 +7,115 @@
   let(:issue) { double("Issue", id: 1) }
   let(:transition) { double("Transition") }
 
+  context "(JIRA works without errors)" do
```
mkasztelnik: Why `()`?

spec/services/project_item/register_spec.rb
```diff
@@ -6,42 +6,62 @@
...
-    expect(project_item).to be_registered
+      expect { described_class.new(project_item).call }.
+          to change { ActionMailer::Base.deliveries.count }.by(1)
+    end
   end
 
-  it "sent email to project_item owner" do
-    # project_item change email is sent only when there is more than 1 change
-    project_item.new_change(status: :created, message: "ProjectItem created")
+  context "(JIRA raises Errors)" do
```
mkasztelnik: why `()`?

spec/services/project_item/register_spec.rb
```diff
@@ -6,42 +6,62 @@
...
+    let!(:jira_client) do
```
mkasztelnik: why bang (`let!`) method?

michal-szostak: because  I need to make stub of Jira::Client, and make it available as a local variable for the `it` blocks.  And make it indirectly available for  `describet_class`  constructor.

This is basically `before` block +  initializing & defining the variable.

I find it much easier to read, and it prevents potential problems, where someone forgets calling `jira_client` variable and all the stubbing will fail.

michal-szostak:
As for the case of too wide PR. I created a separate issue for some functionality which could be easily separated. The rest was very connected (I fix case with errors recasting, so I need to add tests to make sure that it works). The things in this PR are pretty connected. I could technically separate them,   but the only reason which I see for it would be to make the life of the reviewer easier (which is a good point, by the way).

app/jobs/project_item/register_job.rb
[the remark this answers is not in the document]
michal-szostak: I do agree on the fundamental level. But we got a code with no sentry integration *AND* no errors, which would not even generate a single line in the logs  - no information about failed job in sidekiq, no nothing. Which is absolutely unacceptable.

I do agree - IF sentry / backoffice /  some other kind of error handling is added *then* of course recasting will have no meaning. But as  it is now,  it's better to  recast it, so error logs remain,  than to make it fail silently.

PR #565, opened 2018-12-03
[#555] Every open access service can be added to project only once 

### LINKED ISSUE #555: Open access service can be add to the same project only once
Now, when we have an open access service user can add this service twice or more to the one project. We should do some restriction. Open access service can be add to one project only once (but we can add this service to different projects).

martaswiatkowska:
Before going to the last step (summaries) service are checked if it's already added to selected project.   
If it's already added it's shown message and user are redirected to the configuration page

Fix #555

### DIFF
--- a/app/models/project_item.rb
+++ b/app/models/project_item.rb
@@ -42,12 +42,17 @@ class ProjectItem < ApplicationRecord
   validates_associated :property_values
 
   delegate :user, to: :project
-  delegate :service, to: :offer
+
+  def service
+    offer.service unless offer.nil?
+  end
 
   before_save :map_properties
 
+  validate :one_per_project?, on: :create
+
   def open_access?
-    @is_open_access ||= offer&.service&.open_access?
+    @is_open_access ||= service&.open_access?
   end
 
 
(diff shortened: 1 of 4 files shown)


app/controllers/services/summaries_controller.rb
```diff
@@ -42,6 +43,16 @@ def ensure_valid!
       end
     end
 
+    def one_per_project!
+      if @service.service_type == "open_access"
```
michal-szostak: This can be shortened to `@service. open_access?`

agpul:
Change statement: "Project You cannot add open access service DisVis to project Services twice" . @martaswiatkowska please delete word Project from the beggining.

app/models/project_item.rb
```diff
@@ -42,10 +42,15 @@ class ProjectItem < ApplicationRecord
...
+  def service
+    offer.service unless offer.nil?
+  end
 
   before_save :map_properties
 
+  validate :one_per_project?, on: :create
+
   def open_access?
     @is_open_access ||= offer&.service&.open_access?
```
michal-szostak: The `service` method is defined above, but you're using `offer&.service` it's better to pick one

app/models/project_item.rb
```diff
@@ -104,4 +109,13 @@ def property_values=(property_values)
     self.write_attribute(:property_values, @property_values)
     @property_values
   end
+
+  def one_per_project?
+    if !service.nil? && service.open_access?
```
michal-szostak: `open_access?` method has been defined, why not use it?

PR #580, opened 2018-12-10
[#560] Add service status to service

### LINKED ISSUE #560: Add service status in the MP back office
Service owner is allowed to decide wether his service is visible or accessible in the MP for the users.

Each service has status. For now the status can be:
draft
published

In the service description, service owner can decide if the service should be visible to users.

Enum set to published means that the service is visible in categories it's under and is findable with search functionalities. Enum set to draft: service in not visible or findable in the MP GUI
 but is visible and editable to the service user

for now published service is also editable

all current services should have published status

**Only 2 transitions are possible: draft -> published and published -> draft**

service transition publish -> draft makes all underpinned service offers a draft
draft -> publish service transition does not publish service offers. it would be useful to add adequate success info box:
Your service has been published! All offers connected to the service **did not** change their status. If you wish to publish your service offers, please go to the service offer panel and activate chosen offers.

martaswiatkowska:
Add status :draft and :published to services
Fix #560

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,6 +185,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Improve wording of service, affiliation and project item mails (@jswk)
 - Replace dedicated_for with target_group (@martaswiatkowska)
 - Rating to three field form service_rating, order_rating and opinion (@kmarszalek)
+- Category counter cache takes into account service status (@mkasztelnik)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,6 +153,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Possibility to setup RATE_AFTER_PERIOD ENV variable (@kmarszalek)
 - Category field on create service view (@martaswiatkowska)
 - Additional customer typologies fields for project_item and project (@martaswiatkowska)
+- Status to service (@martaswiatkowska)
 
 ### Changed
 - Upgrade Sprockets gem to avoid CVE-2018-3760 vulnerability (@mkasztelnik)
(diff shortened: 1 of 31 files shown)


roksanaer:
3 questions:
* @martaswiatkowska @wziajka can somebody please add me an admin role so I can check this feature? 
* what are our next steps for this PR?
* @mkasztelnik @martaswiatkowska @bwilk @kmarszalek can you please conduct a review?

thanks

wziajka:
@roksanaer  roles added.

mkasztelnik:
This PR has 2 issues:
1. (Easy to fix) Service owner is able to change status using backoffice edit view
2. (Hard - to investigate) Counter cache for categories is wrong when service is created with draft state and next published. This is a problem of with the situation that service can belong to many categories (because of counter cache implementation located in `ServiceCategory` model which has no notification of service change).

To solve the second problem:
1. we need to "touch" all service `ServiceCategories` each time service status is changed
2. (to investigate) we need to check how `counter_culture` discovers changes and add "virtual" `state`  method there (most probably also `state_was`).

mkasztelnik:
Problem with counter cache solved. This is ready for review.

roksanaer:
i believe i need to be added a service owner role again? @wziajka i do not see the backoffice button after logging in:

obraz

roksanaer:
published -> draft transition should be called “Stop showing in the Marketplace”

mkasztelnik:
Maybe additional "retired" state?

śr., 16 sty 2019, 11:39: roksanaer <notifications@github.com> napisał(a):

agpul:
As a service owner I can see only my services and I can't edit them and I believe it is OK, I would like to do the review as a SP Manager but I am waiting for proper role.

agpul:
In my opinion we could be able to delete draft only if the service has never been published before. If we make a draft from service which was published (at least once) we could be able to retire this service or something, but not delete without any information about this service in a database. 
@roksanaer

roksanaer:
The asaumption is good - deleting the draft is possible only in draft status. It is just the initial workflow and it will need upgrading. If test is fine - green light

Issue #581, opened 2018-12-11
Add fields connected with Customer topology on Service configuration step


agpul:
First of all we shoul have 4 instead of 3 customer topology:
1. Single user
2. Representing a research community
3. Represtenting a project 
4. Represtenting a private company

When user chooses (in Customer topology filed on Service configuration step  ): 
1. Representing a research community  - there should appear an additional field "User group name" (text field) 
2. Represtenting a project - there should be two additional fields:  "Project name" (text field)  and "Project website url" (URL)
3. Private a private company - there should be two additional fields: "Company name" (text field)  and "Company website url" (URL)

fields are mandatory
for already existing orders, please populate the new fields with: "Not provided"

bwilk:
@abacz please check how representing these fields works from the UX perspective

abacz:
Additional fields shouldn't make any trouble for users.

PR #595, opened 2019-01-03
[#581] Add customer typology connected fields

martaswiatkowska:
Additional fields are added dynamically after selecting customer typology on configuration step or project  creation view.

Fixed: #581

### DIFF
--- a/app/views/projects/_form.html.haml
+++ b/app/views/projects/_form.html.haml
@@ -1,5 +1,9 @@
 = simple_form_for @project, remote: true,
-  html: { role: "form", id: "modal-form" } do |f|
+  html: { role: "form", id: "modal-form", "data-controller" => "project-item" } do |f|
   = f.input :name
   = f.input :reason_for_access
-  = f.input :customer_typology, collection: Project.customer_typologies.keys.map(&:to_sym)
+  = f.input :customer_typology, collection: Project.customer_typologies.keys.map(&:to_sym),
+    input_html: { "data-target" => "project-item.customer",
+                  "data-action" => "change->project-item#showSelectedSection" }
+
+  = render "services/configurations/customer_typology_fields", f: f
--- a/app/views/services/configurations/show_normal.html.haml
+++ b/app/views/services/configurations/show_normal.html.haml
@@ -27,7 +27,11 @@
   %h5.mb-3.font-weight-bold.configuration-title.text-uppercase Usage
   = f.input :customer_typology,
     collection: @customer_topologies,
-    input_html: { "data-target" => "project-item.customer" }
+    input_html: { "data-target" => "project-item.customer",
+                  "data-action" => "change->project-item#showSelectedSection" }
+
+  = render "services/configurations/customer_typology_fields", f: f
+
   = f.input :access_reason, input_html: { "data-target" => "project-item.reason" }
   = f.input :additional_information
 
--- a/spec/models/project_item_spec.rb
+++ b/spec/models/project_item_spec.rb
@@ -69,4 +69,24 @@ RSpec.describe ProjectItem do
       expect(build(:project_item, status: :rejected)).to_not be_active
     end
   end
+
+  describe "research typology" do
+    subject { build(:project_item, customer_typology: "research") }
+
+    it { is_expected.to validate_presence_of(:user_group_name) }
+  end
+
+  describe "project typology" do
+    subject { build(:project_item, customer_typology: "project") }
+
+    it { is_expected.to validate_presence_of(:project_name) }
+    it { is_expected.to validate_presence_of(:project_website_url) }
+  end
+
+  describe "private_company typology" do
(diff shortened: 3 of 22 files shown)


app/views/projects/_form.html.haml
```diff
@@ -1,5 +1,9 @@
 = simple_form_for @project, remote: true,
-  html: { role: "form", id: "modal-form" } do |f|
+  html: { role: "form", id: "modal-form", "data-controller" => "project-item" } do |f|
   = f.input :name
   = f.input :reason_for_access
-  = f.input :customer_typology, collection: Project.customer_typologies.keys.map(&:to_sym)
+  = f.input :customer_typology, collection: Project.customer_typologies.keys.map(&:to_sym),
+    input_html: { "data-target" => "project-item.customer",
+                  "data-action" => "change->customer-typology#clearElements change->project-item#showSelectedSection" }
```
mkasztelnik: Please create a meaningful method in stimulus controller reflecting this action instead of invoking 2 methods (e.g. `changeCustomerTopology()`)

app/views/services/configurations/show_normal.html.haml
```diff
@@ -27,7 +27,11 @@
   %h5.mb-3.font-weight-bold.configuration-title.text-uppercase Usage
   = f.input :customer_typology,
     collection: @customer_topologies,
-    input_html: { "data-target" => "project-item.customer" }
+    input_html: { "data-target" => "project-item.customer",
+                  "data-action" => "change->customer-typology#clearElements change->project-item#showSelectedSection" }
```
mkasztelnik: Similar as above

db/migrate/20181212102330_add_representing_research_community_fields_to_project_item.rb
```diff
@@ -0,0 +1,16 @@
+class AddRepresentingResearchCommunityFieldsToProjectItem < ActiveRecord::Migration[5.2]
+  def up
+    add_column :project_items, :user_group_name, :text
+
+    ProjectItem.find_each do |item|
```
mkasztelnik: We should not invoke model specific code inside migration because during project evolution e.g. `item.research?` method can be removed/renamed and then invoking this migration will fail. Instead please use pure SQL, e.g.:

```ruby
execute("UPDATE TABLE project_items SET user_group_name = 'Not provided' WHERE customer_typology = 'project'")
```

This comment is for all migrations with a model specific code.

db/migrate/20181212105741_add_representing_project_fields_to_project_item.rb
```diff
@@ -0,0 +1,19 @@
+class AddRepresentingProjectFieldsToProjectItem < ActiveRecord::Migration[5.2]
```
mkasztelnik: Can you merge all these migrations into one?

spec/models/project_item_spec.rb
```diff
@@ -69,4 +69,24 @@
       expect(build(:project_item, status: :rejected)).to_not be_active
     end
   end
+
+  describe "if research" do
```
mkasztelnik: "research typology"?

spec/models/project_item_spec.rb
```diff
@@ -69,4 +69,24 @@
...
+    before { allow(subject).to receive(:research?) { true } }
```
mkasztelnik: You can instead create research subject instead of mocking object here:

```ruby
subject { build(:project_item, customer_typology: "research") }
```

mkasztelnik: This is valid for describe blocks bellow as well.

spec/models/project_item_spec.rb
mkasztelnik: This is valid also for other describe blocks bellow.

spec/rails_helper.rb
```diff
@@ -58,7 +58,7 @@
     Capybara::Selenium::Driver.new(app, browser: :chrome, options: options)
   end
 
-  Capybara.javascript_driver = :chrome
+  Capybara.default_max_wait_time = 3
```
mkasztelnik: Please don't remove chrome :)

roksanaer:
@agpul 
< @roksanaer shouldn't this fields be visible somwhere for user (eg. in My services as project details?)

good call. however, for this we need to add a new ticket. @martaswiatkowska would it be problematic to add it under this PR?

roksanaer:
this PR requires update in JIRA integration: adding one new field in CI-CustomerTypology custom field in JIRA and include it in MP-JIRA integration. To do so, we need:

- have a new option for the CI-CustomerTypology added in JIRA
- (best) have its ID
- include it in the MP-JIRA integration (add it to the yml config file)

roksanaer:
new option for CP-CustomerTypology was added, with ID 10250 (this ID was reported by Mal, should be the right one)

PR #605, opened 2019-01-08
[#465] Add Labels for active filters

michal-szostak:
## What is in this PR?

* Add links in 'Filters' section of the side bar which removes single
  active filter.

## How does it look like?

screen shot 2019-01-08 at 12 53 15

## Notes

I haven't worked on CSS, I would advise @abacz to look at how it looks now.

Fixes: #465


app/controllers/concerns/service/searchable.rb
```diff
@@ -165,4 +168,34 @@ def search_ids
...
+  def active_searchable_fields
+    searchable_fields.select { |field| params[field].present? }
+  end
+
+  def active_filters
+    active_searchable_fields
+        .map { |field| params[field].is_a?(Array) ? [field, params[field]] : [field, [params[field]]] }
+        .map do |field, field_params|
+      field_params.map do |p|
+        [field, self.send("options_#{field}").find do |name, id, count|
```
mkasztelnik: Each filter build is invoked twice right now. This leads to 2x db queries. Maybe it is a time to create a dedicated class to store filter data (not search results but only data allowing to show filters)? As a conclusion, this code can be moved there and a lot of code from this concern (not connected with searching - all `options_*` methods) will be moved there. This will also solve the problem with a [multiple variable assignments](https://github.com/cyfronet-fid/marketplace/blob/465-labels-for-active-filters/app/controllers/services_controller.rb#L11-L17) in `ServicesController` and `CategoriesController`. 

What do you think?

app/controllers/concerns/service/searchable.rb
```diff
@@ -15,25 +15,80 @@ def filter_related_platforms(services, search_value)
           where("service_related_platforms.platform_id IN (?)", search_value)
     end
 
+    def options_related_platforms(category = nil)
```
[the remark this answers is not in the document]
michal-szostak: Name changing and moving had to be done because they are now used in two ways  - first as it was before to generate options to display in the sidebar, second to generate options which are matched against query params to display active filters. As their names had to be changed to be uniform (so they could be called automatically.

app/controllers/concerns/service/searchable.rb
michal-szostak: Yes, I was considering this, but I thought that it could be a scope of a new ticket (refactor +  optimization). What do you think?

Besides those active filters queries are called only if a filter is active, so in *worst case* it's 2x db queries. (I'm not saying it's good though)

michal-szostak: Though as a temporary solution we can call `options_` methods only if it has not been called before, what do you think?

michal-szostak: Hmm, maybe that is a time... I thought that it might be a scope of a new ticket, but I guess, now it's  as good time as any. I should be able to refactor it before new version is  deployed.

kmarszalek:
let us know when to start review

Issue #534, opened 2018-11-27
Possibility to add service offers to backoffice


agpul:
We need to be able to add service offers to backoffice.

mkasztelnik:
Please use this issue to describe how it should look like, who should be able to perform offer creation/deletion/edition, what will happen to existing orders when an offer will be edited/deleted, etc.

martaswiatkowska:
@roksanaer do you have any arrangements?

roksanaer:
Implementation fo now

after a service offer status is added the following should be implemented:

- service owner is able to add new service offer that after setting the status flag to published is becoming visible in the MP
- service owner is able to edit a service offer, no matter the service offer status
- service offer cannot be deleted, only its status can be set to draft (for now)

Due to Bartek's implementation service orders (project items) will not be affected by the updates in the service offer definitions

roksanaer:
after @martaswiatkowska implements first set of the functionalities @abacz takes it and proposes improved views

PR #609, opened 2019-01-10
[#534] Service owner can add offer to service

martaswiatkowska:
Initial implementation of adding offers in the backoffice.

Fixed: #534


app/controllers/backoffice/services/offers_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    template = offer_template
+    authorize(template)
+
+    @offer = Offer::Create.new(template).call
+
+    if @offer.persisted?
+      redirect_to backoffice_service_path(@service),
+                  notice: "New offer has been created"
+    else
+      render :new, status: :bad_request, service_id: params[:service_id]
```
mkasztelnik: Can you explain why `service_id` param is needed here?

PR #612, opened 2019-01-10
[#593] Add custom field with project name to created issue

michal-szostak:
# What is in this PR?

* Add custom field support for `SO-ProjectName`. The field is filled as
  follows: `<project_nanme> (<project_id>)`

Fixes: #593


agpul:
I will be able to check this PR only after merge. Only docker master is connected with JIRA project.

Issue #623, opened 2019-01-14
new fileds in a service offer due to vouchers functionality


roksanaer:
2 things are needed

- way to flag offer as offer which supports vouchers functionality

- if a service is flagged as "supporting vouchers" 2 fields schould be available to the user on the UI side (see)

PR #640, opened 2019-01-22
[#623] Voucher workflow for project item

michal-szostak:
# What is in this PR?

* Offer can be marked as "voucher enabled"
* User can provide Voucher ID or request for a voucher
* Voucher data is stored with project item
* Voucher information is shown in order summary

# How to test  it (click it)

You need to manually mark offer as `voucher_enabled` in the database to be able to test this functionality.

# How does it look like?

## Configuration

screen shot 2019-01-22 at 10 02 46

screen shot 2019-01-22 at 10 02 54

## Summary

screen shot 2019-01-22 at 10 04 36

screen shot 2019-01-22 at 10 04 59

Fixes: #623


app/views/services/configurations/show_normal.html.haml
```diff
@@ -1,14 +1,26 @@
...
-%h2.mb-4.font-weight-normal= t "service.normal.order.configuration"
-%p.mb-4
-  We need more information about your request.
-  We will be able to process your request quicker
-  if the form is filled with as much detail as possible.
-
 = simple_form_for @project_item, url: service_configuration_path(@service), method: :put,
   html: { id: "order-form", "data-controller" => "project-item", "data-project-item-url" => projects_path } do |f|
 
+  - if @project_item.offer.voucher_enabled
```
mkasztelnik: Please move it into separate partial.

app/views/services/configurations/show_normal.html.haml
```diff
@@ -1,14 +1,26 @@
 = render "layouts/order/nav", service: @service, step: 2
 
-%h2.mb-4.font-weight-normal= t "service.normal.order.configuration"
-%p.mb-4
-  We need more information about your request.
```
mkasztelnik: Why this was removed?

app/views/services/summaries/show_normal.html.haml
```diff
@@ -18,6 +18,15 @@
   %h3.mb-4
     Additional information
 
+  - if @project_item.offer.voucher_enabled
+    %h3 Voucher
+    %dl.mb-4
+      - if @project_item.request_voucher
+        %dt I don't have a voucher, and I would like to request for one
```
mkasztelnik: `request one`?

app/views/services/summaries/show_normal.html.haml
```diff
@@ -18,6 +18,15 @@
...
+      - unless @project_item.voucher_id.empty?
```
mkasztelnik: `else`?

db/migrate/20190117135718_add_voucher_support_to_offer.rb
```diff
@@ -0,0 +1,5 @@
+class AddVoucherSupportToOffer < ActiveRecord::Migration[5.2]
+  def change
+    add_column :offers, :voucher_enabled, :boolean, default: false, null: false
```
mkasztelnik: Just a proposition of a different name: `vaucherable`?

spec/models/project_item_spec.rb
```diff
@@ -107,4 +107,22 @@
     it { is_expected.to validate_presence_of(:company_name) }
     it { is_expected.to validate_presence_of(:company_website_url) }
   end
+
+  describe "vouchers" do
```
mkasztelnik: Right now text for the test loos as follow:

`ProjectItem vouchers vaucher_id can not...`

Maybe better naming would be:

`ProjectItem#vaucher_id can not...`

To do it change `vouchers` in describe into `#vaucher_id` and remove `vaucher_id` from `it` sections.

app/models/project_item.rb
```diff
@@ -131,4 +135,12 @@ def one_per_project?
   def research_area_is_a_leaf
     errors.add(:research_area_id, "cannot have children") if research_area&.has_children?
   end
+
+  def voucher_support?
```
mkasztelnik: `vaucherable?`?

app/controllers/services/configurations_controller.rb
```diff
@@ -11,6 +11,11 @@ def show
 
   def update
     @project_item = ProjectItem.new(configuration_params)
+
+    if @project_item.request_voucher
```
mkasztelnik: Mabe this can be removed and logic to send or not to sent voucher value should be moved into jira client?

app/views/services/summaries/show_normal.html.haml
michal-szostak: I assume that this content will be refined by our *content people*

app/views/services/summaries/show_normal.html.haml
michal-szostak: Ok, good point

app/models/project_item.rb
michal-szostak: XD love it

app/controllers/services/configurations_controller.rb
michal-szostak: No, because we have to have consistent DB, I had a discussion about it with @bwilk.
This also generates other problems, because this value will be set by JIRA later. And displayed in order details (and displaying it there is done regardless of `request_voucher` value)

app/views/services/configurations/show_normal.html.haml
michal-szostak: By accident

michal-szostak: It was not deleted, it was moved down

michal-szostak: I thought that voucher is not  stricte part of the service configuration, but Maybe it should be returned to the  top spot

PR #647, opened 2019-01-23
575 about page

jarekzet:


app/views/layouts/_footer.html.haml
```diff
@@ -25,6 +25,8 @@
             %a{ href: "#{portal_base}/about/eosc-projects" } EOSC Projects
           %li.mb-1
             %a{ href: "#{portal_base}/about/funding-opportunities" } Funding Opportunities
+          %li.mb-1
+            %a{ href: "/pages/about" } Marketplace
```
kmarszalek: .

app/javascript/stylesheets/application.css
```diff
@@ -0,0 +1,58 @@
+/*
```
kmarszalek: .

public/application.css
```diff
@@ -0,0 +1,57 @@
+/*
```
kmarszalek: .

db/schema.rb
```diff
@@ -95,6 +95,29 @@
     t.index ["service_id"], name: "index_offers_on_service_id"
   end
 
+  create_table "order_changes", force: :cascade do |t|
```
mkasztelnik: This is not connected with this PR, please remove this change.

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class PagesController < ApplicationController
+  def show
+  end
+
+  def about
+    about_path -> { "#{root_path}/about" }
+    about_url -> { "#{root_path}/about" }
```
mkasztelnik: Can you describe line 8 and 9? I commended them and everything seems to be working.

app/helpers/pages_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module PagesHelper
```
mkasztelnik: This is empty - please remove it and add when it would be really needed.

app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -1129,3 +1129,14 @@ label[data-multicheckbox].small {
   display: none;
 }
 //scss-lint:enable QualifyingElement
+
+.simple-list {
```
mkasztelnik: Is `simple-list` is a part of bootstrap classes? If not then this is not a good place (`_bootstrap_customizations.scss`) to put such class.

mkasztelnik: If this is not a part of bootstrap than another question appears: cannot we use something which is defined in bootstrap? I'm asking because we should avoid creating page specific scsses but we should focus to create a unified set of styles for the whole webpage.

app/views/pages/about.html.haml
```diff
@@ -0,0 +1,47 @@
+!!!
+%html{ lang: "en" }
+  = render "layouts/head"
+  %body
+    %main
```
mkasztelnik: You don't need to define a whole page (with a header, metadata, etc.) - every page is using default template. To use it please remove lines from 1 to 5 and align the rest of the page.

mkasztelnik:
This view is also not scaling well, take a look at this screenshot:

about-scaling

Text should overflow image instead, there is empty space.

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class PagesController < ApplicationController
+  def show
```
mkasztelnik: This method is not used, can be removed.

mkasztelnik:
Also take a look here: https://github.com/thoughtbot/high_voltage - this can simplify static page creation.

app/helpers/pages_helper.rb
goreck888: i deleted it.

app/controllers/pages_controller.rb
goreck888: Do you mean show method? I deleted it.

app/controllers/pages_controller.rb
goreck888: You have right, there's only need to have method about. I just looked how it is done in other projects with static pages because my solution had security vulnerabilities.

db/schema.rb
goreck888: Ok, I deleted it.

app/views/pages/about.html.haml
goreck888: i deleted these lines.

app/javascript/stylesheets/_bootstrap-customizations.scss
goreck888: I know but if I could ask, let's leave it as it is now, there is need to create about page in the next release, we don't have templates for pages like that, so I write TODO comments. Could it be?

PR #671, opened 2019-01-30
[#662] Offers improvements

### LINKED ISSUE #662: As SP manager I have possibility to edit service offer text in an intuitive way
Acceptance criteria:
- existing bug (when I hit enter I don't have a new line but text is doubled) is repair
- the text input box has proper size

martaswiatkowska:
Higher offer description textarea on edit view
Offers on views (backoffice and portal) are displayed in markdown  

Fix #662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Added
 
 ### Changed
+- Offers displaying on backoffice and portal view (@martaswiatkowska)
+- Height of the textarea in the ofers form (@martaswiatkowska)
 
 ### Deprecated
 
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -131,9 +131,9 @@
     t.string "project_website_url"
     t.string "company_name"
     t.string "company_website_url"
+    t.bigint "research_area_id"
     t.boolean "request_voucher", default: false, null: false
     t.string "voucher_id", default: "", null: false
-    t.bigint "research_area_id"
     t.index ["affiliation_id"], name: "index_project_items_on_affiliation_id"
```
michal-szostak: I don’t think that change to `scheme.rb` should be in this PR

martaswiatkowska: This line only changed the number

PR #674, opened 2019-02-01
Searchable backoffice services view

### LINKED ISSUE #632: As a service portfolio manager i want to be able to find easily service i need to update
Possible solution for that can be adding filters to the mp backoffice. If this functionality is too expensive for 0.1 version, we need an equivalent to let SPM easily find services they want to edit at given moment. Cause with a long list of services (at the beginning only 2 people will handle the updates) it will be unbearable to manage. other possibility is search

Acceptance critaria 
I believe those can be agreed after we decide which mechanism we want to use?

mkasztelnik:
This is an implementation of following user story:

> As a service portfolio manager i want to be able to find easily service i need to update

To fullfil this requirement similar view as marketplase services index was created (with small modifications such as urls or service `status` badge). So user can use elasticsearch or dedicated filters. 

searchable-backoffice-services

Code was refactored in such a way that 95% of code for marketplace and backend is the same. That is why there is no new tests, since searching and filtering is already tested in the marketplace. Refactoring includes also paths improvements:

```
/categories/:category_id -> /services/c/:category_id
```

This change was needed because my plan was to reuse all the logic for maketplace and backoffice (take a look at `Service::Categorable` controller concern), but in the backoffice we have `/categories/:id` view already used to show and edit category tree. 

For the sake of compatibility `categories/:category_id` is redirecting service into `services/c/:category_id`.

## Further improvements

Despite the current status can be further improved. Current styling is not flexible enough to render correctly just a little page header:

backoffice-services-ui-problem

This is a place where @abacz and @jarekzet should enter the game. 

There is one last issue which should be solved in the following PRs: counter cache for categories. Currently, we are caching information about published services (drafts are not taken into account). This is wrong from the backoffice point of view because here we have 2 scenarios:
  * service portfolio manager is browsing **all** services so categories tree should show all services count 
  * service owner is browsing **only** owned services. In this case, the only count of owed services should be shown

Fixes #632


michal-szostak:
I'll need a little more time to read all the changes, but first thing - I see no changelog entry.

mkasztelnik:
@michal-szostak yeap. I was doing rebase to latest master and forget about it. Fixed and changelog is in place.

Issue #654, opened 2019-01-24
As SP manager I should be able to fill fields releted to URL in back office without problems related to validation


agpul:
Acceptance criteria (minimum):
- in fields where URL is required I should see a tip about URL format

Nice to have:
- maybe https should be complited automatically or we should change URL validation

PR #678, opened 2019-02-07
#654 backoofice url validation

kmarszalek:
#654 
it was decided that for now i will add hint under the url input and remove leading and trailing spaces before validation.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -94,6 +94,7 @@ class Service < ApplicationRecord
   validates :status, presence: true
 
   after_save :set_first_category_as_main!, if: :main_category_missing?
+  before_validation :strip_whitespace
 
   def main_category
     @main_category ||= categories.joins(:service_categories).
@@ -121,5 +122,16 @@ class Service < ApplicationRecord
 
     def main_category_missing?
       service_categories.where(main: true).count.zero?
+    end
+
+    def strip_whitespace
+      self.terms_of_use_url&.strip!
+      self.access_policies_url&.strip!
+      self.corporate_sla_url&.strip!
+      self.webpage_url&.strip!
+      self.manual_url&.strip!
+      self.helpdesk_url&.strip!
+      self.tutorial_url&.strip!
+      self.connected_url&.strip!
     end
 end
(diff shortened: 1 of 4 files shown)


app/models/service.rb
```diff
@@ -122,4 +123,15 @@ def aod?
     def main_category_missing?
       service_categories.where(main: true).count.zero?
     end
+
+    def strip_whitespace
+      self.terms_of_use_url = self.terms_of_use_url.strip unless self.terms_of_use_url.nil?
```
michal-szostak: you can substitute
`self.terms_of_use_url = self.terms_of_use_url.strip unless self.terms_of_use_url.nil?`
to
`self.terms_of_use_url&.strip!`

(same goes for other `strip`s here

app/models/service.rb
```diff
@@ -94,7 +94,8 @@ class Service < ApplicationRecord
   validates :status, presence: true
 
   after_save :set_first_category_as_main!, if: :main_category_missing?
-
+  before_validation :strip_whitespace, only: [:terms_of_use_url, :access_policies_url, :corporate_sla_url,
+                                                 :webpage_url, :manual_url, :helpdesk_url, :tutorial_url, :connected_url]
```
michal-szostak: I don't think that `only` is a viable argument for `before_validation`, It shouldn't be needed here anyway

config/locales/simple_form.en.yml
```diff
@@ -29,9 +29,24 @@ en:
...
+        manual_url: |
+          Url should start with http or https [e.g. http://webpage.org]
+        webpage_url: |
+          Url should start with http or https [e.g. http://webpage.org]
+        corporate_sla_url: |
+          Url should start with http or https [e.g. http://webpage.org]
+        access_policies_url: |
+          Url should start with http or https [e.g. http://webpage.org]
+        terms_of_use_url: |
+          Url should start with http or https [e.g. http://webpage.org]
```
michal-szostak: I'm not entirely sure, but can't we define single message for URL validation error?

I'm not 100% it can be done, maybe someone more versed in simple_form (@mkasztelnik ?) might be able to answer this question.

But this can be easily fixed later, so you should not spend too much time on this one.

michal-szostak:
Also - there are no tests

Issue #680, opened 2019-02-08
Update email releted to Voucher functionality


agpul:
We need to put voucherID to the email content even if user marks "I need a voucher" in Marketplace.

To reproduce:
1. User marks " I need a voucher"
2. JIRA ticket has filed "Voucher ID" which is empty
3. Operator enters voucherID to the JIRA manually
4. Operator puts the ticket in the state "Approved"
5. User gets e-mail, where: https://portal.exoscale.com/register?coupon=XXXX , XXXX should be repleaced by voucher ID.

Dear User,
This email is to inform you that your request to access the HNSciCloud pilot services via the EC3 portal [] has been approved.
To redeem an Exoscale voucher, please follow these steps:
Open the provided voucher link within a web browser.
A typical link looks like: https://portal.exoscale.com/register?coupon=XXXX
Enter the email address and password you wish to use. Accept the terms and sign up.
A validation email has been sent. Check out your mailbox and click on the verification link.
Choose “for team projects” and fill your details. Choose your Exoscale organisation name and submit.
To configure the tenant settings:
Access the Exoscale dashboard at: https://portal.exoscale.com/ and check the tenant settings (click in the User profile on the left)
Remember the API key and Secret key settings for the EC3 portal.
How to monitor your voucher credit:
Access the Exoscale dashboard at: https://portal.exoscale.com/ (billing section)
== IMPORTANT NOTICE ==
The Applicant is tasked to acknowledge EGI in the scientific publications/presentations benefiting from the Service. The following acknowledgement text can be used for this purpose: "This work used the EGI Applications on Demand service, which is co-funded by the EOSC-hub project (grant number 777536). The HNSciCloud project (grant number 687614) is also sponsoring the service, allowing users to access the HNSciCloud services pilot for limited scale usage using the voucher schemes provided by Exoscale".
== HELPDESK ==
If you need of any assistance, please describe your problems writing at applications-platform-info@mailman.egi.eu. Your request will be processed by the EGI User Support operator of the Service.
Should you have any questions, please contact us at [2].

Your Sincerely,
EGI User Community Support Team
[1] The Elastic Cloud Computing Cluster (EC3) - https://servproject.i3m.upv.es/ec3-ltos/index.php
[2] support@egi.eu

agpul:
now we have JIRA connected and it turned out that this functionality works bad, voucherID still isn't set

kmarszalek:
[...]

PR #692, opened 2019-02-11
#680 vouchers id in mail

kmarszalek:


agpul:
I think that I am not able to check this PR because there is no JIRA connected.

app/mailers/project_item_mailer.rb
```diff
@@ -36,6 +36,7 @@ def rate_service(project_item)
 
   def aod_voucher_accepted(project_item)
     @user = project_item.user
+    @voucher_id = project_item.voucher_id.blank? ? "" : project_item.voucher_id
```
michal-szostak: `voucher_id.blank?` means it is equal to `""` so it can be shortened to  
`@voucher_id = project_item.voucher_id` (`voucher_id` in `schema.rb` is set as `not  null`)

PR #697, opened 2019-02-15
Offer type with service type taken as default

### LINKED ISSUE #687: Ordering feature associated to a service offer
Many services (all EUDAT services for example) foresee a service option that give an immediate access to a service and others that require ordering. Thats the reason why we need to change ordering features.

mkasztelnik:
Starting from now offer can define customized ordering type (one of `normal`, `open_access`, `catalog`). If a type is `nil` in the offer than default is taken from the `Service::service_type`. On the first order wizard view (offer selection) only `normal` and `open_access` offers are presented - `catalog`ues are filtered out.

After this was added it was possible to add dedicated select offer buttons directly on `service#show` view:

screenshot from 2019-02-15 14-38-38

For now, the first ordering step is not removed (so we have responsibility duplication!). I did not remove this view because it is required for "main" `Order` button to work. This is task for @abacz, @roksanaer and @toszep to agree if this should stay of some redesign of this button will be performed.

Fixes #687

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Please view this file on the master branch, on stable branches it's out of date.
 - voucher_id in aod_voucher_accepted email (@kmarszalek)
 - voucher_id updates from jira (@michal-szostak)
 - `elasticsearch` version added to `.tool-versions` file (@mkasztelnik)
+- Offer type (`normal`, `open_access`, `catalog`) with service type taken as
+  default (@mkasztelnik)
 
 ### Changed
 - Change service phase values into keys (@mkasztelnik)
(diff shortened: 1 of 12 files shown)


agpul:
I have some additional comments:

- when we choose one offer by clicking order and then we want to change our choice, we go back to the step number one (Offer selection) and here we have "Select" buttons on offers. In my opinion we should change names of this buttons to Order/Get access/Goto the service etc.
- what with services which have only one offer, for me it is inconsitent that they have order button only next to the service, but  I might be wrong

mkasztelnik:
@agpul I agree with the second point. For the first point, I don't think this is a good change since we have the "Next" button above. A better option IMHO would be to have some visual representation of the box (I'm looking at you @abacz) which shows distinguish between different offer types. 

I believe that this PR is the first step and a lot of other elements will be redefined to fulfill the correct ordering model flow. For now, I was trying to add the required functionality without disturbing the existing one.

Issue #653, opened 2019-01-24
SP Manager should be able to add attributes to offer from backoffice


agpul:
Acceptance criteria:

- I can load or paste json file with attribute to each offer

PR #698, opened 2019-02-15
[#653] Offer attribute in backoffice

martaswiatkowska:
SPM can add offer parameters in backoffice. 
Exemplary json to tests:
{
	"id":"id1",
	"type":"select", 
	"label":"Number of CPU Cores",
	"config":{"mode":"buttons", "values":[1, 2, 4, 8]},	
	"value_type":"integer",
	"description":"Select number of cores you want"
}

Fix: #653

### DIFF
--- a/app/models/attribute.rb
+++ b/app/models/attribute.rb
@@ -101,6 +101,7 @@ class Attribute
 
   def self.from_json(json)
     JSON::Validator.validate!(ATTRIBUTE_SCHEMA, json)
+
     case json["type"]
     when "input"
       attr = Attribute::Input.new
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -14,12 +14,14 @@ class Offer < ApplicationRecord
 
   enum status: STATUSES
 
+  before_save :convert_parameters_to_json, on: [:create, :update]
+
   belongs_to :service
+
+  counter_culture :service, column_name: proc { |model| model.published? ? "offers_count" : nil }
 
   has_many :project_items,
            dependent: :restrict_with_error
-
-  counter_culture :service, column_name: proc { |model| model.published? ? "offers_count" : nil }
 
   validate :set_iid, on: :create
   validates :name, presence: true
@@ -27,6 +29,8 @@ class Offer < ApplicationRecord
   validates :service, presence: true
   validates :iid, presence: true, numericality: true
   validates :status, presence: true
+  validate :parameters_are_valid_attributes, on: [:create, :update]
+  validates :parameters_as_string, attribute_id_unique: true
 
   def to_param
     iid.to_s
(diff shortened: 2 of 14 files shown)


kmarszalek:
changelog :)

app/models/offer.rb
```diff
@@ -12,6 +12,8 @@ class Offer < ApplicationRecord
   validates :description, presence: true
   validates :service, presence: true
   validates :iid, presence: true, numericality: true
+  validates :parameters, json: true, if: :parameters?
+  validate :create_atributes, on:  [:create, :update], if: :parameters?
```
bwilk: Marta I think you should stay away from validators side-effects. The purpose of validator should be to validate, not to convert value types. See create_attributes method. The side effect hidden there is rewriting parameters field with type convertion.

app/models/offer.rb
```diff
@@ -22,6 +24,18 @@ def attributes
   end
 
   private
+    def create_atributes
+      return unless errors.blank?
+
+      self.parameters = (parameters || []).each_with_index.map do |param, index|
```
bwilk: I think you should either - create different public method for setting parameters given as strings OR use different field to store that information (e.g. let's call it `parameters_as_strings`) - this can be valuable because your motivation was probably using a model field to have easy access from the perspective of backoffice form. In the 2. this will mean you could override `parameters_as_strings=` and in the body of this method you set `parameters` field value (as a side effect of setter - think about it - is it good or bad?). Alternative approach to creating `parameters` field value as a side-effect of `parameters_as_strings=` can be using `before_save` handler.

app/models/offer.rb
martaswiatkowska: I agree with that. I should change this function in validation and after that make conversion.

bwilk:
By the way - @mkasztelnik. We found two similar uses of validator - one in the new code, one in the older one. Do you think we should change this as well to abstain from using validator for side-effects?
https://github.com/cyfronet-fid/marketplace/blob/7ef131a2363cae719b2661224bc4958e81623e36/app/models/offer.rb#L10

goreck888:
From the functional side of this feature everything is working correctly, instead of case when you make a mistake by writing attribute json. Then there show additional backslashes by " sign, which causes every backslash needs to delete it manually, because try to save it without these deletions generates the next error and more backslashes.

bwilk:
Once you add 2 attributes to an offer, save it, go back to the edit mode and then remove both - after saving the offer, attributes stay untact.

bwilk:
Known issue and the question - can we live with it? I personally think we cannot. When adding new parameters to an offer, the users needs to use different ids. Otherwise many strange things happen. Id's stored in the database are used to create forms, so user interface relies on their uniqueness in scope of the service offer. Either we need to take the control over ids from the user or make sure that the user provide unique values.

app/models/attribute.rb
```diff
@@ -36,14 +36,20 @@ def value_valid?
 
   def validate_config!
     JSON::Validator.validate!(config_schema, config)
+    rescue => e
+      errors.add(:config, e.message)
   end
 
   def validate_value!
     JSON::Validator.validate!(value_schema, value)
+    rescue => e
```
bwilk: We need to rethink the method of using attribute validation. I can propose some solutions - let's stay in touch. This method was intended to throw exceptions as it has an exclamation mark at the end of the name. Contrary there is another method `value_valid?` that can be used to validate without throwing exceptions. There is  also active record model method `valid?` that will use `value_validity` method and add appropriate errors to the model object. This is underdeveloped in general in `Attribute` class so I think we need to make sure we fix this.

app/models/offer.rb
```diff
@@ -43,7 +46,45 @@ def catalog?
...
+  def parameters_as_string
+    if !@parameters_as_string && !parameters.nil?
+      @parameters_as_string = parameters.map(&:to_json)
+    end
+    @parameters_as_string
+  end
+
   private
+    def parameters_are_valid_attributes
+      return unless errors.blank?
```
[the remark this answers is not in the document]
martaswiatkowska: This line is unnecessary. It's the rest of previous implementation.

PR #701, opened 2019-02-19
[#680] Update Voucher ID from JIRA

michal-szostak:
# What is in this PR?

* Update voucher id after change occurs in JIRA
* Fix checker to critically fail after being unable to create issue

# How does it look like?

image

Fixes: #680


bwilk:
Looks neat. Maybe if I can add something, we could make Voucher ID even more visible by placing the current ID in a dedicated section in the right panel. @abacz please evaluate this possibility.

PR #704, opened 2019-02-21
Date attribute min and max validation

### LINKED ISSUE #501: start date in service configuration shouldn't be older than current day
on the second step of order, some services have Start date attribute

            - id: "id5"
             label: "Start of service"
              description:  "Choose service start date"
             type:  "date"
             value_type:  "string"

It would be best if a user could not choose a date older than current day

mkasztelnik:
Fixes #501

### DIFF
--- a/app/models/attribute/date.rb
+++ b/app/models/attribute/date.rb
@@ -22,7 +22,50 @@ class Attribute::Date < Attribute
     @value = param
   end
 
+  def value_validity
+    super
+
+    errors.add(:id, "Please select date after #{I18n.l(min_date - 1.day)}") if before_min?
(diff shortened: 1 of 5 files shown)


app/models/attribute/date.rb
```diff
@@ -22,7 +22,50 @@ def value_from_param(param)
     @value = param
   end
 
+  def value_validity
+    super
+
+    errors.add(:id, "Please select date after #{I18n.l(min_date - 1.day)}") if before_min?
```
bwilk: you can use :min and :max for distinguishing between error causes

mkasztelnik: Cannot do it right now without heavy attributes refactoring since the error is shown only when the error with `id` identifier is selected.

Issue #686, opened 2019-02-11
Change of names of buttons and headers


agpul:
We need to change names of some buttons and headers:

- [x] My services (https://marketplace.docker-fid.grid.cyf-kr.edu.pl/projects) to My projects and orders
- [x] Go to the service (https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services/amber) to Access the service
- [x] Change the title of ‘Add to my services’ (https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services/amber)to ‘Add to a project’ button
- [x] Change the title of ‘Order’ (https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services/amber)to ‘Add to a project’ button
- [ ] Add more descriptive text on the project management page, especially highlight that ‘by adding a service into a project a respective order request is automatically submitted’
- [x] “My services” in “My projects” on all the pages where this label is used

mkasztelnik:
If value is repeated multiple times please use I18n to store translation (see e.g. `config/locales/en.yml`).

PR #705, opened 2019-02-21
[#686] Change names of buttons and headers

goreck888:


db/schema.rb
```diff
@@ -96,6 +96,29 @@
     t.index ["service_id"], name: "index_offers_on_service_id"
   end
 
+  create_table "order_changes", force: :cascade do |t|
```
mkasztelnik: We don't have order_changes and orders models. Why this was added to this pr?

kmarszalek: what are these changes

mkasztelnik:
Changelog entry is missing.

roksanaer:
what about "Services" label under "My projects and orders". shouldn't it claim "orders" or "projects" to be consistent? @agpul did u check this?

goreck888:
> Changelog entry is missing.

Now it's added

goreck888:
> what about "Services" label under "My projects and orders". shouldn't it claim "orders" or "projects" to be consistent? @agpul did u check this?
> 
> edit: ok, i get it. this is a project name, named Services ;) this one will be removed probably in the foreseeable future. Are we ready to merge this?
> @goreck888 ? why don't u look after ur pr? :P ;)

I edited it exactly as it is in the ticket describes whole changes. The one thing I didn't make is one of the points because I didn't find any similar text like it is in the description.
Why do you think I don't look after, my branch is rebased, up to date and ready to merge :)

PR #706, opened 2019-02-22
Back office research areas list

jarekzet:
One thing - on the list there ae buttons "edit" and "delete" icon. They need to be implemented - now they are static. 

Can anybody add those buttons?
# @mkasztelnik @agpul @martaswiatkowska


jarekzet:
Added button styles modifications:
* font-weight for links and button to 300.
* trash icon to fas https://fontawesome.com/icons/trash-alt?style=regular

PR #723, opened 2019-03-01
Add markdown rendering in the offer's description

goreck888:


goreck888:
> image
> Look at summary. I still see here bad rendering.

It's working now after the new commit.

Issue #584, opened 2018-12-11
Add service offer status in the MP back office


roksanaer:
Service owner is allowed to decide wether his service offer is visible or accessible in the MP for the users.
Each service offer has a status. For now the status can be:
draft
published
In the service offer definition, service owner can decide if the service should be visible to users.
Enum set to published means that the service offer is visible in the service entries and available for ordering process. Enum set to draft: service offer in not visible or findable in the MP GUI

after this feature is added all service offers should be set to published

Logic regarding service offer editing is the same as in services (SP manager can edit service offer regardless of its status - draft or published)
service owner connected to the service should have permission to view all service offers underpinned to a service (regardless of their status) 

Regarding service offer connection with a service:

possibility to edit a service offer is available for statuses draft and published of a service 
service transition publish -> draft makes all underpinned service offers a draft
draft -> publish service transition does not publish service offers. it would be useful to add adequate success info box:
Your service has been published! All offers connected to the service did not change their status. If you wish to publish your service offers, please go to the service offer panel and activate chosen offers.

roksanaer:
this feature should be implemented basing on service status analogy #560

roksanaer:
i've updated the description

Issue #699, opened 2019-02-19
Fixing offer management in Marketplace backoffice


agpul:
**Describe the bug**
Now, when we want to edit the offer we have to click on service "Stop showing in MP". This situation is unintuituve.

**Expected behavior**
We should make possible to edit/delete (mange) offer without such connecion with the service. 
The only connection should be that when we click "Stop showing in MP" on service then the offer also is hidden.

PR #724, opened 2019-03-01
[#584] Add status to service

martaswiatkowska:
Fix: #584 #699

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ## [Unreleased]
 
 ### Added
+- Offer status (@martaswiatkowska)
 
 ### Changed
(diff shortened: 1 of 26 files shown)


agpul:
> The second thing is that when I add the new service (with no offers) I don't have Order button and it is OK, but when I add the first offer which is a draft I can see the Order button and this behavior is wrong because my offer isn't still publish.
> This behavior should be change. Order button should appear after first offer is Published not Draft.

Now this case works OK, but I have antoher problem. I add offer1 and publish it (I have order button) ->it's OK.  I add Offer2 but I don't publish it and I can see offer1 in the front although it should appear only when offer2 is published. Look here:
Backoffice:
image
Service:
image

Moreover I would like to see status (drat, publish) per offer somewhere in the backoffice

Issue #735, opened 2019-03-07
Support for field in JIRA releted to offer status


agpul:
We need to suupport new field in JIRA:
SO-OfferType (select, optional)
FieldId=10906
Options:
normal: selectedValue=10906
open_access:selectedValue=10907
catalog: selectedValue=10908

Expected behaviour:
During service ordering we should send to JIRA the offer type (**not service type**) which was specified during offer specification.

PR #739, opened 2019-03-07
[#735] New jira fields connected to type of offert

martaswiatkowska:
Fix: #735


agpul:
As I commented last time, I ordered the service but the field is empty in JIRA.

PR #740, opened 2019-03-08
Fixed parameters from yaml

martaswiatkowska:
Fixed services creation from yaml

### DIFF
--- a/spec/models/attribute_spec.rb
+++ b/spec/models/attribute_spec.rb
@@ -150,17 +150,18 @@ RSpec.describe Attribute do
   end
 
   it "creates correct number range with invalid value from json" do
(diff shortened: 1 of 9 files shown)


spec/models/attribute_spec.rb
```diff
@@ -150,17 +150,18 @@
   end
 
   it "creates correct number range with invalid value from json" do
```
bwilk: Would you create one more use case with a value that corresponds to the range conditions?

martaswiatkowska: That case already exist. Test start at 139 line

michal-szostak:
As we suspected it was a problem with difference in session storage backends. I changed test storage backend to be the same as production and development, I've also modified travis config to include redis.

PR #751, opened 2019-03-13
Add link to the tutorial of attributes in backoffice

goreck888:
Add a custom label with link to the tutorial

### DIFF
--- a/app/views/backoffice/services/offers/_form.html.haml
+++ b/app/views/backoffice/services/offers/_form.html.haml
@@ -4,8 +4,8 @@
   = f.input :description, input_html: { rows: 10 }
   = f.input :offer_type, collection: Offer.offer_types.keys.map(&:to_sym)
   = f.input :parameters_as_string,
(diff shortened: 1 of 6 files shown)


app/views/backoffice/services/offers/_form.html.haml
```diff
@@ -4,6 +4,8 @@
   = f.input :description, input_html: { rows: 10 }
   = f.input :offer_type, collection: Offer.offer_types.keys.map(&:to_sym)
   = f.input :parameters_as_string,
+      label: "Parameters as string (#{link_to 'See the tutorial', 'https://confluence.egi.eu/x/wYI7Ag',
```
kmarszalek: make this url configurable.

mkasztelnik: I agree with this statement (and I will externalize this link), but I don't like the idea that marketplace documentation lived outside the merketplace. IMHO it should be inside the same repository and should be in sync with current marketplace version.

Issue #736, opened 2019-03-07
The list of small improvements in Backoffice 


agpul:
As a result of this issue I expect to have fixed the list of small changes in Backoffice:

- [ ] Service website filed in Backoffice should appear only when we choose open_access in Service type field
- [ ] Service website field in Backoffice should appear directly under Service type field
- [ ] Add hint to Service webiste field: This url will be available under "Access the service" button in the service view
- [ ] Add hint to Webpage url field: This url will be available under "Webpage" in Support section in the service view

PR #769, opened 2019-03-18
[#736] Backoffice service form updated

martaswiatkowska:
Fix: #736

### DIFF
--- a/app/javascript/controllers/service_controller.js
+++ b/app/javascript/controllers/service_controller.js
@@ -1,0 +1,20 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = ["connectedUrl", "serviceType"]
+
+  connect(){}
+
+  initialize(){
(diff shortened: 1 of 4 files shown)


app/javascript/controllers/service_controller.js
```diff
@@ -0,0 +1,21 @@
...
+
+  connect(){}
+
+  initialize(){
+    this.showConnectedUrl()
+  }
+
+  showConnectedUrl(event){
+    const serviceType = this.serviceTypeTarget.value
+    if(!this.connectedUrlTarget.classList.contains("hidden-fields")){
```
michal-szostak: This condition is not required. `classList` is an object, it won't add multiple classes having the same name.

So this condition can be reversed:

```
 if (serviceType == "open_access"){
     this.connectedUrlTarget.classList.remove("hidden-fields");
} else {
     this.connectedUrlTarget.classList.add("hidden-fields");
}
```

PR #772, opened 2019-03-19
Show accept terms and conditions only when url is defined

### LINKED ISSUE #737: Accept terms and conditions checkbox from the summary step
On the summary step, we have checkbox Accept terms and conditions. Fix this field

mkasztelnik:
Related #737

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


roksanaer:
I'm not sure what to do with this PR as it was decided that we should remove terms and conditions completely (decision takin in Munich).
But as this feature work and has approval from Michał and Krzyś already, maybe we should merge it and then wait till we move to removing the terms and conditions? or can we fix this as a part of this PR?

PR #775, opened 2019-03-25
764 apply ui/ux changes

jarekzet:
TODOs:
1. need to add active filters under category name.
2. Number of services on list was added statically, need to add dynamic information.


abacz:
Dear @agpul and developers,
we need your help with some stuff in this PR:
- filter tags under the header , now they are static
- header "looking for: [keywords]" should be visible only for search result
- improve filters by adding search input for providers and research areas. Choice.js(?)
Thx <3

app/views/services/_categorization.html.haml
```diff
@@ -1,15 +1,21 @@
...
+  %dt.x-small
     %span
       Research area:
+  %dd.x-small
     %strong.mr-4= safe_join(area_links, ", ")
+%dl.mb-0
+  %dt.x-small
     %span
-      For:
+      Dedicatedor:
```
kmarszalek: typo

Issue #756, opened 2019-03-18
Improvement of the order displaying in the search tool


agpul:
As a result of this ticket, I expect to have improved the order displaying as follows (minimum path (from highest to lowest weight)):
1. the highest weight should have a service title
2. next should be service description 

If it will be rather easy and rather fast we can take into consideration a more complicated path, where the order is as follows (from highest to lowest weight): 
1. service title
2. offer title
3. service description
4. offer description
5. tags

Additional input:
1. If we want to do this issue we need to reverse the order of performing activities. Now we have elastic search, filters. We should change this order as follows: narrowing the scope by filters, the rest should go to elastic search. 
2. Idea: Pagination should be pinned to elastic search.

Issue #758, opened 2019-03-18
Support misspelling in the search tool 


agpul:
As a result of this ticket, we would like to handle misspells on the search bar level. As we discussed we should take actions as follows: 
1. make research in the elastic search how can we support that issue
2.  list possible options in this ticket (@martaswiatkowska )
3. apply option which covers this requirement and can be done rather fast

Issue #757, opened 2019-03-18
Improvement of substring displaying 


agpul:
As a result of this ticket, we would like to have functionality which will support substring in the search. We discussed a couple of options and one of this option should be chosen (taking into consideration sprint time):
1. add a tip for the user (something like "You should add * at the end of the search if you want to find substrings" or maybe something better:))
2. add * at the end of each query (it will be invisible for the user)
3. add * at the begging of each query (it will be invisible for the user)
4. @martaswiatkowska can propose other option supported by elastic search

Additional input:
- please check if the elastic search can use more than one tokenizer

PR #778, opened 2019-03-26
Initial usage of searchkick

martaswiatkowska:
Added searchkick as gem to elasticsearch. 
In elastic search are stored title, descriprion, status and rating. 
Services are searched based on title and description. Title is boosted.
Misspelling is set by default. Also added search by substring. 
Changed order of filtering and searching. Now services are firstly filtered.  

Fixed: #756, #757, #758

### DIFF
--- a/app/views/services/_order.html.haml
+++ b/app/views/services/_order.html.haml
@@ -3,4 +3,5 @@
     = options_for_select([["by name A-Z", "title"],
                           ["by name Z-A", "-title"],
                           ["by rate 1-5", "rating"],
-                          ["by rate 5-1", "-rating"]], params[:sort])
+                          ["by rate 5-1", "-rating"],
+                          [" ", nil]], params[:sort])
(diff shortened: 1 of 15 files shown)


martaswiatkowska:
Searchkick automatically uses a different index name in each environment. 
Yes, reindex are required.

agpul:
Problem with Sentinel Hub search disappeared.

app/views/services/_order.html.haml
```diff
@@ -1,7 +1,8 @@
...
-  %select#sort.query-sort.form-control.form-control-sm{ "data-submit-on-change": "data-submit-on-change" }
+= form_tag url_for, method: :get, role: "search", "data-controller": "order" do
+  %select#sort.query-sort.form-control.form-control-sm{ "data-target": "order.select",
+                                                        "data-submit-on-change": "data-submit-on-change" }
     = options_for_select([["by name A-Z", "title"],
                           ["by name Z-A", "-title"],
                           ["by rate 1-5", "rating"],
                           ["by rate 5-1", "-rating"],
-                          [" ", nil]], params[:sort])
+                          ["Most popular", "_score"]], params[:sort])
```
michal-szostak: I think that _Best Match_ (same as e.g. EBAY) would be a better wording.

Issue #765, opened 2019-03-18
Apply UX/UX changes in Item detail view


agpul:
As a result of this ticket, we would like to have applied changes in Item detail view according to:
https://projects.invisionapp.com/share/DSP0QZN5ZX3#/screens/330285958_Item-Detail

Zeplin: zpl.io/bAZL146

PR #782, opened 2019-03-28
#765 Added detail view mods

jarekzet:
also container max-width increased to 1180px


agpul:
@jarekzet can you take care?

jarekzet:
Shadow issue is fixed in latest commit

db/schema.rb
```diff
@@ -81,6 +81,16 @@
     t.index ["sluggable_type"], name: "index_friendly_id_slugs_on_sluggable_type"
   end
 
+  create_table "offer_links", force: :cascade do |t|
```
kmarszalek: where is migration for this change in db schema?

PR #789, opened 2019-04-01
Change elasticsearch import to searchkick reindex in setup

martaswiatkowska:


mkasztelnik:
You should add information about this change to changelog

PR #800, opened 2019-04-04
SO-1 jira field as json

### LINKED ISSUE #743: Transfer information about order and attributes to JSON
As a result of this ticket we would like to:
- have SO-1 field which now we map to JIRA in JSON file (only this one field should be in JSON)
- @michal-szostak will analyze if we can inject this json to SO-1 field in JIRA, if it is not possible, please inject json file to the description

mkasztelnik:
Fixes #743


agpul:
@wziajka can you check connection with EOSCSOPR Jira project, I ordered service but I can't see it in JIRA (last services in JIRA are from 18.03)

PR #802, opened 2019-04-05
Elasticsearch autocomplete

martaswiatkowska:
Added elasticsearch autocomplete. 
I use https://github.com/afcapel/stimulus-autocomplete but copied to our repository.

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -5,6 +5,7 @@ class ServicesController < ApplicationController
   include Service::Searchable
   include Service::Categorable
   include Service::Sortable
+  include Service::Autocomplete
 
   def index
     filtered = filter(scope)
--- a/app/javascript/controllers/autocomplete_controller.js
+++ b/app/javascript/controllers/autocomplete_controller.js
@@ -1,0 +1,236 @@
+// Copy of repository https://github.com/afcapel/stimulus-autocomplete
+
+import { Controller } from 'stimulus'
+import debounce from 'lodash.debounce'
+
+export default class extends Controller {
+
+  static targets = [ 'form', 'input', 'hidden', 'results' ]
+
+  connect() {
+    this.resultsTarget.hidden = true
+    this.inputTarget.setAttribute('autocomplete', 'off')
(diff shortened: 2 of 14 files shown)


app/javascript/controllers/autocomplete_controller.js
```diff
@@ -0,0 +1,227 @@
...
+    const params = new URLSearchParams(url.search.slice(1))
+    params.append('q', query)
+    url.search = params.toString()
+
+    this.element.dispatchEvent(new CustomEvent('loadstart'))
+
+    fetch(url.toString())
+      .then(response => response.text())
+      .then(html => {
+        console.log(this.data)
```
bwilk: remove logs

app/controllers/services_controller.rb
```diff
@@ -25,6 +25,26 @@ def show
...
+      fields: ["title"],
+      operator: "or",
+      match: :word_middle,
+      limit: 5,
+      load: false,
+      highlight: { tag: "<b>" }
+    )
+    html_results = ""
+    query.highlights.each_with_index do |s, i|
+      html_results += "<li class=\"dropdown-item\" role=\"option\" data-autocomplete-value=\"#{i}\">#{s[:title]}</li>"
```
bwilk: This should be changed to 

~~~~ 
query.with_highlights.each do |s, h|
  html_results += "<li class=\"dropdown-item\" role=\"option\" data-autocomplete-value=\"#{s.title}\">#{h[:title]}</li>"
end
~~~~

agpul:
Now it works correctly but I have additional comment. The mouse cursor is changing to something strange when I want to choose the service. It makes me feel that I can't click. Please have a look.
kursor

martaswiatkowska:
@agpul  please check it again in other browser. I don't know what's happened because  I can't reproduce the error.

app/controllers/services_controller.rb
mkasztelnik: More compact version:

```ruby
query.highlights.map.with_index { |s, i| ... }.join
# of if you don't need index
query.highlights.map { |s| ... }.join
```

mkasztelnik:
Also please add changelog entry

PR #811, opened 2019-04-09
Added highlights to main search

martaswiatkowska:


app/controllers/backoffice/services_controller.rb
```diff
@@ -14,7 +14,10 @@ def index
     filtered = filter(scope)
     from_category = category_records(filtered)
 
-    @services = search(from_category.order(ordering))
+    from_search = search(order(from_category))
+    @services = from_search
+    @highlights = {}
+    from_search.is_a?(Searchkick::Results) ? from_search&.with_highlights.each { |s, h| @highlights[s.id] = h  } : from_search
```
mkasztelnik: A little bit simpler version:

```
@highlights = from_search&.with_highlights&.map { |s, h| [s.id, h] }&.to_h
```

What is more this should be moved to concern and reused in front and backoffice

app/helpers/service_helper.rb
```diff
@@ -46,4 +46,8 @@ def providers(service)
   def providers_text(service)
     service.providers.map { |target| target.name }
   end
+
+  def service_title(service, highlights)
+    (highlights.present? && highlights[service.id].present?) ? highlights[service.id][:title].html_safe : service.title
```
mkasztelnik: This can be simplified a bit:
```ruby
highlights&.dig(service.id, :title) || service.title
```

app/models/service.rb
```diff
@@ -5,8 +5,9 @@ class Service < ApplicationRecord
   # ELASTICSEARCH
   # scope :search_import working with should_indexe?
   # and define which services are indexed in elasticsearch
-  searchkick text_middle: [:title, :description]
+  searchkick word_middle: [:title, :description]
```
mkasztelnik: How this is connected with this PR?

app/views/services/_paginate_simple.haml
```diff
@@ -1,12 +1,12 @@
-- if services.total_entries > 10
+- if services.length || services.total_entries > 10
```
mkasztelnik: How this is connected with this PR? Maybe it should be extracted to separate PR?

Issue #803, opened 2019-04-08
Check and change affiliation behaviour 


agpul:
**Describe the bug**
In the profile tab in Marketplace: https://marketplace.eosc-portal.eu/profile I have a list of created and active affiliations. When affiliation has status: active there is no action possible to take and it is OK. I have four affiliations in CREATED status and they have active buttons EDIT/DELETE but two of them behave as ACTIVE ones (I can't edit/delete them although I have Edit/Delete buttons)

**To Reproduce**
1. Go to https://marketplace.eosc-portal.eu/profile
2. Click on EDIT or DELETE button correlated with CREATED affiliation
![image](https://user-images.githubusercontent.com/43000481/55715182-7ddc6980-59f4-11e9-821c-2d4fb0c52ef8.png)
3. Get the message: You cannot edit an affiliation which has been activated or You cannot remove an affiliation which has a project_item

**Expected behavior**
It should be the possibility to edit/delete affiliation which has status: CREATED. For now, Active affiliation stays as it is now.

**Additional context**
It looks like I have an affiliation which has status CREATED but in the database (?) or somewhere it is active and I have services ordered using this affiliation.
Moreover, I think that this problem may touch only old affiliations because the newest affiliations I have don't have such problems. 

Please check and fix it (even if it is a problem of old affiliations only).

PR #821, opened 2019-04-11
[#803] Fix edit and delete affiliation

goreck888:
From now query for user affiliations includes user id.

Before this fix query for edit and delete affiliation included just iid which is counted from 1 for every user, so query has returned all affiliations with current iid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix bug with JIRA issue creation (attribute mapping to SO-1) (@michal-szostak)
 - Remove std out/err logs from test suite output (@mkasztelnik)
 - Fix show opinions for open access services (@mkasztelnik)
+- Fix authorising affiliation from user affiliations, not affiliations at all (@goreck888)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


app/controllers/profiles/affiliations_controller.rb
```diff
@@ -52,7 +52,7 @@ def affiliation_template
     end
 
     def find_and_authorize
-      @affiliation = Affiliation.find_by!(iid: params[:id])
+      @affiliation = Affiliation.find_by!(iid: params[:id], user_id: current_user.id)
```
mkasztelnik: A little bit compact version:
```ruby
user.affiliations.find_by(iid: params[:id])
```

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix bug with JIRA issue creation (attribute mapping to SO-1) (@michal-szostak)
 - Remove std out/err logs from test suite output (@mkasztelnik)
 - Fix show opinions for open access services (@mkasztelnik)
+- Fix edit and delete created affiliation and active affiliation without project items(@goreck888)
```
mkasztelnik: This changelog entry is a bit misleading, IMHO this is not connected with project item at all but the bug is connected with the possibility to modify other user affiliation.

app/controllers/profiles/affiliations_controller.rb
goreck888: Sorry, a bit habit from mongodb :)

CHANGELOG.md
goreck888: I changed it.

PR #823, opened 2019-04-15
Autocomplete content are serve only from ajax request

martaswiatkowska:
It's minor enhancement connected to PR #802

### DIFF
--- a/app/controllers/concerns/service/autocomplete.rb
+++ b/app/controllers/concerns/service/autocomplete.rb
@@ -12,12 +12,21 @@ module Service::Autocomplete
       load: false,
       highlight: { tag: "<b>" }
     )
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/service/autocomplete.rb
```diff
@@ -13,10 +13,11 @@ def autocomplete
       highlight: { tag: "<b>" }
     )
     html_results = query.highlights.map.with_index { |s, i| "<li class=\"dropdown-item\" role=\"option\" data-autocomplete-value=\"#{i}\">#{s[:title]}</li>" }.join
-
-    respond_to do |format|
-      format.html do
-        render status: :ok, html: html_results.html_safe
+    if request.xhr?
```
mkasztelnik: This is minor but you can externalize html result creation into separate method and invoke it only when we are serving xrh html format. Now it is invoked always.

Issue #788, opened 2019-04-01
Improvements on the front 


agpul:
We need to make some improvements according to filters (remains from the old sprint):
* [ ] filter tags under the header, now they are static
* [x] arrow (chevron) in filter category should have two states (up/down) depending on filter state
* [x] "collapse all" in filters should collapse everything below
* [x] header "looking for: [keywords]" should be visible only for search result
* [ ] improve filters by adding search input for providers and research areas. Choice.js(?)  
   (prev: https://invis.io/V2QZE87N9S5#/351955222_Item_List_Filters_Search)

agpul:
Closed by #870

PR #824, opened 2019-04-16
[#788] Improvements in filters on the front

goreck888:
Add changing arrows directions in filters
Add "Collapse All" button
Add "Looking for" if query is not blank
"Collapse All" link styling
Add services from and to counter
Add styles to the new elements


goreck888:
Now it's necessary to add input fields for filer multiselectable values in filters and that's the one thing isn't included yet in this PR.
It's added now for check if it looks and works good.

mkasztelnik:
Why there are some many features packed in one PR, even inside one commit? It is hard to distinguish which change is connected with which feature.

mkasztelnik:
Would it be possible to split it into several PRs?

goreck888:
>     1. I type something in the search bar and I have it in Looking for and it is OK:
>        image
> 
>     2. Then I choose a filter and in Looking for I've got an additional sign (this time +)
>        image
> 
>     3. Then I choose an additional filter and in Looking for I have now two additional signs (%2):
>        image
> 
>     4. The more filters I add the more strange signs I get.
>        image

This bug is not related with my PR, this bug is connected with other features and is visible on master branch.

app/views/services/index.html.haml
```diff
@@ -41,8 +41,11 @@
 
   .col-md-12
     %hr.category-separator
-  .col-md-6
-    = render "services/paginate_simple", services: @services
-  .col-md-6.float-right
+  .col-md-3.pt-md-2
```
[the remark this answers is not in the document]
mkasztelnik: Also, remember that similar behavior needs to be present in the back office, thus you need to test it there as well.

app/views/services/_filters.html.haml
```diff
@@ -1,12 +1,26 @@
...
+  = form_tag "", method: :get, role: "search", class: "filters", "data-sync-query-form": "data-sync-query-form" do
+    - @filters.each do |filter|
+      = render "services/filters/#{filter.type}", filter: filter
 
+    .row
+      .col-md-12
+        %button#filter-submit.btn.btn-primary.mt-4.w-100.filter-submit{ type: "submit" }
+          Filter
+
+:javascript
```
mkasztelnik: Such js should not be added to the page inline. Instead you should use similar style like in https://github.com/cyfronet-fid/marketplace/blob/master/app/javascript/packs/flash.js, https://github.com/cyfronet-fid/marketplace/blob/master/app/javascript/packs/application.js#L28, https://github.com/cyfronet-fid/marketplace/blob/master/app/javascript/packs/application.js#L56

app/views/services/_filters.html.haml
```diff
@@ -1,12 +1,26 @@
...
+  $(document).ready(function(){
```
mkasztelnik: In the new JS please avoid using jquery - it will be removed from the project shortly. Instead, use pure JS.

app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -935,6 +935,18 @@ h5 {
     }
   }
 
+  .collapsed .collapse-icon .fa-chevron-up,
```
mkasztelnik: How this is connected with bootstrap customization? If this does not customize existing bootstrap classes than it should be moved to separate file.

app/views/services/_filters.html.haml
mkasztelnik: Another fundamental question is why we are not using the build-in collapsable feature from bootstrap https://getbootstrap.com/docs/4.3/components/collapse/ ? This should simplify a lot the code.

app/views/services/index.html.haml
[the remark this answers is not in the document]
goreck888: Ok, but it's not my code, I just add styles and changed order of items in row.

app/javascript/stylesheets/_bootstrap-customizations.scss
goreck888: It's moved now to base.scss

app/views/services/_filters.html.haml
goreck888: Deleted, I use bootstrap collapse now.

PR #831, opened 2019-04-25
Hierarchical multiselect Research Area filter

### LINKED ISSUE #766: Change research area field form single select to multi-select
As a result of this ticket, we would like to have changed research area filter. 

Now we can filter only by one research area (single select filed) but we would like to have this field as multi-select, as below:
https://projects.invisionapp.com/share/V2QZE87N9S5#/screens/351955222_Item_List_Filters_Search

mkasztelnik:
In the scope of this PR Research Area filter was changed from select to hierarchical multiselect. What is more, this is implemented in a generic way, so any other model with [`ancestry`](https://github.com/stefankroes/ancestry) included can be configured to be this type of filter (take a look at `app/models/filter/research_area.rb` how it can be done).

Suppose we have the following research areas structure

```
A
 |- B 
 |- C
    |- D
```

## Counting
The count is a unique sum of services for an element and all its children (direct and indirect). 

## Filtering
  * when only `A` is selected then we are searching for services belonging to `A` or `B` or `C` or `D`
  * when `A` and `B` is selected we are searching for services belonging to `A` or `B`
  * when `B` is selected we are searching for services belonging to `B`
  * when `C` is selected we are searching for services belonging to `C` or `D`

## TODOs
  * [x] hierarchical multi-select filter type
  * [x] research area as hierarchical multi-select filter
  * [x] unify `ul` list view for multiselect with `div` version (counter is in wrong place) - @abacz, @jarekzet   can you help me with this issue?
  * [x] reuse `_multiselect_level.html.haml` partial in `_multiselect.html.haml` partial instead of `multi_checkbox` helper method (after that `multi_checkbox` can be removed)

Fixes #766


mkasztelnik:
After UI improvements a lot of refactoring and unification this PR is ready to be reviewed and finally merged.

app/models/filter/ancestry_multiselect.rb
```diff
@@ -0,0 +1,69 @@
...
+      if ids.size.positive?
+        joining_table_name = @joining_model.table_name.to_sym
+        services.joins(joining_table_name).
+          where(joining_table_name => { relation_column_name.to_sym => ids })
+      else
+        services
+      end
+    end
+
+    def relation_column_name
```
bwilk: Wow. Smart 😄. Do we want to be that smart and assume model and relationship names are the same?

app/models/filter/ancestry_multiselect.rb
```diff
@@ -0,0 +1,69 @@
...
+    def children_services(arranged)
+      arranged.inject(Set.new) do |s, (record, children)|
+        s | grouped_services[record.id] | children_services(children)
+      end
+    end
+
+    def grouped_services
+      @grouped_services ||=
+        @joining_model.pluck(relation_column_name, :service_id).
+          inject(Hash.new { |h, k| h[k] = Set.new }) { |h, (k, v)| h[k] << v; h }
```
bwilk: This line is just gorgeous 🎪. I'd love to review this but sorry - you lost me here 😄 `|h, k| h[k] = Set.new `. Is that just an empty `Hash` with values that will be of class `Set` ??

app/models/filter/ancestry_multiselect.rb
```diff
@@ -0,0 +1,69 @@
...
+          id: record.id,
+          count: (grouped_services[record.id] | children_services(children)).size,
+          children: create_ancestry_tree(children)
+        }
+      end
+    end
+
+    def children_services(arranged)
+      arranged.inject(Set.new) do |s, (record, children)|
+        s | grouped_services[record.id] | children_services(children)
```
bwilk: I'm too short for this one as well. Is `| ` sign just an `or` operator in this case or some form of sum??

app/models/filter/ancestry_multiselect.rb
bwilk: Also I'm not sure about that but there is also this `field_name` how does it relate to this one?

app/models/filter/ancestry_multiselect.rb
```diff
@@ -0,0 +1,69 @@
...
+        services
+      end
+    end
+
+    def relation_column_name
+      "#{@model.table_name.singularize}_id"
+    end
+
+    def ids
+      @research_area_ids ||= begin
```
bwilk: This should be renamed as you are using it in a generic Filter class

app/models/filter/ancestry_multiselect.rb
bwilk: Ok. Found it. Ignore me 😄 https://ruby-doc.org/stdlib-2.6.3/libdoc/set/rdoc/Set.html#method-i-7C

app/models/filter/ancestry_multiselect.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def children_services(arranged)
+      arranged.inject(Set.new) do |s, (record, children)|
+        s | grouped_services[record.id] | children_services(children)
+      end
+    end
+
+    def grouped_services
+      @grouped_services ||=
+        @joining_model.pluck(relation_column_name, :service_id).
```
bwilk: I am biased here - will it be scalable in terms of performance? We are fetching at least as many records as number of services here. For now this number is only 150-ish, but I was trying to consider this factor in scope of filter counters. This implementation gives me some courage to do the same 😄

mkasztelnik: This is a consequence of choosing an elasticsearch query as the main filtering element. You cannot introduce counter cache, the only thing which can improve the situation is to use some kind of caching, but I was thinking about adding caching above filter implementation by introducing some kind of decorator with is aware of the query. We can add such caching layer when performance problem will arrise.

app/models/filter/ancestry_multiselect.rb
mkasztelnik: * `Hash.new { |h, k| h[k] = Set.new }` - create a new hash with initialization block, so each time you will try to access  hash element which does not exist yet it will be initialized with empty `Set`
* `inject(...) { |h, (k, v)| h[k] << v; h }` reduce array of 2 dimentional arrays (e.g. `[research_area_id, service_id]`) into hash where key is the `research_area_id` and the values is a `Set` with `service_ids`

app/models/filter/ancestry_multiselect.rb
mkasztelnik: Yes, we do, because of convention over configuration. This will be valid for 95% (right now for 100%), for the rest 5% you can also override this method.

bwilk: 👍

app/models/filter/ancestry_multiselect.rb
bwilk: Thanks for the clarification - got it. Do you know where I can read more about the first example. The second is pretty straight forward.

app/models/filter/ancestry_multiselect.rb
bwilk: 👍

app/models/filter/ancestry_multiselect.rb
mkasztelnik: Take a look at [Hash](https://ruby-doc.org/core-2.5.3/Hash.html#method-c-new) documentation for `new` method.

PR #846, opened 2019-04-30
[#783] Service counters fix. Filters and categories moved to elasticsearch

### LINKED ISSUE #783: As a user when searching for service I want to see item count related to number of search results.
#### 1. Query entered by the user in the search bar

This is the main querying mechanism and all other levels need to receive result scope of this query

#### 2. Categories

Category item shows a number of services in a category and all subcategories limited by query scope (from point 1). When a category is selected query scope need to limit services only to services from the selected category.

#### 3. Filters

> :warning: This is the most complicated element :warning: 

For every filter, we need to calculate how many services are in each filter element without this filter applied. So for each filter, we need to take scope from point 2, add constants from other filters and do aggregation for the selected filter.

**Optimization** not selected filters can be grouped together and joined aggregation query can be applied once. 

#### 4. Shown services

The scope from point 2 should be taken and all selected filters should be applied.

#### 1 elasticsearch query - no filters selected

This is possible only when filters are not selected, category and query string can be present. In this situation, one elasticsearch query with aggregation for categories and all filters can be sent. 

#### 2-N+1 elasticsearch queries - 1 or more filters selected

 - First elasticsearch query is done to calculate categories count
 - The second elasticsearch query is done to calculate filters count for all unselected filters
 - N elasticsearch queries (where N is a number of selected filters) is done to calculate selected filter items counts.
  - N+1 elasticsearch query is done with all filters values is sent to show services list 

### Caching possibilities

Caching can speed up service browsing a lot, especially, when the user is navigating through the searched services (paging).  Caching can be done on different levels (category counts, filter counts).

@roksanaer, @bwilk, @abacz, @agpul, @kmarszalek, @martaswiatkowska, @michal-szostak comment if something in this description is wrong, not clear. I will try to improve this description and finally update this issue description.

### Filter items sorting and visibility

  1. Items are sorted using `count` value. If the filter is tree based subtrees are sorted in a similar manner.
  1. No items are removed from the list, what is more, a user should be able to select item even when currently count is equal to 0. The reason for this behavior is that this count could be changed on other filters changes.

### Filter cleaning

Filters are cleaned only when search query changes.

bwilk:
Fixes #783

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -6,6 +6,7 @@ module Service::Searchable
   included do
     include Paginable
     include Service::Sortable
+    include Service::Categorable
   end
 
   def search(search_scope, filters)
@@ -68,17 +69,35 @@ module Service::Searchable
           scope_results: ->(r) { r.includes(:research_areas, :providers, :target_groups).with_attached_logo }
       ]
     end
+  end
 
-    def id_constr(search_scope, constr = {})
-      constr.merge!({ id: search_scope.ids })
+  def category_counters(search_scope, filters)
+    services = search_for_categories(search_scope, filters)
+    counters = services.aggregations["categories"]["categories"]["buckets"].
+        inject({}){ |h, e| h[e["key"]]=e["doc_count"]; h}
+    counters.tap { |c| c[nil] = services.aggregations["categories"]["doc_count"] }
+  end
+
+  private
+
+    def query_present?
+      params[:q].present?
+    end
+
+    def query
+      query_present? ? params[:q] : "*"
+    end
+
+    def scope_constr(search_scope, constr = {})
+      constr.tap { |c| c[:id] = search_scope.ids.uniq }
     end
 
     def category_constr(constr = {})
-      constr.tap { |c| c[:categories] = category.id unless category.nil? }
+      constr.tap { |c| c[:categories] = category.descendant_ids + [category.id] unless category.nil? }
     end
 
     def filter_constr(filters, constr = {})
-      filters.reduce(constr) { |h, f| h.merge(f.constraint) }
+      filters.reduce(constr) { |c, f| c.merge(f.constraint) }
     end
 
     def highlights(from_search)
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -10,71 +10,52 @@ module Service::Searchable
   end
 
   def search(search_scope, filters)
-    Service.search *(params_for_search(search_scope, filters))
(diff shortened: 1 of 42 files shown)


app/controllers/concerns/service/searchable.rb
```diff
@@ -18,16 +22,29 @@ def query_present?
...
+          fields: [ "title^5", "description"],
+          operator: "or",
+          where: { id: search_scope.ids },
+          page: params[:page],
+          per_page: per_page,
+          order: params[:sort].blank? ? nil : ordering,
+          match: :word_middle,
+          highlight: { fields: [:title], tag: "<b>" },
+          scope_results: ->(r) { r.includes(:research_areas, :providers, :target_groups).with_attached_logo }
+      ]
```
bwilk: one of these methods will decorate the other in the target impl

app/controllers/concerns/service/searchable.rb
```diff
@@ -18,16 +22,29 @@ def query_present?
...
+      [
+          params[:q],
+          fields: [ "title^5", "description"],
+          operator: "or",
+          where: { id: search_scope.ids },
+          match: :word_middle,
+      ]
+    end
+
+    def complex_params(search_scope)
```
bwilk: this method should decorate the `simple_params` in the target impl

app/models/filter/provider.rb
```diff
@@ -7,10 +7,24 @@ def initialize(params = {})
           field_name: "providers",
           title: "Providers",
           query: ::Provider.select("providers.name, providers.id, count(service_providers.service_id) as service_count"))
+    @filter_scope = params[:filter_scope]
   end
 
   private
 
+    def fetch_options
```
bwilk: here I should do some magic trick just as @mkasztelnik in his recent implementation of hierarchical multiselect, but for now it is as it is

db/schema.rb
```diff
@@ -90,13 +90,36 @@
...
     t.boolean "voucherable", default: false, null: false
-    t.string "status"
     t.string "offer_type"
+    t.string "status"
     t.index ["iid"], name: "index_offers_on_iid"
     t.index ["service_id", "iid"], name: "index_offers_on_service_id_and_iid", unique: true
     t.index ["service_id"], name: "index_offers_on_service_id"
   end
 
+  create_table "order_changes", force: :cascade do |t|
```
bwilk: didn't make any changes here - I need to check why it was commited...

bwilk:
I think it'd be wise to create Elasticsearch for categories and get rid of `Categorable` concern - both searches used make use of categories so this can be completed within a single query. Maybe filters can be moved there as well, but for now I don't see any significant impact of this change.

PR #857, opened 2019-05-09
Elasticsearch index changes

### LINKED ISSUE #856: Elasticsearch index and highlights for service tagline
**Describe the bug**
Instead of tagline we use descrtiption for search 

**To Reproduce**
In `app/views/services/_service.html.haml` we use `%p.mb-3= service.tagline` - check it yourself. This field should be indexed and used for elasticsearch queries. 

**Expected behavior**
Tagline should be used instead. Selected services should have highlights for tagline.

martaswiatkowska:
Added tagline to elasticsearch index
Setted different weights for elasticsearch fields
Added highlights for tagline
Changed autocomplete routing and fixed autocomplete function

Fix #856

### DIFF
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -47,4 +47,8 @@ module ServiceHelper
   def service_title(service, highlights)
     highlights&.dig(service.id, :title)&.html_safe || service.title
   end
+
+  def service_tagline(service, highlights)
+    highlights&.dig(service.id, :tagline)&.html_safe || service.tagline
+  end
 end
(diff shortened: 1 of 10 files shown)


app/helpers/service_helper.rb
```diff
@@ -47,4 +47,8 @@ def providers_text(service)
   def service_title(service, highlights)
     highlights&.dig(service.id, :title)&.html_safe || service.title
   end
+
+  def service_tagline(service, highlights)
```
mkasztelnik: This method is similar like `service_title`, you can extract it into separate method and reuse code.

PR #864, opened 2019-05-16
Deleted elasticsearch scope_import and use pundit scope

martaswiatkowska:
The pundit scope was used instead of searchkick scope to resolve problems with not showing draft services after search in backoffice. That's why, the autocomplete concer was turned into a helper.

### DIFF
--- a/app/controllers/concerns/service/autocomplete.rb
+++ b/app/controllers/concerns/service/autocomplete.rb
@@ -10,6 +10,7 @@ module Service::Autocomplete
       match: :word_middle,
       limit: 5,
       load: false,
+      where: { id: scope.ids },
       highlight: { tag: "<b>" }
     )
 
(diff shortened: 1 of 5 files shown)


app/helpers/autocomplete_helper.rb
```diff
@@ -1,24 +1,22 @@
...
-  def autocomplete
-    query = Service.search(params[:q],
-      fields: ["title"],
-      operator: "or",
-      match: :word_middle,
-      limit: 5,
-      load: false,
-      highlight: { tag: "<b>" }
-    )
+module AutocompleteHelper
```
mkasztelnik: I'm not sure why this change was performed, can you explain it? Looking into code all `autocomplete` can be done in the concern and injected to both front and back office service controllers.

app/controllers/concerns/service/autocomplete.rb
```diff
@@ -3,13 +3,14 @@
 module Service::Autocomplete
   extend ActiveSupport::Concern
 
-  def autocomplete
+  def autocomplete_in_scope(scope)
```
mkasztelnik: Please consider renaming method signature into `authenticate`, then no controller modification will be needed - only concern inclusion

Issue #834, opened 2019-04-26
Add the possibility to create a project from My project and orders


agpul:
**Is your feature request related to a problem? Please describe.**
Now we can create a project only during service ordering or adding service to the project (configuration step for eg: https://marketplace.eosc-portal.eu/services/cloudferro-infrastructure/configuration )

**Describe the solution you'd like**
According to https://kcjh3g.axshare.com/#g=1&p=my-services-list we would like to have the possibility to create a project from My project and orders tab: https://marketplace.eosc-portal.eu/projects

Created project is empty. After project creation, we are redirected to My project and orders tab: https://marketplace.eosc-portal.eu/projects

PR #865, opened 2019-05-16
[#834] Create new project on projects page

martaswiatkowska:
Project can be created from My projects page.

Fix [#834]

### DIFF
--- a/config/breadcrumbs/marketplace.rb
+++ b/config/breadcrumbs/marketplace.rb
@@ -57,3 +57,8 @@ crumb :congratulations do |project_item|
   link "Congratulations", project_item_path(project_item)
   parent :marketplace_root
 end
+
+crumb :project_new do
+  link "New project", new_project_path
+  parent :projects
+end
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -30,20 +30,33 @@ class ProjectsController < ApplicationController
   end
 
   def new
+    @project = Project.new(user: current_user)
+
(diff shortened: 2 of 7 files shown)


agpul:
@roksanaer adding new fields is the topic of another ticket

roksanaer:
> @roksanaer adding new fields is the topic of another ticket

ok, so only the typo issue :) also, notice the filtering next to the create project button. it should not stay the same, or clearly indicate it filters the orders, not projects. but this is a remark not in the scope of this PR ;) sorry!

roksanaer:
@mkasztelnik @kmarszalek or @bwilk please perform the review of this PR

martaswiatkowska:
@roksanaer  I can add tip "Url should start with http or https [e.g. http://webpage.org]" as is in service form

abacz:
We didn't implement GUI yet. It is only development work.

agpul:
@roksanaer we discussed this topic because we had the same problem with http:// and https:// in backoffice and we can add only a tip because we don't know if user want to add a link with http or https, so @martaswiatkowska please add the tip as we have in other http fields

config/breadcrumbs/marketplace.rb
```diff
@@ -57,3 +57,8 @@
   link "Congratulations", project_item_path(project_item)
   parent :marketplace_root
 end
+
+crumb :project_new do
+  link "New projects", new_project_path
```
mkasztelnik: Here we are creating a single project, so "New project"

Issue #854, opened 2019-05-08
Toggle instead of Collapse All button for filters


bwilk:
**Is your feature request related to a problem? Please describe.**
Once filters have been collapsed there is no easy way to expand them back

**Describe the solution you'd like**
The button should be named Expand All after collapsing  

**Describe alternatives you've considered**
none

**Additional context**
<img width="1680" alt="Zrzut ekranu 2019-05-8 o 15 57 49" src="https://user-images.githubusercontent.com/2674939/57381269-e3b24180-71aa-11e9-8eb7-8fe03093d678.png">

PR #886, opened 2019-05-31
collapse/expand link fix

jarekzet:
Collapse All link mod. Now we have Collapse/Expand link on top of the filter panel. 
Issue #854


michal-szostak:
I fixed the issue, and added test. The test itself contains small hack, but general solution would be disabling animations or something similar

PR #887, opened 2019-05-31
Show error message if logo is in invalid format

martaswiatkowska:
The ActiveStorage::InvariableError are rised when image cannot be transformed (resized). Used "variable?" to determine that. This is related to the logo format.
Links to documentation  https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blob/Representable.html#method-i-variable-3F
Supported formats: http://www.searchcodeserver.com/file/61f1bac55070369392fc3ee4b289f2c4dcac12cb/rails/activestorage/lib/active_storage/engine.rb#25


agpul:
image
@martaswiatkowska can we delete text; "Logo" before Roksana message? if yes, please do that if no we can go with this Pr

Issue #787, opened 2019-04-01
Transfer fields related to customer country to JIRA 


agpul:
As a result of this ticket, I expect to have fields: Country of the customer, Country of collaboration
transferred to JIRA.  Fileds should be propagated to both Master ticket (project level) and sub ticket (order level). 
Parameters of this field in JIRA:
wait for Mal 

This field in JIRA need to support statistics and Jira needs to be able to analyse these fields. That's why we consider making this filed in Jira as select type. For now, we consider to support plugin in Jira but the topic is still unrecognized. @mkasztelnik has some comments on this topic, you can write your ideas in this ticket

agpul:
fields in JIRA;

CP-CustomerCountry  customFieldId=11204
CP-CollaborationCountry customFieldId=11203

roksanaer:
@goreck888 is this ticket still valid (it's in in progress state)

PR #889, opened 2019-06-03
[#787] transfer country fields to jira

goreck888:
Add possibility to create tickets in jira with country of customer
and country of collaboration fields added to project.


app/models/country.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def self.customer_countries_list
+    get_countries(NON_EUROPEAN)
+  end
+
+  def self.countries_list
+    get_countries(INTERNATIONAL, NON_EUROPEAN)
+  end
+
+  def self.validator_countries_list
```
bwilk: I still think this should end up in the validator if it's called validator_countries_list. Contrary maybe changing the name would make up for this mismatch - maybe "allowed_countries" or similar would fit here. 

You don't need "_list" in this name no matter how you approach earlier comments. Ruby is always trying to be concise - let's respect this approach. Besides Ruby conventions this is not a list from the data type perspective - it's an array ;)

app/models/project.rb
```diff
@@ -11,9 +12,20 @@ class Project < ApplicationRecord
             presence: true,
             uniqueness: { scope: :user, message: "Project name need to be unique" }
 
+  validates :country_of_customer, presence: true, inclusion: { in: Country.validator_countries_list }
```
bwilk: could you unit-test all the validations in this model?

app/models/country.rb
goreck888: Ok, now it should be ok :) I changed the names of these arrays

app/models/project.rb
goreck888: Yes, of course, in this branch I'm focused on jira synchronization with new country fields and fix unit tests concerning this aspect of my task. This validator unit test will be written in the base branch 785-information-about-country, PR #872 and then 787 will be rebased and merged.

Issue #837, opened 2019-04-26
Transfer fields from configuration step to the project


agpul:
**Is your feature request related to a problem? Please describe.**
Now when a user creates the project he/she needs to fill only: Name, Reason for access, Customer typology and related to customer topology fields. 

**Describe the solution you'd like**
We would like to transfer some fields from Service configuration step (https://marketplace.eosc-portal.eu/services/egi-cloud-compute/configuration). According to https://kcjh3g.axshare.com/#g=1&p=create_new_project
add to the project:
- Additional information
This fields should be removed from the configuration step but configuration step should have static information about these fields. According to axshare, when user chooses a project on configuration step the view should be like: https://kcjh3g.axshare.com/#g=1&p=request_egi_cloud_-step-2-configuration
Change names of existing fields from:
- Name-> Project name
- Reason for  access -> Reason to request access to the EOSC services

agpul:
@martaswiatkowska, for now, we agreed that we don't touche the affiliation

agpul:
After meeting from 28.05:
- the affiliation shouldn't be moved to the project (we will touch the affiliation problem during the "profile sprint", the affiliation will be hidden in the first iteration from the main view but not deleted, then the migration will take place and in the end, we will remove affiliation)
- in this sprint we should add fields to the customer topology from user profile (this fields should be added to the single researcher option and representing the research community:
   - email address (optional)
   - organization (mandatory)
   - department (optional)
   - webpage (mandatory)

@roksanaer check and confirm

roksanaer:
it seems ok. Howevert, affiliation should be removed from any object in the order scope (order firm, project item view etc.) and probably this should be a new ticket. In this ticket applies the same as Agnieszka has written above
1. removal of the affiliaiton from the views
2. migration of the data from affiliation to the customer typology field (none of the fileds from the affiliation will be mapped to new profile structure, so no migration will be needed there).
3. removal of affiliations from the system.

@agpul this also implies removal of the affiliation-connected fields in JIRA and extension of the customer typology field in JIRA. We need to have a ticket for Mal for that. This is very important for MP-JIRA inetgration part. Also as we are planning to have a Nationality field in profile, this probably might end up in JIRA.
So, the JIRA issue templates need to be updated accordingly (the master ticket template) and the documentation in confluence (master and sub ticket hierarchy) as well.

roksanaer:
[...]

PR #893, opened 2019-06-04
Delete duplicated project fields from project_item

martaswiatkowska:
Its a first part of implementation #837 
On configuration step project fields are static. That's why removed duplicated project fields ( access_reason, custom_typology and connected) from project item. 
Reason_for_access and customer_typology are required to create a project.


app/javascript/controllers/project_controller.js
```diff
@@ -0,0 +1,47 @@
...
+                    "project", "privateCompany", "input"];
+
+  connect() {
+  }
+
+  initialize() {
+    this.CUSTOMER_TYPOLOGIES = { "single_user": "single_user",
+                               "research": "research",
+                               "private_company" : "private_company",
+                               "project" : "project" }
```
michal-szostak: Generally if dict fields are valid JS property names they should not be enclosed in quotes

app/javascript/controllers/project_item_controller.js
```diff
@@ -1,61 +1,41 @@
...
+    this.fetchProjectData(this.projectTarget.value)
   }
 
   projectChanged(event) {
-    fetch(this.data.get("url") + "/" + event.target.value, { dataType: "json" })
+    this.fetchProjectData(event.target.value)
+  }
+
+  fetchProjectData(value) {
+    fetch(this.data.get("url") + "/" + value, { dataType: "json" })
```
michal-szostak: I remember discussion we had regarding use of es5 `fetch` and that it would have been better to use request methods provided by rails js component (which works better with rails' `format.js` etc)

PR #904, opened 2019-06-06
[#837] Move additional_information to Project

### LINKED ISSUE #837: Transfer fields from configuration step to the project

martaswiatkowska:
Moved additional_information to project
Name-> Project name
Reason for access -> Reason to request access to the EOSC services

Fix #837

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Changed
 - Improve empty project description (@mkasztelnik)
 - Improve collapse / expand all behaviour for filters (@michal-szostak)
+- Move additional_information from project_item to project (@martaswiatkowska)
 
 ### Removed
 - Remove rating filter from the view until we have enough ratings to show it (@mkasztelnik)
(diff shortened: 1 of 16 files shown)


db/migrate/20190517091846_remove_fields_from_project_item.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveFieldsFromProjectItem < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :project_items, :additional_information, :text
```
bwilk: shouldn't we migrate these information to Project instead of just dropping them?

Issue #891, opened 2019-06-03
Each user can see draft service using a direct link 


agpul:
**Describe the bug**
When the user has a link to service which has a draft state, he/she can see this service although it should be forbidden. 

**To Reproduce**
1. Go to https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services/draft
2. Make sure that you are logged out or you don't have service portfolio manager role
3. See the service that you shouldn't be able to see

**Expected behavior**
We shouldn't show services which have draft state to the normal user. 
We can consider showing such a link if the user has a special role eg. service portfolio manager -> @roksanaer

PR #905, opened 2019-06-06
[#891] Block access to the draft service

goreck888:
**What's in this PR?**
This PR fixes bug which is described in https://app.zenhub.com/workspaces/cyfronet-fid-5b1e4b786c29461038c37df5/issues/cyfronet-fid/marketplace/891 

Changes:
Add authorization for services
Add policy blocking access to the draft if user isn't service_portfolio_manager, 
because this is a kind of sample service view and I won't block to allow managers to get this view.

### DIFF
--- a/app/policies/service_policy.rb
+++ b/app/policies/service_policy.rb
@@ -8,7 +8,7 @@ class ServicePolicy < ApplicationPolicy
   end
 
   def show?
-    true
+    record.published?
   end
 
   def order?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Blocked access to the draft service via direct link
 
 ### Security
 
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -19,6 +19,7 @@ class ServicesController < ApplicationController
     @service = Service.
                includes(:offers, related_services: :providers).
                friendly.find(params[:id])
+    authorize @service
     @offers = policy_scope(@service.offers)
     @related_services = @service.related_services
 
--- a/config/locales/pundit.en.yml
+++ b/config/locales/pundit.en.yml
@@ -4,3 +4,7 @@ en:
     affiliation_policy:
       "edit?": "You cannot edit an affiliation which has been activated"
       "destroy?": "You cannot remove an affiliation which has a project item"
+    service_policy:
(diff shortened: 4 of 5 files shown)


app/policies/service_policy.rb
```diff
@@ -3,12 +3,12 @@
 class ServicePolicy < ApplicationPolicy
   class Scope < Scope
     def resolve
-      scope.where(status: :published)
+      scope
```
martaswiatkowska: If you remove this line draft services will be shown in autocomplete because autocomplete use pundit scope pr - https://github.com/cyfronet-fid/marketplace/pull/864/files
Remove this change and in schow? add
``` 
record.published?

Comment (goreck888) [reply app/policies/service_policy.rb]: Yeah, I know my mistake. Already done.

**Comment by goreck888** [comment]
> according to my manual tests - it does not work. I've created 2 draft services
> obraz
> which are visible and searchable in the MP part for a user without the portfolio manager or service owner role:
> obraz
> obraz
> 
> so the fix is not working at all :)
> 
> But! it has to be given that - trying to access the service entry as a user without the portfolio manager role i get the access deny info:
> 
> obraz
> 
> when trying to reach it as portfolio manager in the mp part, i can:
> obraz
> 
> It looks like it the fix has not been checked after the dev at all, please do it next time.
> Draft service should not be visible, findable or included in MP counters for any role - even with portfolio manager role one should not see draft services in the MP part.

Now everything should work fine. I blocked access to the view for all roles and now drafts shouldn't be visible in the category view.

**Comment by goreck888** [comment]
> It works much better right now :)
> as a portfolio manager i can see and edit draft services only from the back office.
> 
> Only one question - right now, when i try to enter the draft service entry in the MP user part (by pasting the URL) i get Unauthorised access message (You are not allowed to ...). In my opinion, it is not the right message. Can we change the message shown to:
> "This service is not published in the Marketplace yet, therefore it cannot be accessed. If you are the Service Owner or Service Portfolio Manager and wish to manage this service, please log in and go to the Backoffice tab."
> 
> If anyone has a better proposition for this, please do share it here :)
> @abacz @bwilk @agpul @martaswiatkowska

I think it may be a very good idea to create custom error message, but now there's only one, default Not authorized message. Custom messages can be a new feature, because it's necessary to change code of policies and application controller.

**Comment by martaswiatkowska** [comment]
You can change this in config/locales/pundit.en.rb. You just need to add 
``` 
service_policy:
    "show?": "new message"

agpul:
Ok so if it is easy to add I will not make a different ticket. Please add this message in this PR.

agpul:
for my point of view, it works ok

PR #906, opened 2019-06-06
Quickfix errors with roles check in the back-end

goreck888:
This commit contains check if user exist by checking the role
of user. It's necessary because ruby tries to call not existing
method fo a NilClass.

### DIFF
--- a/app/policies/backoffice/category_policy.rb
+++ b/app/policies/backoffice/category_policy.rb
@@ -40,6 +40,6 @@ class Backoffice::CategoryPolicy < ApplicationPolicy
   private
 
     def service_portfolio_manager?
-      user.service_portfolio_manager?
+      user&.service_portfolio_manager?
     end
 end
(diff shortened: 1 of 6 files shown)


app/policies/backoffice/category_policy.rb
```diff
@@ -40,6 +40,6 @@ def permitted_attributes
   private
 
     def service_portfolio_manager?
-      user.service_portfolio_manager?
+      user && user.service_portfolio_manager?
```
martaswiatkowska: user&.service_portfolio_manager? isn't simpler ?

goreck888: Yes, it is. I didn't know that notation. Thanks!

agpul:
Can you help me and tell me how can I test this PR?

goreck888:
Try to go to the backoffice without login and ask @wziajka about logs.

PR #907, opened 2019-06-07
[841] Edit and delete project

### LINKED ISSUE #841: Add possibility to edit and delete the project
Now, we can't edit and delete an existing project. 

We would like to have two functionalities:
1. DELETE
- user can delete **an empty project** from project detail view: 
https://kcjh3g.axshare.com/#g=1&p=view_empty_project_-_uj_class_1d
- user can't delete a project related to service (no matter of service state)
2. EDIT
- user can edit a project from project detail view: https://kcjh3g.axshare.com/#g=1&p=view_empty_project_-_uj_class_1d
- user can edit **all** fields at any moment (this is a first implementation. it is possible that in the future, there will be a set of mandatory fields, that after filling in and creating a project will not be editable. One of the candidates, that should not be changed is Customer Typology)

martaswiatkowska:
Added possibility to edit and delete project from project details view

Fix #841

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -2,6 +2,7 @@
 
 class ProjectsController < ApplicationController
   before_action :authenticate_user!
+  before_action :find_and_authorize, only: [:show, :edit, :update, :destroy]
 
   def index
     @projects = policy_scope(Project).order(:name).eager_load(:project_items)
@@ -10,9 +11,6 @@ class ProjectsController < ApplicationController
   end
 
   def show
-    @project = Project.find(params[:id])
-    authorize(@project)
-
     respond_to do |format|
       format.json do
         render status: :ok, json: {
@@ -69,6 +67,25 @@ class ProjectsController < ApplicationController
     end
   end
 
+  def edit
+    @show_as_modal = false
+  end
+
+  def update
+    if Project::Update.new(@project, permitted_attributes(@project)).call
+      redirect_to project_path(@project),
+                  notice: "Project updated correctly"
+    else
+      render :edit, status: :bad_request
+    end
+  end
+
+  def destroy
+    Project::Destroy.new(@project).call
+    redirect_to projects_path,
+                notice: "Project destroyed"
+  end
+
   private
     def filterable?(param)
       (param.present? && ProjectItem.statuses.has_key?(param))
@@ -81,5 +98,10 @@ class ProjectsController < ApplicationController
                 action_btn: "Create new project",
                 form: "projects/form"
               }
+    end
+
+    def find_and_authorize
+      @project = Project.find(params[:id])
+      authorize(@project)
     end
 end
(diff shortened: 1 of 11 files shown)


config/locales/pundit.en.yml
```diff
@@ -1,6 +1,8 @@
 en:
   pundit:
     default: "You are not authorized to see this page"
+    project_policy:
+      "destroy?": "You cannot remove an project witch has a project item"
```
kmarszalek: typo

app/controllers/projects_controller.rb
```diff
@@ -67,6 +65,25 @@ def create
     end
   end
 
+  def edit
+    @show_as_modal = false
```
bwilk: forgive my ignorance but - where are you using this field? is it used elsewhere?

app/controllers/projects_controller.rb
```diff
@@ -67,6 +65,25 @@ def create
...
+  end
+
+  def update
+    if Project::Update.new(@project, permitted_attributes(@project)).call
+      redirect_to project_path(@project),
+                  notice: "Project updated correctly"
+    else
+      render :edit, status: :bad_request
```
bwilk: Please check negative path in a test. This can be either Project::Update unit test or feature integration test or both, but until it's implemented we don't know what happens when .call returns false

app/controllers/projects_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class ProjectsController < ApplicationController
   before_action :authenticate_user!
+  before_action :find_and_authorize, only: [:show, :edit, :update, :destroy]
```
bwilk: we don't have any tests of anauthorized access neither in this PR nor in the other specs - it's advised to implement any form of anauthorized access testing

roksanaer:
>>If she/he wants to DELETE project with project_item we have a message: "You cannot remove an project witch has a project_item" which contain typos and I think that project_item is not easy to understand for user point of view.
Or maybe we should consider to hide or make inactive this button @abacz @roksanaer

I believe it should be hiden from the user interface

we might add an information somewhere in the view like:
"Projects with pending or active service access request cannot be deleted or archived"

What is more, @agpul, did we include the archivisation of the project. Meaning: if the project has only rejected or ended orders, it should be possible to be archived. Of course it can be the same action: delete (before adding the orders). Just make sure it's active in this state of the project.

In the end, I would only think about the name of the button. @abacz @agpul @bwilk  Delete might be not the best suit. "Archive", "Move to archived projects" maybe? it has to imply that the project will not be lost. In the future we are plannung a dedicated view with a chronological order of the projects and orders, where archived projects will be available to the user.

abacz:
@roksanaer Archive sounds better.

agpul:
OK so archive action will be a scope of another PR, here you can see the ticket: https://github.com/cyfronet-fid/marketplace/issues/920

So @martaswiatkowska it the scope of this PR, please add only the tip and make archive button in another PR

PR #917, opened 2019-06-12
[#783] Service counters fix. Filters and categories moved to elasticsearch 

### LINKED ISSUE #783: As a user when searching for service I want to see item count related to number of search results.

bwilk:
Fixes #783 
Description in the ticket itself

Changed:
- Reimplemented filters and categories after creating indexes in Elasticsearch 

Fixed:
- Category and filter counters 
- Hierarchical filters deactivation buttons
- Choice of "best match" sorting strategy after search 

Moved redirect after selecting a service in the search field after autocompletion: https://github.com/cyfronet-fid/marketplace/pull/916

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -6,18 +6,57 @@ module Service::Searchable
   included do
     include Paginable
     include Service::Sortable
+    include Service::Categorable
   end
 
   def search(search_scope, filters)
-    Service.search *(params_for_search(search_scope, filters))
+    Service.search(query,
+                   fields: [ "title^7", "tagline^3", "description"],
+                   operator: "or",
+                   match: :word_middle,
+                   where: filter_constr(filters, scope_constr(search_scope, category_constr())),
+                   page: params[:page],
+                   per_page: per_page,
+                   order: ordering,
+                   highlight: { fields: [:title], tag: "<b>" },
+                   scope_results: ->(r) { r.includes(:research_areas, :providers, :target_groups).with_attached_logo })
   end
 
   def search_for_filters(search_scope, filters, current_filter)
-    Service.search *(params_for_filters(search_scope, filters, current_filter))
+    Service.search(query,
+                   fields: [ "title^7", "tagline^3", "description"],
+                   operator: "or",
+                   match: :word_middle,
+                   where: filter_constr(filters.reject {|f| f == current_filter}, scope_constr(search_scope, category_constr())),
+                   aggs: (!current_filter.index.blank? ? [current_filter.index] : []),
+                   load: false)
   end
 
   def search_for_categories(search_scope, filters)
-    Service.search *(params_for_categories(search_scope, filters))
+    Service.search(query,
+                   fields: [ "title^7", "tagline^3", "description"],
+                   operator: "or",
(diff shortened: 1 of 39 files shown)


app/models/filter/location.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 class Filter::Location < Filter
+  #   TODO finish this filter
```
kmarszalek: TODO?

bwilk: yep :)

app/controllers/concerns/service/searchable.rb
```diff
@@ -6,10 +6,57 @@ module Service::Searchable
...
     include Paginable
     include Service::Sortable
+    include Service::Categorable
   end
 
-  def search(search_scope)
-    query_present? ? search_fields(search_scope) : paginate(order(search_scope.distinct))
+  def search(search_scope, filters)
+    Service.search(query,
+                   fields: [ "title^7", "tagline^3", "description"],
```
michal-szostak: I think we could try to extract constant & common part of the `Service.search` call:

```
{
fields: [ "title^7", "tagline^3", "description"],
operator: "or",
match: :word_middle,
}
```

This would have an added benefit of extraction of "magic constants" (`"title^7"` etc)

bwilk: Done

PR #933, opened 2019-06-18
[#785] Store information about country of customer

bwilk:
Add CountryHelper for create country list
Add new columns in the database
Add gem country


agpul:
On marketplace side, everything looks OK but when I create a project in MP, the Epic issue in JIRA isn't created (it works on Wednesday) @michal-szostak can you have a look?

PR #936, opened 2019-06-21
Migrate affiliation fields to project

### LINKED ISSUE #903: Move fields from affiliations to the project and hide them from the user's view
We plan to delete affiliation soon, so please move fields from affiliation to the project. Relevant fields that need to be moved:
- email address (mandatory) - there should be a hint: email address in your institute's domain, move directly to the project
- organization (mandatory) - move under customer topology (single user and research community)
- department (optional)  - move under customer topology (single user and research community)
- webpage (mandatory) - hint: webpage of the institute or department  - move under customer topology (single user and research community)

Fr now, all these fields should be possible to edit in every moment.

As the second step of this issue, we need to hide from the user view all fields related to the old way of presenting the affilaition, I mean from:
- profile: https://marketplace.eosc-portal.eu/profile (hide all affilation section)
- service configuration step eg; https://marketplace.eosc-portal.eu/services/egi-notebooks/configuration

**Additional context**
We need to take into consideration that some users use these fields in an old way. Make sure that we can simply move these fields from affiliation to the project if no please duplicate them and we will delete them after migration.

mkasztelnik:
In the scope of this PR following changes are applied:
  * [x] affiliations were removed from all views (to make it happened relation to affiliation is changed to optional when needed)
  * [x] required fields (`email`, `organization`, `department`, `webpage`) were added to project
  * [x] migration rack task created to move affiliation data to project
  * [x] jira registration is using data from the project instead of affiliation

Fixes #903


roksanaer:
add an option in Customer Typology 'not specified' (but this option should not be possible to choose by the user in GUI) and apply this value to all unspecified fields. Send an email to users with unspecified projects asking them to fill in the missing data

mkasztelnik:
@wziajka after this is released deploy procedure should look like follows:
  1. background jobs processing should be stopped (to avoid project registration in jira)
  1. migrations are performed
  1. `RAILS_ENV=production rails projects:migrate_from_affiliation` is executed
  1. background job processing can be started now

It would be nice to have some kind of file describing this kind of step needed when the execution of a dedicated rake task is needed. Any ideas where we can put it?

mkasztelnik:
One last element is missing here -> reorganization of the jira client to use project information instead of affiliation.

PR #938, opened 2019-06-21
Project box improvements

### LINKED ISSUE #931: Changes in project box
We agreed that we should make two small changes in project box according to @abacz prev (please attach if it is ready):
-  change button name from Details to Manage and support
-  Add a link "support" which will redirect the user to Chat view

mkasztelnik:
* Details button name changed to "Manage and support"
  * Link to chat with support added

Link to open access service was not added because of the way how service item looks like now:

project

This change deserves for another PR.

Fixes #931


bwilk:
Please review this name "Manage and Support" - this means "Zarządzaj i Wesprzyj". "Management and Support" means - "Zarządzanie i Wsparcie" - which is more accurate but probably too long.

mkasztelnik:
@roksanaer, @agpul  please provider final label for the button mentioned by @bwilk

roksanaer:
I would go with either one of the two:
 'Manage and get consultancy' or
 'Manage and get support'

I prefer consultancy as it better reflects what we want to give the user there but maybe it's too... something :D @agpul @bwilk @abacz ?

agpul:
I vote for 'Manage and get support'

abacz:
I have an impression that text on button (get support)  will trigger an action which I don't want.
But it is an impression.

PR #939, opened 2019-06-24
[#919] Stay in project modal when project empty while ordering

### LINKED ISSUE #919: Clicking twice "Create' button during creating a new project causes incorrect redirection
**Describe the bug**
When we click twice on the "Create" button during creating a new project from service ordering form we are redirecting from the view and we lost information about the service

**To Reproduce**
1. Go to any service eg. Cloud Ferro 
2. Click on 'Add to a project'
3. Click "Add new project"
4. See the empty form

5. Click "Create" while the form is empty or form doesn't have filled all relevant fields 
 
6. Click "Create" one more time and see
 error

**Expected behavior**
Clicking the "Create" button should always leave a user on the current form

**Additional context**
You can check this bug on Sprint PR: https://marketplace-14.docker-fid.grid.cyf-kr.edu.pl/

bwilk:
Redirection to project edit view on project creation failure while ordering is fixed. User stays in this modal after fix:

Zrzut ekranu 2019-06-24 o 12 14 47

Fixes #919

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -33,11 +33,8 @@ class ProjectsController < ApplicationController
     @project = Project.new(user: current_user)
 
     respond_to do |format|
-      format.html do
-        @show_as_modal = false
-      end
+      format.html
       format.js do
-        @show_as_modal = true
         render_modal_form
       end
     end
@@ -69,7 +66,6 @@ class ProjectsController < ApplicationController
   end
 
   def edit
-    @show_as_modal = false
   end
 
   def update
@@ -93,6 +89,7 @@ class ProjectsController < ApplicationController
     end
 
     def render_modal_form
+      @show_as_modal = true
       render "layouts/show_modal",
               locals: {
                 title: "New project",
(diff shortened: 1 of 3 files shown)


spec/features/order_spec.rb
```diff
@@ -261,6 +261,26 @@
...
+
+      click_on "Order"
+      click_on "Add new project"
+
+      click_on "Create new project"
+      within("#ajax-modal") do
+        expect(page).to have_button("Create new project")
+      end
+
+      click_on "Create new project"
```
mkasztelnik: Why do you need to validate this action twice?

app/controllers/projects_controller.rb
```diff
@@ -62,6 +62,7 @@ def create
           Project::Create.new(@project).call
           render :show
         else
+          @show_as_modal = true
```
mkasztelnik: Maybe this is a good time to get rid of this flag?

spec/features/order_spec.rb
bwilk: TDD approach here - because this is the method to reproduce the bug I'm fixing. First you try to save empty, then you try again - this is when the old implementation first.

app/controllers/projects_controller.rb
bwilk: I've reduced the use of it to minimum.

Issue #842, opened 2019-04-26
Add customer service functionality in project details


agpul:
***Review this functionality and change from "chat" to "customer service" everywhere*** (front, links, code)

This ticket is blocked by #768 
**Is your feature request related to a problem? Please describe.**
Now in the project_item, we have a possibility to contact with customer support which is linked with EOSCSO project in JIRA and is called "Order history"

**Describe the solution you'd like**
When we will split projects in JIRA on the master ticket (related to the project) and sub ticket (project_item) we will need:
- [ ] make sure that order history in project item to be related to sub ticket
- [ ] add **Customer Service** for the customer as here:  https://kcjh3g.axshare.com/#g=1&p=view_empty_project_-_uj_class_1d which will be related to master ticket
- [ ] change headers of the messages to the user does not make much sense - instead of "Service provider message" we should have something like "Customer service message" - there is no provider at the other end. "Your question to the service provider" likewise
- [ ] "Order history" -we need to change this name 
- [ ] the display order of the conversation should be changed (it is a conversation so the latest news should be at the bottom of the window)
- [ ] possibility to comment should be available all the time.

Additional information after meeting (9.07.19):
- [ ]  add email notification for users, this notification should be sent after receiving response from JIRA, for now apply, title: test, email content: test

agpul:
Need to take into consideration:
1. Webhook -check project id
2.  How to deal with it in MP? Do we need history? Now we need only comments. For now we should have only chat but what with project fields?
3. Do we want to reuse chat code (from project_item) or write a new one?

agpul:
Estimation: 1 - 2 weeks

PR #960, opened 2019-06-27
Reimplementation of conversation in project and project_item

martaswiatkowska:
This simplifies the implementation of conversation connected with Project Item and Project.  
In this User can contact with Project support on project conversation view. This conversation is connected to master ticket, Project Item conversation is connected with subticket.

Fix: #842


kmarszalek:
converations -> conversation?

bwilk:
@martaswiatkowska conversation works one way. Customer service can write messages in JIRA and they are show in projects conversations. Contrary - messages written in projects are not shown in JIRA

bwilk:
I had a look at the view and have some comments - we should discuss them with the others @abacz @roksanaer @agpul. 

Zrzut ekranu 2019-07-1 o 14 49 16

1. Headers of the messages to the user does not make much sense - instead of "Service provider message" we should have something like "Customer service message" - there is no provider at the other end. "Your question to service proivder" likewise 
1. "Order history" - this isn't an order (this is a project), this isn't history (it's conversation). We need to change this name

bwilk:
you've added _dump.rdb_ file by accident

app/models/project.rb
```diff
@@ -40,4 +42,8 @@ def require_jira_issue?
   validates :project_website_url, url: true, presence: true, if: :project?
   validates :company_name, presence: true, if: :private_company?
   validates :company_website_url,  url: true, presence: true, if: :private_company?
+
+  def question_url
```
bwilk: is it ok to put route speciffic logic to model?

app/models/project_item.rb
```diff
@@ -43,6 +45,10 @@ class ProjectItem < ApplicationRecord
 
   delegate :user, to: :project
 
+  def question_url
```
bwilk: same here

app/controllers/api/webhooks/jiras_controller.rb
```diff
@@ -14,14 +14,16 @@ def authenticate_jira!
...
+      find_by(issue_id: params["issue_id"])
+
+    if element_updated
       case params["webhookEvent"]
       when "jira:issue_updated"
-        Jira::IssueUpdated.new(project_item, params["changelog"]).call
+        Jira::IssueUpdated.new(element_updated, params["changelog"]).call
       when "comment_created"
-        Jira::CommentCreated.new(project_item, params["comment"]).call
+        Jira::CommentCreated.new(element_updated, params["comment"]).call
```
bwilk: this is a sidenote - but probably important in scope of this PR. How do we handle JIRA comment delete? Maybe we should rise a ticket for this situation?

app/controllers/project_items_controller.rb
```diff
@@ -2,13 +2,37 @@
...
+      redirect_to project_item_path(@messageable)
+    else
+      redirect_to project_item_path(@messageable)
+    end
+  end
+
+  private
+
+    def load_messageable
+      @messageable = ProjectItem.find(params[:project_item_id])
```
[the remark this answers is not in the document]
bwilk: `@question = Question.new(messageable: @project_item)` here you are using different field `@project_item` for the same purpose

app/helpers/project_items_helper.rb
```diff
@@ -2,14 +2,14 @@
...
-      "Your question to service provider"
-    elsif previous
-      if answer?(previous, current)
-        "Service provider message"
+    if current.is_a?(Message)
+      if current.question?
+        "Your question to service provider"
       else
-        "Status changed from #{previous.status} to #{current.status}"
+        "Service provider message"
```
bwilk: please rename "Service provider message" to "Customer service message"

app/helpers/project_items_helper.rb
```diff
@@ -2,14 +2,14 @@
...
 module ProjectItemsHelper
   def status_change(previous, current)
-    if current.question?
-      "Your question to service provider"
-    elsif previous
-      if answer?(previous, current)
-        "Service provider message"
+    if current.is_a?(Message)
+      if current.question?
+        "Your question to service provider"
```
bwilk: please rename "Your question to service provider" to "Your  question to customer service" @roksanaer is informed and in the loop. Maybe later we'll represent JIRA roles here but for now there is no provider in this workflow, and this name will suit the situation much better

app/views/messages/_messages.html.haml
```diff
@@ -0,0 +1,10 @@
+%h2.mb-4 Order history
```
bwilk: Please move this header outside of this fragment to allow us to change labels in scope of the use either in project or project item. In project related conversation please name it "Conversation with customer service", in project item conversation for now leave it as is "Order history"

app/models/project.rb
```diff
@@ -21,7 +23,7 @@ class Project < ApplicationRecord
 
   belongs_to :user
   has_many :project_items, dependent: :destroy
-
+  has_many :messages, as: :messageable, dependent: :destroy
```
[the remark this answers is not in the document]
martaswiatkowska: https://guides.rubyonrails.org/association_basics.html#polymorphic-associations
Its deffinition of polymorphic association

app/controllers/api/webhooks/jiras_controller.rb
martaswiatkowska: We not handling this. Yes probably we should rise ticket for that.

agpul:
I have two comments because or maybe questions:
1. @martaswiatkowska  Now, when I (as a user) write several messages to the customer service I am on the top of the list of my messages but in fact, I would like to see my latest message first. Is it possible to change it?
How it looks now: image
How I would like to see this: 
image

2. @abacz I would like to ask you if is it possible to distinguish the message written by a customer from this written by support for eg, by using different colours or smth, maybe you have an idea

3. @martaswiatkowska is it possible to make this feature "live". For eg., I write something to support and now I am waiting for a response but I need to refresh the page to see it, but maybe it is not w key feature and maybe it is not necessary, @abacz please comment also this idea, because it may be stupid

abacz:
1. Agree
2. Ok. I will prepare something.
3. It's ok. But more important thing is that user should receive an email notification about answer (or any change from support) with link to this tab. Do we have it @martaswiatkowska ?

martaswiatkowska:
@abacz No, we don't have it because I didn't have it in my ticket.

agpul:
Fix typo:
- from costomer to customer

roksanaer:
can you please also add "ago" at the end of timestamp label
obraz

roksanaer:
another thing: is it possible to filter messages sent from JIRA that are labeldd are 'for administrators' 
obraz
at the moment if i write 'internal' message (that is the intention behind it)
obraz
it is still visible to the user
obraz

if this is too complex, we can leave it for the time being (will add another ticket for it for the next sprints)

app/controllers/api/webhooks/jiras_controller.rb
bwilk: @agpul unless @martaswiatkowska has already created this ticket - this is a known issue

app/models/project.rb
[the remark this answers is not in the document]
bwilk: Ok. Makes sense

app/views/messages/_message.html.haml
```diff
@@ -0,0 +1,11 @@
+%li.step-progress-item.current.ml-5
+  .time.x-small
+    = time_ago_in_words(current.created_at)
```
bwilk: \+ " ago"

app/views/messages/_messages.html.haml
```diff
@@ -0,0 +1,13 @@
+#scrollbar.progress-box.mt-4
+  %ul.step-progress.pl-5.pb-4
+    - if messages.count.positive?
+      - messages.each do |current|
+        = render "messages/message", current: current
+    - else
+      %p Conversation not started yet.
```
bwilk: @roksanaer don't forget to change this text as well

app/models/project_item.rb
```diff
@@ -27,7 +27,8 @@ class ProjectItem < ApplicationRecord
   belongs_to :project
   belongs_to :research_area, required: false
   has_one :service_opinion, dependent: :restrict_with_error
-  has_many :project_item_changes, dependent: :destroy
+  has_many :messages, as: :messageable, dependent: :destroy
+  has_many :statuses, as: :pipeline
```
bwilk: I understand that this is made polymorphic just in case we need same statuses somewhere else e.g. in projects yes?

spec/models/status_holder_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+shared_examples "status_holde" do
```
bwilk: typo "status_holde"

Issue #899, opened 2019-06-05
Allow copying projects


agpul:
In the old version (with affiliation) he/she could fill affiliation once but use it many times. Due to the removal of affiliation, we would like to give the user the possibility of copying the project (user will not be able to fill the same information many times).

Prepare UI for this issue.

abacz:
I suggest to add Duplicate button where after click user will see form "create new (or sth)" with fields already filled. @roksanaer 

https://invis.io/TJRBIGOZG8P#/355651212_MyServices_Mngmt_-_Detail_Empty_Project

agpul:
For me this idea is nice but what don't you think that we should have also the possibility to duplicate the project from this view: https://kcjh3g.axshare.com/#g=1&p=my-services-list

agpul:
1. We leave buttons only on the project details view,
2. Remember to copy projects in the database except for JIRA fields (issue id, issue status ), then project register.
3. This project should be empty after registration.
4. @agpul @roksanaer consider if we should for sure copy all fields for eg reason for access maybe shouldn't be copied?

agpul:
Estimation: 1-2 days

mkasztelnik:
Time: 2h

mkasztelnik:
Done in the scope of #968

PR #968, opened 2019-06-28
Create new project basing on existing one

mkasztelnik:
User is able to create project duplicate. All form elements are copied except the name.

project-duplicate

project-cloned

Fixing #899


spec/features/projects_spec.rb
```diff
@@ -51,7 +51,7 @@
 
     scenario "I can create project duplicate" do
       allow(project_create).to receive(:call)
-      project = create(:project)
+      project = create(:project, user: user)
```
bwilk: good catch :)

Issue #929, opened 2019-06-18
Delete repeating buttons on projects page


agpul:
**Describe the bug**
In PR Sprint when we add more than one service to the project we see "Add another service" button many times.

**To Reproduce**
1. Go to https://marketplace-14.docker-fid.grid.cyf-kr.edu.pl/projects
2. Create the project
3. Add a least 3 services to the project
4. See error: ![image.png](https://images.zenhubusercontent.com/5b8fbb317859ca34f6abbcb2/bf2395f4-36a1-435e-b360-7f86906885e5)

**Expected behavior**
We should have only one button "Add another service" at the bottom of the list

mkasztelnik:
Done in the scope of #970

PR #970, opened 2019-07-01
Project list view redesign

mkasztelnik:
The name of the branch is misleading. Fix for #929 was done as a result of implementing the new @abacz idea for projects list look and feel. This is not finished yet. Some CSS alignments are still needed (especially `card` component and HTML elements in this component is too specific for offer presentation and need to be refactored).

TODOs:
  * [x] @mkasztelnik initial redesign

This is how it looks like right now:
new-projects-list


mkasztelnik:
Since styling will be done globally by @jarekzet this can be merged as it is after fixing scss lint errors.

mkasztelnik:
Merget do #979

Issue #902, opened 2019-06-05
Move "Research Area" field to the project and change it to multiselect


goreck888:
**Is your feature request related to a problem? Please describe.**
Because of creating a hierarchy of tickets in jira there is a need to change Research area field to multiselect and move it to the project. It avoids possible conflicts by moving research area from project item to project, because project items can have multiple research areas as part of the project.

**Describe the solution you'd like**
- [ ] Change field to multiselect on front-end, 
- [ ] move it to project form, 
- [ ] change ResearchArea column to an array in the back-end and make migration including move it to the projects table.

agpul:
Estimation: 2 days

PR #978, opened 2019-07-02
[#902] move research area to project

goreck888:
Move research area to the project
Change research area to multiselect


goreck888:
#knownIssue There's a problem with views. In modal there's no choices loaded and the multiselect view is basic. Moreover in the details view under "Research areas" it shows [Object object] instead of Research areas names. That will be fixed in the views fix PR

goreck888:
Views added.

agpul:
known issues looks ok

PR #989, opened 2019-07-10
Service owner can edit service draft

### LINKED ISSUE #983: Add edit privilages for service owners
**Current situation**
SP managers can edit and publish all services, service owners have read only privilages.

**what is needed**
Allow service owners to edit service entries **belonging to them** which are in a **draft state**

mkasztelnik:
Service draft can be now edited by the owner. I'm waiting for decision if the owner should be able to edit service draft offers as well.

Fixes #983

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ## [Unreleased]
 
 ### Added
+- Service owner can edit service draft (@mkasztelnik)
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


mkasztelnik:
Offer in the draft state can be removed only when there is no service ordered using this offer. This should be good enough for the case you are describing.

agpul:
image

I ordered one offer and I can't edit both of them.

mkasztelnik:
There is a bug connected with checking if the offer can be removed. Now if service is ordered using **any** offer than you cannot remove any service offer. This need to be fixed.

Issue #982, opened 2019-07-04
Import of OpenAIR services to the MP


roksanaer:
**Current situation**
at the moment we have a possibility to import services from eIC. we can import all of them at once
**What we need**
We need to import OpenAIR services only (20 out of 190)

**What is missing**
Posiibility to depict services we want to import (in command line is enough. no gui needed). we can use e.g. ids from eIC. Ideally, we want to have a possibility to import services with a 'provider' filter.

**Additional requirements**
Apart from pulling the info from eIC to the MP, Service Portfolio Team asks very kindly, to generate some files (any format) containing **all information** for chosen services (we import only a subset of them)

PR #993, opened 2019-07-10
[#982] Selective service import from EIC

michal-szostak:
## What is in this PR?

* Add selective import from eic based on service ID
* Fix image import & add tests for it
* Add 'Other' research area to seeds, so that PRs can be tested
  more easily

## How to test it?

You need to log into the container on which server is running, and have access to `rake`

You can import specific services (defined by IDS) via following command:

```
bin/rake import:eic IDS=service_id_1,service_id_2 OUTPUT=path_to_json_output_file
```

* `IDS` is a comma separated list of ids from eIC, without any spaces
* `OUTPUT` is a path to the output file in which all imported services' original JSON entries will be written

Closes: #982


db/data.yml
```diff
@@ -239,6 +239,8 @@ area:
         name: &socialSciences "Social Sciences"
     Philosophy:
         name: &philosophy "Philosophy / History"
+    Other:
+        name: &other "Other"
```
mkasztelnik: This will not be added on production after the deploy, it needs to be added manually using backoffice (cc @agpul).

lib/import/eic.rb
```diff
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
 
+require "mini_magick"
+
 module Import
   class EIC
     def initialize(eic_base_url,
-                   dry_run = true, dont_create_providers = true, unirest = Unirest,
+                   dry_run = true,
```
mkasztelnik: Can we move these parameters to a dedicated class? Or at least use named parameters?

lib/import/eic.rb
```diff
@@ -47,17 +55,23 @@ def call
...
       not_modified = 0
       total_service_count = r.body["results"].length
+      output = []
 
       log "EIC - all services #{total_service_count}"
 
       r.body["results"].each do |service|
         eid = service["id"]
+
+        if !@ids.empty? && !@ids.include?(eid)
```
mkasztelnik: This is an imperative way of doing things, maybe functional style can be used instead?

```ruby
r.body["results"].select { |r| (@ids.include?(r["id"]) }.each....
```

mkasztelnik: This can go even further and instead of `each` `map` can be used to create new (not saved yet) `Services` which depending on the dry run mode or not will be saved to the DB.

spec/lib/import/eic_spec.rb
```diff
@@ -4,12 +4,37 @@
...
+  def make_and_stub_eic(ids = [], dry_run = false, dont_create_providers = false, filepath: nil, log: false)
+    options = {}
+
+    unless log
+      options[:logger] = ->(_msg) { }
+    end
+
+    eic = Import::EIC.new(test_url, dry_run, dont_create_providers, ids, filepath, unirest, options)
+
+    r1 = open(file_fixture("PhenoMeNal_logo.png"))
```
mkasztelnik: This can be extracted to a method and reused in other 2 places.

spec/lib/import/eic_spec.rb
```diff
@@ -43,7 +68,7 @@ def expect_responses(unirest, test_url, services_response = nil, providers_respo
       provider_response = double(code: 500, body: {})
       expect_responses(unirest, test_url, response, provider_response)
 
-      eic = Import::EIC.new(test_url, false, true, unirest, logger: nil_logger)
+      eic = make_and_stub_eic(log: false)
```
mkasztelnik: What is the difference between this variable and the one defined in `let` ?

spec/lib/import/eic_spec.rb
```diff
@@ -56,12 +81,12 @@ def expect_responses(unirest, test_url, services_response = nil, providers_respo
     end
 
     it "should not create if 'dont_create_providers' is set to true" do
-      eic = Import::EIC.new(test_url, false, true, unirest, logger: nil_logger)
+      eic = make_and_stub_eic([], false, true)
```
mkasztelnik: It is hard to understand what is going on here because of many magic parameters (e.g. what do false or true parameters mean).

lib/import/eic.rb
```diff
@@ -16,6 +22,8 @@ def initialize(eic_base_url,
           "trl-9" => "production"
       }
       @logger = logger
+      @ids = ids
```
mkasztelnik: `@ids = ids || []`

lib/import/eic.rb
michal-szostak: I'm not sure about moving them out, but named parameters sound like a good idea.

spec/lib/import/eic_spec.rb
michal-szostak: the fact, that in other places additional parameters might be supplied to refine returned object.
With let there is a need to predefine every possibility as a separate _variable_

spec/lib/import/eic_spec.rb
michal-szostak: Sorry about that, I kind of got used to some tool assist (look at the screen below (maybe you can configure your editor to do the same?)

image

But nevertheless - I can change parameters to named ones

db/data.yml
michal-szostak: I am well aware of it, the required research area has already been added on production, but in order to make testing on automatically deployed PRs easier I also added it to the seeds.

roksanaer:
list of OpenAIR services:
OpenAIR services ids:

"openaire.openaire_data_provider_dashboard"
„openaire.amnesia”
"openaire.greek_sustainable_development_solutions_network_sdsn_openaire_dashboard"
"openaire.zenodo"
"openaire.technical_support_towards_openaire_compliance"
"openaire.open_science_helpdesk"
"openaire.openaire_validator"
"openaire.neuroinformatics_openaire_dashboard"
"openaire.openaire_graph"
"openaire.openaire_inference_"
"openaire.open_science_training"
"openaire.fisheries_and_aquaculture_management_openaire_dashboard"
"openaire.openaire_broker"
"openaire.openaire_usage_statistics"
"openaire.european_marine_science_openaire_dashboard_"
"openaire.openaire_discovery_portal"
"openaire.openaire_funder_dashboard"
"openaire.openaire_scholexplorer"

spec/lib/import/eic_spec.rb
michal-szostak: Ah, you meant in this particular case - nothing, my bad I have overlooked the fact that I had readily available variable already

bwilk:
@roksanaer @agpul you can test now services has been imported in PR deployment instance

roksanaer:
i don't see links  in our backoffice for imported services. For example for OpenAIR FUnder Dashboard
https://marketplace-6.docker-fid.grid.cyf-kr.edu.pl/backoffice/services/openaire-funder-dashboard

i don't see links for HD, userManual etc

order | "https://monitor.openaire.eu/"
-- | --
helpdesk | "https://www.openaire.eu/support/helpdesk"
userManual | "https://www.openaire.eu/monitoring-guide"
trainingInformation | "https://www.openaire.eu/category/funders-2"
feedback | null
price | null
serviceLevelAgreement | "https://www.openaire.eu/service-level-agreement"

roksanaer:
> i don't see links in our backoffice for imported services. For example for OpenAIR FUnder Dashboard
> https://marketplace-6.docker-fid.grid.cyf-kr.edu.pl/backoffice/services/openaire-funder-dashboard
> 
> i don't see links for HD, userManual etc
> order 	"https://monitor.openaire.eu/"
> helpdesk 	"https://www.openaire.eu/support/helpdesk"
> userManual 	"https://www.openaire.eu/monitoring-guide"
> trainingInformation 	"https://www.openaire.eu/category/funders-2"
> feedback 	null
> price 	null
> serviceLevelAgreement 	"https://www.openaire.eu/service-level-agreement"
> 
> only service website and (empty) terms of use

OK, my bad, sorry. those are just not visible in the GUI... everything is fine checked for the majority of services. greenlight from me

roksanaer:
important!
For the import to work, 'Others' category needs to be added to research areas (e.g. each time docker is restarted)

PR #996, opened 2019-07-12
Project view have projects navigation

mkasztelnik:
In the scope of this PR we will rearrange all projects views.


roksanaer:
quick remark. i did early tests. after improved redirecton from ordering (to the project/project item) in this PR we lost a possibility to add another service (see screens)
obraz
obraz

app/controllers/projects_controller.rb
```diff
@@ -90,6 +93,10 @@ def destroy
 
   private
 
+    def projectless?
```
martaswiatkowska: Where do you use this?

mkasztelnik:
@agpul fixed, now after project create user is redirected to the new project instead of the first project.

app/controllers/projects_controller.rb
mkasztelnik: Good catch, I already removed it.

Issue #974, opened 2019-07-02
Cleaning the mess related to user "country"


agpul:
**Is your feature request related to a problem? Please describe.**
Now we have in Marketplace at least 3 places where we ask the user about different countries:
-> we ask about "Country of the customer" in project form
-> we ask about "Country of collaboration" in Customer topology section in the project
We've tried to understand the difference between these fields and clean this small mess.

**Describe the solution you'd like**
- [x] leave Country of collaboration in the current place as multiselect and change the name of this field depending of chosen option:
   - [x]  for Representing a research community it should be called: Which countries are involved in this community? Please select those you are aware of.
   - [x] for Representing a project it should be called:  Which countries are involved in these projects? Please select those you are aware of. 
   - [x] for Representing a private company remove this field
- [x] move Country of customer (select) under Customer topology, for options:
    - [x] Representing a private company it should be called: Where is it located? 
    - [x] Single user it should be called: In which country is your institution located? 
- [x] make sure that all changes are compatible with JIRA

goreck888:
What with details view? Should it be similar like in form or should I change it or leave as it is? (Customer country and collaboration country)

bwilk:
This is how the fields should be ordered in this view. 

When `Customer typology` is not selected  
![bez typology.png](https://images.zenhubusercontent.com/5b0fe7c84b5806bc2bd57e5c/803ece68-0061-4b69-9669-fae9b05cae38)

When `Customer typology` is selected  
![z typology.png](https://images.zenhubusercontent.com/5b0fe7c84b5806bc2bd57e5c/70982d91-48e5-43f7-b269-66c124f5ff3c)

bwilk:
Please organize information the same way in the `Project details` view

![Zrzut ekranu 2019-07-10 o 16.44.31.png](https://images.zenhubusercontent.com/5b0fe7c84b5806bc2bd57e5c/7d313674-aa0c-4970-9a0c-42d2a6bc64a7)

bwilk:
Also add `Country of customer` to the details view

PR #997, opened 2019-07-12
[#974] user country improvements

goreck888:


app/models/project.rb
```diff
@@ -19,25 +19,25 @@ class Project < ApplicationRecord
   enum customer_typology: CUSTOMER_TYPOLOGIES
   enum issue_status: ISSUE_STATUSES
 
-  NON_APPLICABLE = "N/A"
-  INTERNATIONAL = "I/N"
-  NON_EUROPEAN = "N/E"
-
-  allowed_countries = [NON_APPLICABLE, INTERNATIONAL, NON_EUROPEAN] +
-        ISO3166::Country.find_all_countries_by_region("Europe").sort.map { |c| c.alpha2 }
+  ALLOWED_COUNTRIES = [Country::NON_APPLICABLE, Country::INTERNATIONAL, Country::NON_EUROPEAN] +
```
mkasztelnik: Why this is not defined in the `Country` model?

app/helpers/countries_helper.rb
```diff
@@ -2,29 +2,15 @@
 
 # Used to show countries list in the front
 module CountriesHelper
-  NON_EUROPEAN = ["non-European", "N/E"]
-  INTERNATIONAL = ["International", "I/N"]
-  NON_APPLICABLE = ["non-applicable", "N/A"]
-
   def customer_countries
-    countries(NON_EUROPEAN)
+    [Country::NON_EUROPEAN] + Country.european_countries
```
mkasztelnik: It looks like this list is different than the list of allowed countries defined in project field validation. Can you explain the difference?

app/models/project.rb
```diff
@@ -60,6 +60,22 @@ def single_user_or_community?
...
+  def research_or_project?
+    research? || project?
+  end
+
+  def single_user_or_private_company?
+    single_user? || private_company?
+  end
+
+  def country_of_customer=(value)
+    self[:country_of_customer] = Country.for(value)
```
mkasztelnik: Does it work with all callbacks and so on? Take a look here for alternative solution https://stackoverflow.com/questions/10464793/what-is-the-right-way-to-override-a-setter-method-in-ruby-on-rails

app/models/project.rb
```diff
@@ -60,6 +60,22 @@ def single_user_or_community?
...
+  end
+
+  def country_of_collaboration=(value)
```
mkasztelnik: Maybe better name for table variable?

app/models/country.rb
```diff
@@ -0,0 +1,57 @@
...
+  class << self
+    def for(value)
+      return value if value.is_a?(ISO3166::Country)
+      ISO3166::Country.new(value)
+    end
+
+    def load(json)
+      return nil if json.blank?
+      if json.is_a? Array
+        json.map { |e| Country.for(JSON.parse(e)) }
```
mkasztelnik: Why JSON here? The simple string should be enough here, isn't it?

app/models/country.rb
```diff
@@ -0,0 +1,57 @@
...
+  ISO3166::Data.register(
+    alpha2: "N/A",
+    name: "non-applicable",
+    translations: {
+        "en" => "non-applicable"
+    })
+
+  NON_EUROPEAN = ISO3166::Country.new("N/E")
+  INTERNATIONAL = ISO3166::Country.new("I/N")
+  NON_APPLICABLE = ISO3166::Country.new("N/A")
```
mkasztelnik: Why nil value cannot be treated as `NON_APPLICABLE` value? Right now there is a problem that user can select e.g. "N/A" and another country in multiselect fields - which is a nonsense.

app/models/project.rb
goreck888: i can move it but we left it with @bwilk here.

app/helpers/countries_helper.rb
goreck888: Yes, allowed countries have one option more, it's "non-applicable" for projects which are created before countries fields were added.

app/models/country.rb
goreck888: N/A is a value for projects created earlier than countries fields. It's not visible on front.
Countries fields are mandatory now.
I agree in case country of collaboration is a multiselect we can throw out the "INTERNATIONAL" value.

app/models/country.rb
goreck888: I was sure i changed json to str. It's possible i didn't pushed my commits.

PR #1001, opened 2019-07-17
[#995] Disable draft offers in the offers list in offers controller

### LINKED ISSUE #995: Draft offer can be used to order a service
**Describe the bug**
I can order service using draft offer 

**To Reproduce**
1. As a service portfolio manager go to Amber service and add a new draft offer
1. Add Amber service to the project
1. Open `rails c` and inspect last created `ProjectItem` (`ProjectItem.last.offer`)

**Expected behavior**
Published offer is used to order this service but instead draft one is selected

goreck888:
Add reject condition to check if service offer is draft
Fixes #995


app/controllers/services/offers_controller.rb
```diff
@@ -37,7 +37,7 @@ def offer_params
     end
 
     def init_offer_selection!
-      @offers = @service.offers.reject { |o| o.catalog? }
+      @offers = @service.offers.reject { |o| o.catalog? || o.draft? }
       @project_item = ProjectItem.new(session[session_key])
```
goreck888: This bug was caused in this line of code, because offers loaded all offers without catalog, including drafts and offers counter doesn't.

app/models/offer.rb
```diff
@@ -59,6 +59,10 @@ def catalog?
     offer_type == "catalog"
   end
 
+  def draft?
```
mkasztelnik: You don't need to define this method, because it is autogenerated for each enum value (e.g. you have `draft?` method and `draft!` by default).

goreck888: Ok, i will remove it.

PR #1008, opened 2019-07-24
1002 gui changes

jarekzet:
added CSS fixes & mods


abacz:
@jarekzet almost done. Buttons in content box have mb-2 so are not align properly. Visually is more space under. Please change bs class for buttons from mb-2 to mt-1 mb-1.

PR #1010, opened 2019-07-26
Fix vulnerabilities alerts

martaswiatkowska:


babel.config.js
```diff
@@ -0,0 +1,72 @@
+module.exports = function(api) {
+  var validEnv = ['development', 'test', 'production']
```
bwilk: @mkasztelnik @martaswiatkowska were we considering using https://babeljs.io/docs/en/config-files#apicache. Seems like a standard prcedure while writing `babel.config.js` files https://babeljs.io/docs/en/configuration. Or maybe config compliation  overhead is too little to bother webpacker to cache anything..?

mkasztelnik: We can back to this in the future.

mkasztelnik: We need to configure it ASAP because right now webpack is compiling assets for each test.

Issue #897, opened 2019-06-05
Disable animations in capybara tests


michal-szostak:
In some cases (usually in functionalities using bootstrap) animations cause capybara tests to fail and require additional workarounds.

A good idea would be a global animation disabling which would make writing tests easier and less prone to "trial and error"

Preliminary search shows that including CSS disabling all animations and including it only in tests would be a good start. Additional JQUEY animation disable might also be necesary.

roksanaer:
@michal-szostak . this task is quite low at the moment with priority. should we change it?

goreck888:
This is time without and with animations in tests:

Without animations
real    2m16,459s
user    0m58,995s
sys     0m6,143s
With animations
real    2m20,539s
user    1m6,161s
sys     0m6,311s

PR #1028, opened 2019-08-08
[#897] Disable animations in capybara tests

goreck888:
Turn off all css animations in the test environment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Use project details partial on service order configuration step (@mkasztelnik)
 - Apply a new view and functionalities of the start page (@jarekzet, @goreck888)
 - Rename buttons for creating new platforms, providers and categories in the backoffice (@goreck888)
+- Disabled css animations in the test environment (@goreck888)
 
 ### Removed
 - Affiliation model removed (@mkasztelnik)
--- a/spec/features/service_spec.rb
+++ b/spec/features/service_spec.rb
@@ -228,8 +228,6 @@ RSpec.feature "Service filtering and sorting" do
     create(:service, title: "DDDD Something 3", rating: 3.9, platforms: [platform_2], categories: [category_1])
 
     Service.reindex
-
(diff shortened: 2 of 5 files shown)


spec/features/service_spec.rb
```diff
@@ -407,7 +398,7 @@
     click_on(id: "filter-submit")
 
     expect(page).to have_selector(".media", count: 3)
-    find(:css, "a[href=\"#collapse_target_groups\"][role=\"button\"] h6").click
+    # find(:css, "a[href=\"#collapse_target_groups\"][role=\"button\"] h6").click
```
goreck888: in fact this test passed earlier because the animation wasn't complete in the moment of collapse.
Now default behavior is all active filters are opened so this line is unnecessary.

martaswiatkowska: So if is unnecessary you can delete this

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Added
 
 ### Changed
+- Disabled css animations in the test environment
```
mkasztelnik: Please add your nickname in the changelog entry.

app/javascript/stylesheets/_base.scss
```diff
@@ -36,3 +36,23 @@ h1 {
   height: auto;
   margin-bottom: 15px;
 }
+
+.test {
```
mkasztelnik: Please move this rule to the separate file, e.g. `_test.scss`. 

A question to @michal-szostak can configure webpack in such a way that this file will be included to final css only in the test scope? If yes then these rules can be defined on the body level and injected only in the test scope.

CHANGELOG.md
goreck888: Ok

bwilk:
@goreck888 is this PR ready to be merged?

goreck888:
> @goreck888 is this PR ready to be merged?

On my side yes, all things are ready, I just need to take approve from someone :)

PR #1029, opened 2019-08-08
[#949] Add names uniqueness validation in models

### LINKED ISSUE #949: Add validation for Providers, Research Areas, Categories and Platforms
**The same situation is in all tabs (providers, Research area, categories and platforms) but I will describe it for providers tab as an example.**

Now, when I use Backoffice for defining the new provider I can specify one item many times. To reproduce:
1. Go to: https://marketplace.eosc-portal.eu/backoffice/providers
2. Click "New Provider" button
3. Create Provider whose name is for eg. ABC 
4. Then add "New Provider" one more time
5. Type the same name for new provider as in point 3. (ABC)
6. See here: https://marketplace.eosc-portal.eu/backoffice/providers that we have ABC provider many times.

I would like to have a verification and the info for the user which will inform that the given
- [x] provider 
- [x] research area
- [x]  category
- [x] platform
 exists in the system.

goreck888:
Add uniqueness validation in models:
- provider
- research area
- category
- platform
Fixes #949

### DIFF
--- a/spec/models/platform_spec.rb
+++ b/spec/models/platform_spec.rb
@@ -3,5 +3,10 @@
 require "rails_helper"
 
 RSpec.describe Platform, type: :model do
-  it { should validate_presence_of(:name) }
+  describe "validations" do
+    it { should validate_presence_of(:name) }
+
+    subject { create(:platform) }
+    it { should validate_uniqueness_of(:name) }
+  end
 end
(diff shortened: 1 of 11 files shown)


spec/models/platform_spec.rb
```diff
@@ -4,4 +4,11 @@
 
 RSpec.describe Platform, type: :model do
   it { should validate_presence_of(:name) }
+
+  it "should have an unique name" do
+    Platform.create!(name: "abc")
```
martaswiatkowska: I think you can use "should validate_uniqueness_of(:name)" here

goreck888: Unfortunately it doesn't work. You can check build failures.

goreck888: Now it's ok

PR #1030, opened 2019-08-12
Showing only published services at the homepage

bwilk:
Quick fix for not public services that are currently shown (eIC services are not visible in catalogue but homepage shows them even though they are Drafts)

Zrzut ekranu 2019-08-12 o 18 34 55


bwilk:
> I do not see any tests...

Here you are 😄

Issue #951, opened 2019-06-27
Add possibility to sort services in Backoffice by their Status


agpul:
**Is your feature request related to a problem? Please describe.**
 Often I need to update or make changes in "draft" services but I don't remember their names. Now I have to find such services manually and it is very uncomfortable because of the number of services.

**Describe the solution you'd like**
As a Service Portfolio Manager, I would like to have the possibility to sort services by their status. Possible state:
- draft
- published

**Additional context**
You can check it here: https://marketplace.eosc-portal.eu/backoffice/services .  Draft services for eg. AppDB
This change should be applied only in Backoffice.

PR #1034, opened 2019-08-13
[#951] Sort services in the backoffice by their status

goreck888:
Add new service sort order in the backoffice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Added
 - Add new field helpdesk_email in the service model (@goreck888)
+- Services sort by their status in the backoffice (@goreck888)
 
 ### Changed
 - Move project create logic from controller to service (@mkasztelnik)
--- a/app/views/services/_order.html.haml
+++ b/app/views/services/_order.html.haml
@@ -5,8 +5,4 @@
     .col-sm-9.pl-0
       %select#sort.query-sort.form-control.form-control-sm{ "data-target": "order.select",
                                                             "data-submit-on-change": "data-submit-on-change" }
-        = options_for_select([["name A-Z", "title"],
-                              ["name Z-A", "-title"],
-                              ["rate 1-5", "rating"],
-                              ["rate 5-1", "-rating"],
-                              ["Best match", "_score"]], params[:sort])
+        = options_for_select(sort_options, params[:sort])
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ## [Unreleased]
 
 ### Added
-
+- service sort by their status in the backoffice
```
mkasztelnik: Add your nickname to the changelog

app/views/services/_order.html.haml
```diff
@@ -5,8 +5,17 @@
     .col-sm-9.pl-0
       %select#sort.query-sort.form-control.form-control-sm{ "data-target": "order.select",
                                                             "data-submit-on-change": "data-submit-on-change" }
-        = options_for_select([["name A-Z", "title"],
-                              ["name Z-A", "-title"],
-                              ["rate 1-5", "rating"],
-                              ["rate 5-1", "-rating"],
-                              ["Best match", "_score"]], params[:sort])
+        - if current_page? backoffice_services_path
```
mkasztelnik: Please move it to the controller. Set it as variable and use it in the view. In this way, you can avoid if/else condition.

spec/controllers/concerns/sortable_spec.rb
```diff
@@ -20,13 +20,21 @@ def initialize
...
+    it "should return ascending ordering by rating" do
       controller.params = { sort: "rating" }
       expect(controller.send(:ordering)).to eq("rating" => :asc)
     end
-    it "should return descending ordering" do
+    it "should return descending ordering by rating" do
       controller.params = { sort: "-rating" }
       expect(controller.send(:ordering)).to eq("rating" => :desc)
     end
+    it "should return ascending ordering by status" do
```
mkasztelnik: I know that you follow the other specs convention but this is wrong because the internal implementation is tested here. For now, leave it as it is, but this is another paint point which should be refactored (/cc @michal-szostak).

CHANGELOG.md
goreck888: Done

spec/controllers/concerns/sortable_spec.rb
goreck888: Ok.

app/javascript/controllers/order_controller.js
```diff
@@ -3,9 +3,38 @@ import { Controller } from "stimulus"
 export default class extends Controller {
   static targets = ["select"]
 
-  connect(){}
-
   initialize(){
+    this.options = [
```
mkasztelnik: By controller I mean `app/controllers/services_controller.rb` and `app/controllers/backoffice/services_controller.rb`. JS is not needed here.

goreck888: Now it should be ok :)

PR #1036, opened 2019-08-16
Project item id is project scoped

### LINKED ISSUE #1035: Project item id is project scoped
The situation right now is that we have two root elements:
  * `/projects/:iid` (`iid` - is user scoped - `iid` = `internal id`)
  * `/project_items/:id` (`id` is global)

We can improve it by changing the second URL into `/projects/project_iid/services/iid`. With adding `iid` to existing rows there is a problem with generating this value during the migration. We can try to use subquery with `select MAX`.

mkasztelnik:
Move `project_items/:id` and all subpaths to `projects/:project_id/services/:iid`. `iid` is generated when the model is saved. In the migration also `iid`s are generated for existing project items.

Fixes #1035


app/controllers/projects/services/conversations_controller.rb
```diff
@@ -18,23 +18,31 @@ def create
...
   private
 
-    def load_project
-      @project = @project_item.project
+    def load_projects
       @projects = policy_scope(Project).order(:name)
     end
 
+    def load_messages
+      @messages = (@project_item.messages + @project_item.statuses).
```
martaswiatkowska: Previously we don't showed project_item statuses

mkasztelnik: Yeap this was a bug fixed here => when a new message cannot be created existing messages were not shown.

Issue #955, opened 2019-06-27
Change names of buttons


agpul:
**Is your feature request related to a problem? Please describe.**
Some Service Portfolio Managers have problems with using the Backoffice because of bad button naming.

**Describe the solution you'd like**
Necessary changes:
- [x] https://marketplace.eosc-portal.eu/backoffice/categories FROM New Category TO Add new category
- [x]  https://marketplace.eosc-portal.eu/backoffice/providers FROM New Provider TO Add new provider
- [x] https://marketplace.eosc-portal.eu/backoffice/platforms FROM New Platform TO Add new platform

PR #1045, opened 2019-08-19
[#955] Change names of buttons

goreck888:
Rename button labels of creating new:
- categories
- platforms
- providers
Add button labels to the translation file

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -95,6 +95,24 @@ en:
       normal: Order
       open_access: Add to a project
       catalog: Go to the service
+  backoffice:
+    category:
+      new: Add new Category
+    platform:
+      new: Add new Platform
+    provider:
+      new: Add new Provider
+    research_area:
+      new: Add new Research Area
+    service:
+      new: Create new Service
+      edit: Edit
+      offer:
+        new: Add new Offer
+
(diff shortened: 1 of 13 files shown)


config/locales/en.yml
```diff
@@ -86,6 +86,17 @@ en:
       normal: Order
       open_access: Add to a project
       catalog: Go to the service
+  backoffice:
+    buttons:
+      new_category: Add new Category
+      new_platform: Add new Platform
+      new_provider: Add new Provider
+      new_research_area: Add Research Area
```
martaswiatkowska: maybe "Add new Research Area"

goreck888: i left it as it was but of course i can change that to Add new Research Area.
@roksanaer  we need your opinion :)

goreck888:
> Right now the implementation is button centric. I have mixed filings if this is good or bad. Most frameworks (e.g. `simple_form`) is using the resource-centric approach. For example:
> 
> ```yaml
> research_area:
>   new: New research area
> project:
>   new: New Project
>   create: Create project
>   update: Update project
> ```

I changed it. I don't if it's good to group the translations as backoffice but I suppose it's ok.

PR #1053, opened 2019-08-29
Remove jQuery fade out

goreck888:
Remove jQuery function for hide scrolling arrow on mobile devices
Move it to the home_page_controller.js


goreck888:
This PR must be tested with mobile mode in the web browser.

Issue #984, opened 2019-07-04
filter via service source


roksanaer:
**current situation**
We don't have functionality allowing user in the backoffice to filter services using the source of the service

**what is needed**
filter (for the back office only!) based on the source of the services

goreck888:
What do you mean, "source"? MP or eIC?

PR #1055, opened 2019-08-29
[#984] Filter via service source

bwilk:
Add a new filter showing services from external sources

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -21,7 +21,8 @@ class Service < ApplicationRecord
       providers: providers.map(&:id),
       platforms: platforms.map(&:id),
       target_groups: target_groups.map(&:id),
-      tags: tag_list
+      tags: tag_list,
+      source: upstream&.source_type
     }
   end
 
--- a/spec/controllers/concerns/searchable_spec.rb
+++ b/spec/controllers/concerns/searchable_spec.rb
@@ -232,5 +232,25 @@ RSpec.describe ApplicationController, type: :controller do
         basic_test
       end
     end
+
+    context Filter::UpstreamSource do
+      let!(:service1) { create(:service) }
+      let!(:service2) { create(:service) }
+      let!(:service4) { create(:service) }
+      let!(:service_source1) { create(:service_source, source_type: "eic", service_id: service1.id) }
+      let!(:service_source2) { create(:service_source, source_type: "eic", service_id: service2.id) }
+      let!(:filter_class) { Filter::UpstreamSource }
+      let!(:values) { "eic" }
+      let!(:param_name) { :source }
+      let!(:field_name) { :source }
+      it "checks if upstream source filter works" do
+        service1.upstream = service_source1
+        service2.upstream = service_source2
+        service1.save!
+        service2.save!
+        Service.reindex
+        basic_test
+      end
+    end
   end
 end
--- a/spec/models/filter/upstream_source_spec.rb
+++ b/spec/models/filter/upstream_source_spec.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+require "rails_helper"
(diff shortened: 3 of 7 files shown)


spec/models/filter/upstream_source_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+RSpec.describe Filter::UpstreamSource do
+  context "#options" do
+    let!(:service_mp) { create(:service) }
+    let!(:service_eic1) { create(:service) }
+    let!(:service_eic2) { create(:service) }
+    let!(:service_source1) { create(:service_source, service_id: service_eic1.id) }
+    let!(:service_source2) { create(:service_source, service_id: service_eic2.id) }
```
mkasztelnik: Currently, these services are not used in defined tests.

spec/models/filter/upstream_source_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    it "returns all sources specified in filters" do
```
mkasztelnik: Title of this spec is misleading. I'm not sure if I'm fan of such tests, which checks if we are returning exactly the same as defined in class.

spec/models/filter/upstream_source_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      filter = described_class.new
+      expect(filter.options).to contain_exactly({ name: "Any", id: "" },
+                                                { name: "Internal", id: "mp" },
+                                                { name: "eInfraCentral", id: "eic" })
+    end
+  end
```
mkasztelnik: Can you add a test which actually does the filtering?

spec/models/filter/upstream_source_spec.rb
goreck888: These tests are removed and proper test is written in the searchable_spec.rb

spec/models/filter/upstream_source_spec.rb
goreck888: It is in the searchable_spec.rb

spec/controllers/concerns/searchable_spec.rb
```diff
@@ -232,5 +232,25 @@ def basic_test
...
+      let!(:service2) { create(:service) }
+      let!(:service4) { create(:service) }
+      let!(:service_source1) { create(:service_source, source_type: "eic", service_id: service1.id) }
+      let!(:service_source2) { create(:service_source, source_type: "eic", service_id: service2.id) }
+      let!(:filter_class) { Filter::UpstreamSource }
+      let!(:values) { "eic" }
+      let!(:param_name) { :source }
+      let!(:field_name) { :source }
+      it "checks if upstream source filter works" do
+        service1.upstream = service_source1
```
bwilk: why don't you do this in create method call?

app/models/service.rb
```diff
@@ -5,7 +5,7 @@ class Service < ApplicationRecord
   # ELASTICSEARCH
   # scope :search_import working with should_indexe?
   # and define which services are indexed in elasticsearch
-  searchkick word_middle: [:title, :tagline, :description], highlight: [:title, :tagline]
+  searchkick word_middle: [:title, :tagline, :description, :source], highlight: [:title, :tagline]
```
bwilk: is this intended that we search in sources in the search bar?

spec/controllers/concerns/searchable_spec.rb
goreck888: Because if I try to create service_source without service it causes an error "service must exist".
This is only method I know to do this test work.

bwilk:
@goreck888 it was me that created this PR so you need to approve yourself :p. Review verdict is "approval" 👍

PR #1060, opened 2019-08-30
Homepage text changes

bwilk:


roksanaer:
will it be better with a smaller gap between the caption and text?
obraz

app/views/home/_products.html.haml
```diff
@@ -9,16 +9,16 @@
       .home-image
     .col-12.col-lg-7.left-column
       %p.title
-        Welcome to Marketplace
+        Welcome to the Marketplace
```
roksanaer: is it possible that this verse does not break in two lines?
obraz

bwilk: Done

PR #1064, opened 2019-09-02
[#1063] Removed uniqueness constraint on user email

### LINKED ISSUE #1063: Cannot log in with another account using same email.
**Describe the bug**
When a user has previously logged in using one IdP they are refused to create account using another IdP when the same email is used again.   

**To Reproduce**
1. Log in with IdP
2. Log in with another IdP from account using the same email 
3. See error

**Expected behavior**
Creation of another user account 

**Screenshots**
Zrzut ekranu 2019-09-2 o 15 46 39

**Additional context**
Email uniqueness validation should be removed

bwilk:
Fixes #1063

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Removed uniqueness constraint on user email (@bwilk)
 
 ### Security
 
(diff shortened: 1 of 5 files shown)


bwilk:
Validatable was removed as it is described here
http://jessewolgamott.com/blog/2011/12/08/the-one-where-devise-validations-are-customized/

PR #1076, opened 2019-09-09
[#1074] Display root research areas on the main page

### LINKED ISSUE #1074: Change categories to root in the backoffice
There's a need to change list of categories in the front to the root categories.

Now it's not checked if category is a root or a leaf.

goreck888:
Fixes #1074


db/data.yml
```diff
@@ -197,50 +197,86 @@ providers:
 area:
 
     Interdisciplinary:
```
goreck888: These changes are needed to seed proper root research areas in the PR instance, as it is in the production instance.

PR #1084, opened 2019-09-12
Include show arrow, when scroll to the top

goreck888:


jarekzet:
On this screen resolution .home-anchor has "display:none" style. Try to refresh CSS?

PR #1116, opened 2019-09-19
Publish service as unverified service

### LINKED ISSUE #1103: Add  a new status in the system to be able to flag the eIC services
We need to support the following use case:
We need to import simultaneously a lot of eIC services. After the import, those services will have 'draft' status (as it is now). As we need to publish them before October, we need to revise them internally in the team ( @agpul @MinLidia ) as Owen will not be able to do that. Then, make them visible in the catalogue for the users (publish them but also flag them that they were not verified by the onboarding team). The onboarding team needs to know which services were revised by us and 'published as unverified service'. After they can revise them and 'publish' (this will not change their visibility, only mark those services as revised by them).

Services marked as 'unverified' should be filterable on the back office for the sake of the SPM and have a way to mark them as verified.

As a result of this task we would like to:
- [x] add "unverified" status to the service model
- [x] add a new transition between "draft" and "unverified": "publish unverified service"
- [x] make the transition "publish unverified service" visible as a button in the back office
- [x] add an (already) existing transition "publish" between "unverified" and "published"
- [x] add an (already) existing transition "unpublish" between "unverified" and "draft" (*not sure if this transition is called unpublished or else)
- [x] change show? policy so it includes "unverified" status (or add a new policy notShow? for "draft")

**Additional context**
Relevant info is marked yellow:

mkasztelnik:
The new state was added to service: `unverified`. It means that service is published (it is visible to end-user) but is not confirmed by onboarding team.

Fixes #1103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ## [Unreleased]
 
 ### Added
+- Service unverified but published state (@mkasztelnik)
 
 ### Changed
 
--- a/app/controllers/backoffice/services/publishes_controller.rb
+++ b/app/controllers/backoffice/services/publishes_controller.rb
@@ -4,7 +4,7 @@ class Backoffice::Services::PublishesController < Backoffice::ApplicationControl
   before_action :find_and_authorize
 
   def create
-    Service::Publish.new(@service).call
+    Service::Publish.new(@service, verified: verified?).call
     redirect_to [:backoffice, @service]
   end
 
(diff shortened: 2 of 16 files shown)


roksanaer:
@agpul i'm not sure i get ur comment from above. the screenshot u showed implies (if i understand correctly) that it is possible to change the status from published to unverified which contradicts the specification:
obraz
is this the issue? if yes, it needs to be fixed.


@mkasztelnik i thought about your sugestion about the name of the state.
it makes sense assuming we will add certification soon, which is not certain at all. For thoday, the unveryfied state is (unfortunatelly) a workaround to cope with the situation we are facing (not able to proceed all services from eIC before publication, this is directly connected with the mess in eosc onboarding). I would stick to the 3 statuses: draft, unverified, published. As it reflects the current situation best. The unverified status will be used only once probably. after this one time action, all services will be either draft or published (and this names correctly carry the semantics /procedure behind)

mkasztelnik:
@agpul a dedicated filter to show service in concrete state is added (last one in the list). I believe this is better than sort by status, what is your opinion? Do we still need an additional sorting method?

The problem with changing published service into published as unverified fixed, please take another try.

PR #1124, opened 2019-09-25
Simple version of reload

### LINKED ISSUE #1086: Add auto-reload in the searching area
Right now to filter the information in the searching area we need to choose the criteria and click on the button Filter.

As a result of this issue, we would like to have auto-refresh (reload) after each filter action in the navigation panel.

**Additional context**
Button which is the topic of this task:

martaswiatkowska:
Fix #1086


agpul:
For me, it works quite fast but if it can work faster and the development can be done in a reasonable amount of time you can try to make it faster. However, I can't tell that it works slow.

abacz:
Works fine for me. I know that is a simple version but that we need to remember to add loading animation on list. Now or in improve version.

Issue #1087, opened 2019-09-17
Expand all filters by default 


agpul:
**Is your feature request related to a problem? Please describe.**
For now, the default view in the filtering panel assumes that all the filters are collapsed by default but we agreed that this idea is unfriendly for the users. 

**Describe the solution you'd like**
As a result of this task, we would like to have all filters expanded by default. 
- [ ] for non-hierarchical and hierarchical (providers, dedicated for, related infrastructures, rating, provider location, research areas) show 5 items with "Show more" phase

mkasztelnik:
Done in the scope of #1127

PR #1127, opened 2019-09-30
Store collapsed filter state

mkasztelnik:
The state is stored inside the user cookie

References #1087.


mkasztelnik:
@martaswiatkowska, @kmarszalek, @agpul this is ready to be tested and reviewed. I need to discover why one of the tests fail in travis but works locally (most probably this is because JS did not finish before assertion is made).

PR #1117, opened 2019-09-19
After project edit jira issue is updated

martaswiatkowska:
Fix #947

PR #1130, opened 2019-10-02
Filter unnecessary fields

martaswiatkowska:
It's filter unnecessary customer_typology fields saved during editing Projects to jira and db. 

To merge after #1117 !


app/javascript/controllers/project_controller.js
```diff
@@ -63,4 +63,12 @@ export default class extends Controller {
       el.classList.remove("hidden-fields");
     });
   }
+
+  filterFileds(event) {
+    this.inputTargets.forEach((el) => {
+      if(el.classList.contains("hidden-fields")){
+        el.children[1].value = null;
```
michal-szostak: This is too fragile in my opinion, using selector instead of `.children[1]` should be a better solution

michal-szostak:
@mkasztelnik - it can be done, but It would require more refactoring (this PR is a quick fix, not a long term solution, and should be treated as such)

mkasztelnik:
@michal-szostak you did not convince me :smile: I understand that such shortcuts we can do to fix fast some issues in the bugfix release but here we have/had time to do it correctly.

mkasztelnik:
After discussion with @martaswiatkowska I agree that it can be done in the front. The follow up of this issue is to move customer topology into a separate db model. A new issue for this will be created by @martaswiatkowska and mentioned in this PR.

PR #1137, opened 2019-10-08
Fix services iid for project

martaswiatkowska:
Fix generating iid for project_items in project

PR #1143, opened 2019-10-08
Fix services iid for project

michal-szostak:
This is a copy of https://github.com/cyfronet-fid/marketplace/pull/1137
It will be merged into `2.1.x` branch to facilitate hotfix


michal-szostak:
The parameters of this PR has mistake in it, thus I'm closing it

PR #1144, opened 2019-10-08
[HOTFIX to release] Fix services iid for project

michal-szostak:
This is a mirror of https://github.com/cyfronet-fid/marketplace/pull/1137
It was created so it can be merged into `2.1.x` branch and released as `2.1.5` version

### DIFF
--- a/db/migrate/20191003120324_fix_iid_incrementation.rb
+++ b/db/migrate/20191003120324_fix_iid_incrementation.rb
@@ -1,0 +1,15 @@
+class FixIidIncrementation < ActiveRecord::Migration[5.2]
+  def up
+    execute("UPDATE project_items item SET iid = 0")
+    execute("SELECT id FROM project_items").
+      each{ |pi| execute(
+        <<~SQL
+          UPDATE project_items item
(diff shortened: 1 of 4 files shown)


db/migrate/20191003120324_fix_iid_incrementation.rb
```diff
@@ -0,0 +1,15 @@
+class FixIidIncrementation < ActiveRecord::Migration[5.2]
+  def up
+    execute("UPDATE project_items item SET iid = 0")
+    ProjectItem.pluck(:id).
```
mkasztelnik: This is ok for a quick fix but we have to avoid active-model specific elements in migration. The main reason is that we can e.g rename `ProjectItem` to something else. To fix this issue you should execute raw SQL query, which will fetch `ids` from `project_items` table. Something like:

```ruby
execute("SELECT id FROM project_items").each { |pi| ... pi["id"] }
```

michal-szostak: True, I was only cherry picking other PR, and didn't analyze the code too much. But it's valid concern.

Issue #1123, opened 2019-09-25
Disable fields in edit form if upstream is different than MP (for now eIC)


agpul:
**Describe the solution you'd like**
 If service has upstream set to EIC the fields which are updated from the EIC should be disabled in the edit form (and possibly have some help tip), this way user will be clear that those fields are managed by another entity, and he/she should contact EIC to have them changed.

PR #1147, opened 2019-10-09
[#1123] Disable fields in service's edit with eid

michal-szostak:
## What is in this PR?

When upstream in service is set to EIC some fields (which are updated from EIC should be disabled)

image

image

The list of fields which should be disabled:

* `logo`
* `title`
* `description`
* `tagline`
* `connected_url`
* `places`
* `languages`
* `dedicated_for`
* `terms_of_use_url`
* `access_policies_url`
* `sla_url`
* `webpage_url`
* `manual_url`
* `helpdesk_url`
* `tutorial_url`
* `phase`
* `service_type`
* `providers`

Closes: #1123

### DIFF
--- a/app/controllers/backoffice/services_controller.rb
+++ b/app/controllers/backoffice/services_controller.rb
@@ -9,6 +9,7 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
   before_action :find_and_authorize, only: [:show, :edit, :update, :destroy]
   before_action :sort_options
   prepend_before_action :index_authorize, only: :index
+  helper_method :cant_edit
 
   def index
     if params["service_id"].present?
@@ -50,6 +51,10 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
     Service::Destroy.new(@service).call
     redirect_to backoffice_services_path,
                 notice: "Service destroyed"
+  end
+
+  def cant_edit(attribute)
+    !policy([:backoffice, @service]).permitted_attributes.include?(attribute)
   end
 
   private
--- a/app/policies/backoffice/service_policy.rb
+++ b/app/policies/backoffice/service_policy.rb
@@ -1,5 +1,28 @@
 # frozen_string_literal: true
 
+
+EIC_SOURCE_FIELDS = [
+  :logo,
+  :title,
+  :description,
+  :tagline,
+  :connected_url,
+  :places,
+  :languages,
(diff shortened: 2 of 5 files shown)


app/controllers/backoffice/services_controller.rb
```diff
@@ -9,6 +9,7 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
   before_action :find_and_authorize, only: [:show, :edit, :update, :destroy]
   before_action :sort_options
   prepend_before_action :index_authorize, only: :index
+  helper_method :cant_edit
```
mkasztelnik: Maybe `editable?`

app/policies/backoffice/service_policy.rb
```diff
@@ -72,6 +99,12 @@ def permitted_attributes
       [owner_ids: []], :status, :upstream_id,
       sources_attributes: [:id, :source_type, :eid, :_destroy]
     ]
+
+    if !@record.is_a?(Service) || @record.upstream.nil?
```
mkasztelnik: `@record&.upstream.nil?` should be enough :)

spec/features/backoffice/services_spec.rb
```diff
@@ -381,6 +381,80 @@
...
     end
+
+    scenario "if upstream is set to MP (nil) all fields should be enabled" do
+      service = create(:service, title: "my service", upstream: nil)
+      create(:service_source, service: service, source_type: :eic)
+
+      visit backoffice_service_path(service)
+      click_on "Edit"
+
+      expect(page).to have_field "Logo", disabled: false
```
mkasztelnik: You can extract names into table and do the iteration. Something like:

```ruby
names.each { |n| expect(page).to have_field "Logo", disabled: false }
```

The advantage of such extraction is that you can reuse names in second test.

michal-szostak: Not exactly, because disabled states are different depending on the test, it's a little bit verbose - I agree, but I think that in the long run it'll be more readable then having array of fields and then splice it with another array of expected disabled states

app/policies/backoffice/service_policy.rb
michal-szostak: Not exactly - in some cases `@record` is not a class instance, if I use the option you've provided there is `NoMethodError: undefined method `upstream' ` thrown there.

Issue #798, opened 2019-04-04
As a user I want to be redirected to an initiated order after being asked to log in 


bwilk:
**Describe the bug**
When a user want's to order a service they are requested to log in. Then they are correctly redirected to AAI, but after successful logging in they are not redirected back to the product they were trying to order nor to the initiated ordering process. 

**To Reproduce**
1. Enter MP not logged in 
1. Go to a desired service page 
2. Click on Order button at the Service Offer (you will be redirected to AAI)
3. Log in 
4. See the Home page instead of the product/started order page 

**Expected behavior**
A user should be redirected back to the ordering process wizard 

**Screenshots**
None 

**Additional context**
None

PR #1203, opened 2019-10-23
Offer selection available for anonymous user

mkasztelnik:
Offers selection is made available for anonymous users. There is one problem: while the user is clicking on a specific offer then after checkin login session is cleared and the selected offer is not available. This only happened in the development environment. The solution to this issue was to change the development `cache_store` to `memory_store`. 

There is another change connected with redirecting user after a successful login. Before ` request.env["omniauth.origin"]` was used and it was done in the PR #941 as a solution for #798. After going back to the default implementation (method `after_sign_in_path_for` was removed from `app/controllers/users/omniauth_callbacks_controller.rb`) I was not able to reproduce the issue described in #798 but is require to double-check (@bwilk can you do that?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Changed
 - Internal server error message (@goreck888)
+- Anonymous user can enter service offer selection section (@mkasztelnik)
 
 ### Deprecated
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   include Turbolinks::Redirection
   include Sentryable
   include Pundit
+  include Devise::StoreLocation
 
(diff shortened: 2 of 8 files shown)


mkasztelnik:
@agpul in the universe this PR lives information step does not exist yet.

bwilk:
Redirection back to the MP after login doesn't work as expected.  

#### To replicate:

1. Go to service and click Login:

Zrzut ekranu 2019-10-28 o 15 22 24

2. You are redirected to the home page 

Zrzut ekranu 2019-10-28 o 15 22 57


#### Expected behavior:
Redirection to the service page

mkasztelnik:
@bwilk problem fixed using the solution described here https://stackoverflow.com/questions/15944159/devise-redirect-back-to-the-original-location-after-sign-in-or-sign-up/21632889. Please do retests.

mkasztelnik:
This feature was no changed in this pr. Redirecting do root_path was from
the beginning.

wt., 29 paź 2019, 08:03 użytkownik agpul <notifications@github.com> napisał:

PR #1205, opened 2019-10-24
Add catalog to project

martaswiatkowska:
Catalog service can be added to Project similarly as open_access.


app/controllers/services/summaries_controller.rb
```diff
@@ -6,7 +6,7 @@ class Services::SummariesController < Services::ApplicationController
 
   def show
     @project_item = project_item_template
-    render "show_#{@service.service_type}"
+    @service.normal? ? (render "show_normal") : (render "show_open_access")
```
mkasztelnik: Most probably there is a better way for `show_open_access` which will be ok for both `open` and `external` services. @roksanaer, @bwilk any ideas?

app/models/project_item.rb
```diff
@@ -48,8 +48,8 @@ def service
     offer.service unless offer.nil?
   end
 
-  def open_access?
-    @is_open_access ||= service&.open_access?
+  def normal?
+    @is_normal ||= service&.normal?
```
mkasztelnik: You can simplify this a bit with delegate method:

```ruby
delegate :normal?, to: :service, allow_nil: true
```

app/models/project_item.rb
```diff
@@ -91,7 +91,7 @@ def to_s
   end
 
   def one_per_project?
-    if open_access?
+    unless  normal?
```
mkasztelnik: I'm not sure if this is valid anymore. As far as I understand now service can have many offers, what is more offer can have parameters. @roksanaer, @agpul can you describe how it should look like? Do we need any logic on the offer level saying that service with this offer can be added to the project only once?

app/models/project_item.rb
```diff
@@ -91,7 +91,7 @@ def to_s
   end
 
   def one_per_project?
```
mkasztelnik: This is not connected with this PR but generally, methods with `?` at the end should always return `true`/`false` and not change any object state. In this case, it should be renamed into something like `ensure_one_per_project`.

app/models/service.rb
```diff
@@ -99,7 +99,7 @@ def search_data
   validates :title, presence: true
   validates :description, presence: true
   validates :tagline, presence: true
-  validates :connected_url, presence: true, url: true, if: :open_access?
+  validates :connected_url, presence: true, url: true, if: :not_to_order?
```
mkasztelnik: `unless normal?`

app/controllers/services/summaries_controller.rb
martaswiatkowska: Yes but this will be done in other PR where I'm planning refactor 4th step. Then will be one view for every type of service

app/models/project_item.rb
roksanaer: to be honest, we assumed that this particular logic should rely not on the service offer access type nut existance of parameters. If the offer has no parameters at all, it's impossible to create 'an instance' of it and, in consequence, distinguish project items. If the offer has params that allow to create distinguishable instances of the project_item (based on an offer with params), we should allow adding one offer multiple times to the project.
in summary: I wouldn't sustain a rule based on service_offer_access_type but  add one relying on existance of offer params (if possible)
this was discussed but apparently not written down so @mkasztelnik thanks for spotting that!

martaswiatkowska: Ok but for me its topic for another PR. I propose to leave it and add additional bullet to main ticket

spec/features/order_spec.rb
```diff
@@ -210,6 +239,35 @@
...
+
+
+      # Summary page
+      expect(page).to have_current_path(service_summary_path(catalog_service))
+      expect(page).to have_selector(:link_or_button,
+                                    "Add to a project", exact: true)
+
+      expect do
+        click_on "Add to a project", match: :first
+      end.to change { ProjectItem.count }.by(1)
```
michal-szostak: Ok, this is pedantic so you don't have to take it into account at all, but one line expects requiring blocks are more readable in form of `expect { click_on "Add to a project", match: :first }.to change ...`

PR #1222, opened 2019-11-04
[#1201] Service summary unification

### LINKED ISSUE #1201: Reorganisation service summary view
In this task we need to change logic of service summary.

Now there we have different summary views depending on service type.
The main goal of this ticket is summary step unification and disable dependency of service type.

The main reason is, we want to remove service type from service model, from now we respect only offer type.

Tasks:
- [x] unification of service summary view

goreck888:
Make summary independent of offer type
Fixes #1201


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Internal server error message (@goreck888)
 - Anonymous user can enter service offer selection section (@mkasztelnik)
 - Service access button (@martaswiatkowska)
+- Service and offer type enum values (@goreck888)
```
martaswiatkowska: This changes are not connected to this pr

app/views/services/summaries/show.html.haml
```diff
@@ -23,14 +23,12 @@
         %dt Voucher ID
         %dd= @project_item.voucher_id
 
-
   = render "projects/details", project: @project_item.project
 
 = content_for :next_button do
-  = link_to t("service.normal.order.title"),
+  = link_to t("service.#{@service&.service_type}.order.title"),
```
martaswiatkowska: I think it should be @offer.offer_type

goreck888: Done.

CHANGELOG.md
goreck888: Done.

app/models/project_item.rb
```diff
@@ -97,12 +97,12 @@ def research_area_is_a_leaf
     errors.add(:research_area_id, "cannot have children") if research_area&.has_children?
   end
 
-  def vaucherable?
+  def voucherable?
```
mkasztelnik: Please avoid changes not related to the problem you are solving. Such modification should be done in separate PR.

app/views/services/summaries/show.html.haml
```diff
@@ -23,14 +23,12 @@
         %dt Voucher ID
         %dd= @project_item.voucher_id
 
-
   = render "projects/details", project: @project_item.project
 
 = content_for :next_button do
-  = link_to t("service.orderable.order.title"),
+  = link_to t("service.#{@project_item&.offer.offer_type}.order.title"),
```
mkasztelnik: This line has 2 problems:
  1. `@project_item&.offer` can return nil if @project_item is nil, then an exception will be thrown
  1. such variables should be set in the controller, a view should be logic-less.

app/models/project_item.rb
goreck888: Ok, I can revert these changes.

app/views/services/summaries/show.html.haml
goreck888: Do you mean I forget about `&.offer_type`?
The same thing is in Marta's PR: 
https://github.com/cyfronet-fid/marketplace/pull/1224/files#diff-e32da2486f29a680e2ab6d4305bbe868R3

app/models/project_item.rb
goreck888: Marked as protip for the future :)

PR #1224, opened 2019-11-05
Offer type is required

martaswiatkowska:


app/views/services/configurations/show.html.haml
```diff
@@ -1,7 +1,7 @@
 = render "layouts/order/nav", service: @service, step: 3
 
-%h2.mb-4.font-weight-normal= t "service.#{@service.service_type}.order.configuration.title"
-%p.mb-4= t "service.#{@service.service_type}.order.configuration.description"
+%h2.mb-4.font-weight-normal= t "service.#{@project_item&.offer.offer_type}.order.configuration.title"
```
bwilk: General question - maybe for later enhancements (@mkasztelnik think about it as well). Project item should not be prone to changes in offers and services. Project item in general is meant to be saved forever once order/request/att-to-project-action is performed. What happens when the service(old version)/offer(current) type changes? My opinion is - project item should not change at all. I think the state of these properties should be copied to project_item once the order/... has been placed, and stay that way forever. Please take that into consideration for the future implementations...

spec/features/order_spec.rb
```diff
@@ -464,15 +464,15 @@
...
       service = create(:service)
-      create(:offer, offer_type: :open_access, service: service)
+      create(:open_access_offer, service: service)
 
       visit service_path(service)
 
       click_on "Access the service"
 
-      expect(page).to have_link("Go to the service")
+      expect(current_path).to eq service_information_path(service)
```
mkasztelnik: Why did you change this line? Now it tests different things. What is more, I don't see connection between this change and PR scope.

martaswiatkowska: I cleaned it up

PR #1142, opened 2019-10-08
Allow only child research areas in service

goreck888:
This PR contains a change the way user can choose research area.
Previously there was a possibility to chose from all research areas.
From now we can choose only from the child research areas.

Fixes #1089

PR #1226, opened 2019-11-06
Move project selection to service order summary step

### LINKED ISSUE #1223: Move project selection to the summary step
The new ordering process needs to move project selection to its 4th step

Make a wizard for the ordering process and then move project selection to the last step of order.

**Additional context**
https://medium.com/@nicolasblanco/developing-a-wizard-or-multi-steps-forms-in-rails-d2f3b7c692ce

mkasztelnik:
At the beginning moving one field selection to the last step looks easy but it was not true. We based on the assumption that `project_item` will be validated fully started from configuration step. This is not true after project selection was moved to the last order wizard step. 

To make it work validation mechanism for the order process was refactored. Now we have a kind of wizard object with steps. Each step introduces validation for all previous steps and current step validation. This refactoring was done in the first 3 PR commits. Please take a look that no spec changes were introduced here (so no API changes). 

After the refactoring moving project selection to the last step was easy (thx https://www.youtube.com/watch?v=LhX5COR8WXc for the inspiration). This was done in the last PR commit. 

Fixes #1223

### Future TODOs
  1. Since I had limited time to finish this PR not everything is perfect. As you can see there is a lot of duplication in each order step controller. I would love to work on it to refactor it and improve it in the near feature. 
  1. Other nice thing related to the wizard and step model is that we can use it to calculate if the step should be shown. This can be done by introducing `show?` method in each wizard step - the challenge here is to have all the knowledge in the step model to make the decision. 
  1. Order specs are ugly, there is a lot of clicking "Next" sections. As a result, it is hard to see what is going on in the spec - this for sure should be refactored.


bwilk:
Let me paste the log:

```
[2019-11-07T09:39:01.859052 #1] DEBUG -- : [25958ff1-c618-4bad-8770-7ff458888a90] undefined local variable or method `properties' for #<ProjectItem::Wizard::ConfigurationStep:0x000055728602a4d8>
Did you mean?  property_values excluded from capture: DSN not set
F, [2019-11-07T09:39:01.860382 #1] FATAL -- : [25958ff1-c618-4bad-8770-7ff458888a90]   
F, [2019-11-07T09:39:01.860457 #1] FATAL -- : [25958ff1-c618-4bad-8770-7ff458888a90] ActionView::Template::Error (undefined local variable or method `properties' for #<ProjectItem::Wizard::ConfigurationStep:0x000055728602a4d8>
Did you mean?  property_values):
F, [2019-11-07T09:39:01.860593 #1] FATAL -- : [25958ff1-c618-4bad-8770-7ff458888a90]     1: - unless project_item.property_values.blank?
[25958ff1-c618-4bad-8770-7ff458888a90]     2:   %h5.mb-4.mt-5.font-weight-bold.configuration-title.text-uppercase Technical parameters
[25958ff1-c618-4bad-8770-7ff458888a90]     3:   - project_item.property_values.each do |attr|
[25958ff1-c618-4bad-8770-7ff458888a90]     4:     = form.simple_fields_for :property_values, attr do |field|
F, [2019-11-07T09:39:01.860642 #1] FATAL -- : [25958ff1-c618-4bad-8770-7ff458888a90]   
F, [2019-11-07T09:39:01.860695 #1] FATAL -- : [25958ff1-c618-4bad-8770-7ff458888a90] app/models/project_item/customization.rb:27:in `part'
[25958ff1-c618-4bad-8770-7ff458888a90] app/models/project_item/customization.rb:16:in `property_values'
[25958ff1-c618-4bad-8770-7ff458888a90] app/views/services/configurations/_attributes.html.haml:1:in `_app_views_services_configurations__attributes_html_haml__2240524930348473940_46975169903220'
[25958ff1-c618-4bad-8770-7ff458888a90] app/views/services/configurations/show.html.haml:10:in `block in _app_views_services_configurations_show_html_haml___4537483617090280221_46975175292380'
[25958ff1-c618-4bad-8770-7ff458888a90] app/views/services/configurations/show.html.haml:7:in `_app_views_services_configurations_show_html_haml___4537483617090280221_46975175292380'
```

app/models/project_item/wizard.rb
```diff
@@ -0,0 +1,45 @@
...
+    def error
+      "Please select one of the offer"
+    end
+  end
+
+  class ConfigurationStep < OfferSelectionStep
+    include ProjectItem::Customization
+    include ProjectItem::VoucherValidation
+
+    delegate :created?, to: :project_item
```
bwilk: you probably need to delegate `properties` attribute as well - vide the error at deployment instance

bwilk: maybe `property_values` should be used here - I'm not sure. Sorry I had too little time to check in detail

mkasztelnik: This exception is very odd. Delegation to properties is done here:

```ruby
delegate(*::ProjectItem.attribute_names.map { |a| [a, "#{a}="] }.flatten, to: :project_item)
```

Take a look at `ProjectItem.attribute_names` result:

```
[1] pry(main)> ProjectItem.attribute_names
=> ["id",
 "status",
 "created_at",
 "updated_at",
 "issue_id",
 "issue_status",
 "project_id",
 "offer_id",
 "properties",
 "request_voucher",
 "voucher_id",
 "iid"]
```

`properties` attribute is there, so change introduced by @martaswiatkowska duplicates this behavior, what is more in the code without last commit:

````
[4] pry(main)> w=ProjectItem::Wizard::ConfigurationStep.new(offer: Offer.first).properties
=> []
```

agpul:
after Marta changes now the same situation occurs on the next step. So after clicking "Next" on Technical Parameters step, I have the above error.

goreck888:
I tried to investigate issues in this PR and i found that:
- we use in this case 2 reserved words: `attribute` and `Configuration` (source: http://reservedwords.herokuapp.com/words)
- we don't use `attr_accessor` for `hidden_field :foo` (e.g. in #1142 I had to add `attr_accessor :move_ra_id` to send form)
- It may be caused by some variable in production.rb or some environment variable on our machine.

I can't see logs so I don't know if my commit changes something but I tried to do something with all these cases.

mkasztelnik:
The problem described by @agpul and @bwilk was connected with the way how we are starting an application for PR. We are using docker for this and the source of the problem was connected with order how an application is started and DB structure is created. To recreate the bug you need to:
  * create db 
  * start application in production mode
  * run db migrations and seeds

Application started in production mode is loading all classes on startup. The active model class is inspecting DB structure and discover table columns. As a result properties column was not found + there is a magic connected how `delegate` is done. All these small pieces lead to the error described above. 

To solve this issue @wziajka will change the way how PRs are building. He will make sure that the application will be started only when DB is created and all migrations are performed.

PR #1227, opened 2019-11-06
Use service type as order type while seeding data

mkasztelnik:
Before there was `nil` as a result all order views have corrupted translations.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -96,6 +96,7 @@ yaml_hash["services"].each do |_, hash|
       service.offers.create!(name: h["name"],
                              description: h["description"],
                              parameters: h["parameters"] || [],
+                             offer_type: service_type,
                              status: :published)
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -96,6 +96,7 @@
       service.offers.create!(name: h["name"],
                              description: h["description"],
                              parameters: h["parameters"] || [],
+                             offer_type: service_type,
```
martaswiatkowska: Doesn't it do the same thing?  https://github.com/cyfronet-fid/marketplace/blob/e200c4385c24fdcb1ddcf1f05cb6ec225ccd2a3d/app/models/offer.rb#L46

goreck888: You are right, but `create!` doesn't call model methods and it doesn't fill `offer_type` column.

PR #1228, opened 2019-11-07
[HOTFIX] EIC Import, invalid EIC urls

michal-szostak:
* Fix EIC import - in case of model validation errors log them, and
  don't import that particular service, instead of crashing whole task


wziajka:
Import looks ok.

PR #1243, opened 2019-11-13
Hide service order configuration step when no data to be configured

mkasztelnik:
The code responsible for deciding if a step should be visible was moved to the wizard step class. This leads to wizard, controllers, and views refactoring. As a conclusion now we can easily add new steps to the wizard, add visibility logic, etc.

This PR should be merged after #1226 is merged.


agpul:
image

Still visible Add to project tab.

PR #1245, opened 2019-11-14
Add affitional comment input to summary step

### LINKED ISSUE #1218: Create comment box in the Final step ( 3rd ) during ordering process
Now there is no place where the end-user can leave his comment to the order during the ordering process. 

We want to have a comment box created on the last page ( Final Step ) 3rd one. 

- [ ] Content: Please enter additional details of your request. 
- [ ] Comment box with the possibility to add TXT format 
- [ ] This field should be transferred to JIRA as a first comment and should be displayed in the CONTACT WITH SERVICE SUPPORT SECTION in Marketplace

martaswiatkowska:
Fix #1218

### DIFF
--- a/app/controllers/services/summaries_controller.rb
+++ b/app/controllers/services/summaries_controller.rb
@@ -16,7 +16,7 @@ class Services::SummariesController < Services::ApplicationController
     @step = step(summary_params)
 
     if @step.valid?
-      do_create(@step.project_item)
+      do_create(@step.project_item, message_text)
     else
       setup_show_variables!
       flash.now[:alert] = @step.error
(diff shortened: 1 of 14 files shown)


app/controllers/services/summaries_controller.rb
```diff
@@ -26,12 +26,18 @@ def create
...
       authorize(project_item_template)
 
       @project_item = ProjectItem::Create.new(project_item_template).call
 
       if @project_item.persisted?
+        if message_text
+          message = Message.new(message: message_text,
+                                author: current_user,
+                                messageable: @project_item)
+          Message::Create.new(message).call
```
mkasztelnik: We can have race conditions here (this can be a reason for the issue reported by @agpul): `ProjectItem::Create` triggers delayed job and `Message::Create` is triggers delayed job. You don't really know which delayed jobs will be executed in the first place. 

To solve this issue you need to pass the message into `ProjectItem::Create` next to the delayed job and only when project item is registered in jira trigger new job to post new user comment.

PR #1251, opened 2019-11-18
Fix no offer type selected in view

goreck888:
Add fix hiding badge if offer type is blank.

### DIFF
--- a/app/views/services/_offer.html.haml
+++ b/app/views/services/_offer.html.haml
@@ -5,7 +5,8 @@
         .card-label
           %i.fas.fa-ticket-alt
           %span Voucher
-      %span.text-sm-right.fa-pull-right.badge= t("service.type.#{offer.offer_type}")
+      - unless offer.offer_type.blank?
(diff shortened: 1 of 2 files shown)


app/views/services/_offer.html.haml
```diff
@@ -5,7 +5,8 @@
         .card-label
           %i.fas.fa-ticket-alt
           %span Voucher
-      %span.text-sm-right.fa-pull-right.badge= t("service.type.#{offer.offer_type}")
+      - unless offer.offer_type.blank?
```
mkasztelnik: `offer_type` should be mandatory. We should not need this change.

goreck888: I know that, but Marta didn't merge her PR with the db-level solution and Aga asked me if I could disable showing offer type if it's blank.

goreck888:
To reproduce:

As logged user with SPM role:
1. create service with no service type
2. create at least two offers, at least one of them without offer_type
3. Publish service and offers
4. Go to the order service view

PR #1254, opened 2019-11-18
Fix cancel order redirection

goreck888:
For non logged-in user order cancel link was redirecting
to the login page. This commit fixes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Active Order button in service preview (@goreck888)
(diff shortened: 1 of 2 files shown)


app/controllers/services/application_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Services::ApplicationController < ApplicationController
-  before_action :authenticate_user!
+  before_action :authenticate_user!, except: [:destroy]
```
mkasztelnik: This is not a good place for skiping authentication. Try to add:

```ruby
skip_before_action :authenticate_user!
```

into `Services::CancelsController`

goreck888: Ok

PR #1255, opened 2019-11-18
Move test service generation to `dev:prime` task

mkasztelnik:
`db:seed` should be used to setup production db with the data necessary to run the application. In our case, we used it to generate test services. To cleanup service generation was moved to `dev:prime` task.

This is important for @wziajka. After this is merged you need to update your building scripts for PRs and invoke `dev:prime` in the initialization script.


mkasztelnik:
`dev:prime` is not used anymore to seed PR. It is used only locally during the development

PR #1264, opened 2019-11-21
1089 children research areas view

goreck888:
in the backoffice service view added parent's name in research area

### DIFF
--- a/app/models/research_area.rb
+++ b/app/models/research_area.rb
@@ -13,4 +13,21 @@ class ResearchArea < ApplicationRecord
   def self.names
     all.map(&:name)
   end
+
+
+  private
+
+    def self.child_names(records = ResearchArea.arrange, parent_name = "", result = [])
+      records.each do |r, sub_r|
+        result << [name_with_path(parent_name, r.name), r]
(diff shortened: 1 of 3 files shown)


app/models/research_area.rb
```diff
@@ -13,4 +13,21 @@ class ResearchArea < ApplicationRecord
   def self.names
     all.map(&:name)
   end
+
+
+  private
+
+    def self.nested_names(records = ResearchArea.arrange, parent_name = "", result = [])
```
martaswiatkowska: Maybe "child_names"?

goreck888: Ok, i can change it

app/models/research_area.rb
```diff
@@ -13,4 +13,21 @@ class ResearchArea < ApplicationRecord
   def self.names
     all.map(&:name)
   end
+
+
+  private
+
+    def self.child_names(records = ResearchArea.arrange, parent_name = "", result = [])
```
martaswiatkowska: Maybe function "descendant" from "ancestry" will be helped to simplify this function

Issue #1140, opened 2019-10-08
Change information which we display for the user on the timeline 


agpul:
**Is your feature request related to a problem? Please describe.**
Information, which we display for the user on the timeline is poor and doesn't mean anything for the user. We need to change them.

**Describe the solution you'd like**
Make proper changes in the timeline. Take into consideration statuses (grey one) from the attached screens and change the information in the box.

Changes for orderable services (numbers are related to the enclosed picture) NORMAL TYPE:
- [ ] from 1. to Your service request has been created and is waiting for being registered 
- [ ] from 2. to Service request status: new
- [ ] from 3 to Your service request was registered and shared with service provider
- [ ] from 4 to Service request status: registred
- [ ] from 5 to Your service request is under review by service provider
- [ ] from 6 to Service request status: under review
- [ ] from 7 to Service provider is waiting for your response
- [ ] from 8 to Service request status: provide additional information 
- [ ] from 9 to Your service request is under review by service provider
- [ ] from 10 to Service request status: under review
- [ ] from 11 to You service request has been accepted and is currently being processed by service provider
- [ ] from 12 to Service request status: being prepared
- [ ] from 13 to Your message to the service provider
- [ ] from 14 to Message from the service provider
- [ ] from 15 to Your service is now ready to be used
- [ ] from 16 to Service request status: ready to use
- [ ] from 17 to Your service request is closed. If you want to use the service again, please contact service provider
- [ ] from 18 to Service request status: closed
- [ ] from 19 to Your service request has been rejected. Please contact service provider if you have any questions. 
- [ ] from 20 to Service request status: rejected

Changes for open_ access/ catalog services:
- [ ] do not show this tab any more

General changes:
- [ ] do not allow to write a message to the customer from this tab

**Additional context**
1. Orderable service (normal type):
![image.png](https://images.zenhubusercontent.com/5b8fbb317859ca34f6abbcb2/0bf41c73-d5fb-4e08-aa37-a298661121d6)
![image.png](https://images.zenhubusercontent.com/5b8fbb317859ca34f6abbcb2/5d71aba1-459d-4829-83e6-c22b90f7a70c)
![image.png](https://images.zenhubusercontent.com/5b8fbb317859ca34f6abbcb2/e5a4aea4-4b49-4a7c-b116-3f571a8fae89)
[...]

PR #1276, opened 2019-11-27
Split conversation and timeline on project service view

### LINKED ISSUE #1155: Create additional tab " Contact with Service Provider"  in the Service details page

Now the order history plus communication with provider is shown in one tab called " communication with project support". We would like to split this into two tabs.

As a result of this issue: We want to add tab "Contact with the service provider". To this tab, we want to move the communication feature. So we want to show only conversation between customer and provider.

- [ ] add a new tab called  " Contact with the service provider"  based on Andrzej prototype @abacz attach it here if it is ready
- [ ] move the communication box from "Timeline tab" to "Contact with the service provider" 
- [ ] change the label of the message box from Message to customer service about your project * to Message to the service provider
- [ ] if it is possible we should filter comments coming from JIRA and in MP we should display only comments marked in JIRA as "visible for the user"
- [ ] we will have two labels here: 
   - Your message to the service provider
   - Message from the service provider

Prev: https://kcjh3g.axshare.com/#g=1&p=service_1_normal

martaswiatkowska:
fix #1155 
Additionally labels text will be changed in pr to ticket #1140

### DIFF
--- a/app/controllers/projects/services/conversations_controller.rb
+++ b/app/controllers/projects/services/conversations_controller.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
 class Projects::Services::ConversationsController < ApplicationController
-  before_action :authenticate_user!
-  before_action :load_and_authorize_project_item!
+  include ProjectItem::Authorize
 
   def show
     @message = Message.new(messageable: @project_item)
@@ -33,16 +32,6 @@ class Projects::Services::ConversationsController < ApplicationController
     end
 
     def load_messages
-      @messages = (@project_item.messages + @project_item.statuses).
-                  sort_by(&:updated_at)
-    end
-
-    def load_and_authorize_project_item!
-      @project_item = ProjectItem.joins(:project).
-                      find_by(iid: params[:service_id],
-                              project_id: params[:project_id])
-      @project = @project_item.project
-
-      authorize(@project_item, :show?)
+      @messages = @project_item.messages.order(:updated_at)
     end
 end
--- a/app/controllers/projects/services/timelines_controller.rb
+++ b/app/controllers/projects/services/timelines_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class Projects::Services::TimelinesController < ApplicationController
+  include ProjectItem::Authorize
+
+  def show
+    load_timeline
+    load_projects
+  end
+
+  private
+
+    def load_projects
+      @projects = policy_scope(Project).order(:name)
+    end
+
+    def load_timeline
+      @statuses = @project_item.statuses.order(:updated_at)
+    end
+end
--- a/app/controllers/concerns/project_item/authorize.rb
+++ b/app/controllers/concerns/project_item/authorize.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module ProjectItem::Authorize
+  extend ActiveSupport::Concern
+
+  included do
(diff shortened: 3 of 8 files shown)


app/controllers/api/webhooks/jiras_controller.rb
```diff
@@ -45,4 +47,13 @@ def valid_jira_request?
     def jira_client
       @jira_client ||= Jira::Client.new
     end
+
+    def visible_for_user?
+      visibility = params["comment"]["visibility"] || nil
+      if visibility && visibility["value"] != "User"
```
mkasztelnik: You can simplify this method in the following way:

```ruby
def visible_for_user?
  # most probably it should be == not != right?
  params.digg("comment", "visibility", "value") == "User"
```

see https://apidock.com/ruby/Hash/dig for details.

app/controllers/api/webhooks/jiras_controller.rb
```diff
@@ -22,7 +22,9 @@ def create
       when "jira:issue_updated"
         Jira::IssueUpdated.new(element_updated, params["changelog"]).call
       when "comment_created"
-        Jira::CommentCreated.new(element_updated, params["comment"]).call
+        if visible_for_user?
```
mkasztelnik: Maybe `Jira::CommentCreated` service is a better place to put this logic?

app/controllers/api/webhooks/jiras_controller.rb
```diff
@@ -45,4 +47,13 @@ def valid_jira_request?
     def jira_client
       @jira_client ||= Jira::Client.new
     end
+
+    def visible_for_user?
```
mkasztelnik: If this logic will be moved to the service you can simplify the name of this method to `visible?`

app/controllers/projects/services/conversations_controller.rb
```diff
@@ -33,8 +33,7 @@ def load_projects
     end
 
     def load_messages
-      @messages = (@project_item.messages + @project_item.statuses).
-                  sort_by(&:updated_at)
+      @messages = @project_item.messages.sort_by(&:updated_at)
```
mkasztelnik: I know that is was done this way previously but I'm curious if we can use DB sort here instead of sorting records in the runtime.

app/controllers/projects/services/timelines_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  private
+
+    def load_projects
+      @projects = policy_scope(Project).order(:name)
+    end
+
+    def load_messages
+      @messages = @project_item.statuses.sort_by(&:updated_at)
```
mkasztelnik: Similar like above, maybe we can use DB sort here?

app/controllers/projects/services/timelines_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def load_and_authorize_project_item!
```
mkasztelnik: `Projects::Services::TimelinesController` and `Projects::Services::ConversationsController` controllers share the same logic for loading and authorizing `project_item`. Please consider extracting this logic into concern and reuse it there.

app/controllers/api/webhooks/jiras_controller.rb
mkasztelnik: This change is not connected with the rest of this PR (and with the PR description), maybe such changes can be introduced in separate PRs in the future?

PR #1286, opened 2019-11-29
Don't register internal messages

mkasztelnik:
This is an extraction (and small improvements) of  "don't register internal messages" feature from #1276. Basically it skips internal messages reported by Jira.


mkasztelnik:
There are 3 message types: 
  1. visible to all (default)
  2. visible only to the user
  3. visible only to the admin

We are handling correctly cases 2 and 3. We need to improve our code to include also case 1.

PR #1291, opened 2019-12-04
1268 order path enhancement

jarekzet:
added mods:
1. Technical Parameters page view is enhanced the same way/ as Final page.
2. Comment box to be sized based on the text


app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -3014,10 +3010,36 @@ textarea.form-control.is-invalid {
...
+.technical {
+  .form-control-label {
+    border-bottom: 2px solid $palette-yellow;
+    font-size: 0.875rem;
+    padding-bottom: 0.35rem;
+    text-transform: uppercase;
+    margin-bottom: 17px;
+    font-weight: bold;
+    width: 100%;
+  }
```
goreck888: I think if you apply the same styles to the different classes you can use in this case:
```
.col-form-label .technical.form-control-label {
    border-bottom: 2px solid $palette-yellow;
    font-size: 0.875rem;
    padding-bottom: 0.35rem;
    text-transform: uppercase;
    margin-bottom: 17px;
    font-weight: bold;
    width: 100%;
  }
```

goreck888:
Moreover add an entry in changelog :)

app/javascript/stylesheets/_bootstrap-customizations.scss
jarekzet: added this mod

PR #1302, opened 2019-12-09
1283 access instructions

jarekzet:
added 3 types of access information checkout step


jarekzet:
added illustration positioning and mobile version paddings/margins fixes

Issue #1298, opened 2019-12-05
Improve UI in the ordering form and on project item


agpul:
1 ------------------
- [x]  In this view, we need to improve the order history and contact with the service provider section
![image](https://user-images.githubusercontent.com/43000481/70242443-20f6a600-1771-11ea-88f7-4dcf3067d350.png)
What we have and we can use it this view: https://invis.io/NASYZ4Y9CU8#/373762364_Projects_Sidebar_Contact_With_Support

2 ------------------
- [ ] After choosing offer and clicking "Go to the service", a new Tab opens in the browser. We should rethink, if it is a good solution
![Adnotacja 2019-12-16 095707.png](https://images.zenhubusercontent.com/5d89b22b4cfe78000124e106/f771561e-73d6-4f0f-b867-a839698715fd)

3 ------------------
- [ ] The shape of the "arrow" between "Offer selection" and "Access instructions" is sharper then the "arrows" between "Access instructions" and "Configuration"
![Adnotacja 2019-12-16 095707 strzalki.png](https://images.zenhubusercontent.com/5d89b22b4cfe78000124e106/3fbb861a-5f23-46b5-9907-19b1110fa3d3)

4 ------------------
- [ ] The whole View in "Review" has to be chekcked - some questions there could be unclear to user ![Adnotacja 2019-12-12 140010.png](https://images.zenhubusercontent.com/5d89b22b4cfe78000124e106/5912caf9-202e-41a6-bcfa-ef2be86e6978)

5 ------------------
- [ ] in the view of an service, two buttons have lower case letters, the other has upper case letters. Should it be the same or ist this the way it was designed
![Adnotacja 2019-12-12 140512.png](https://images.zenhubusercontent.com/5d89b22b4cfe78000124e106/4356eced-b9aa-4f1f-8581-696d9364119f)

6 ------------------
- [ ] The Logo of service providers is not stretched and that does not look estetic. There is an thin frame around the logo, which

abacz:
Order state should be align center (vertically).

PR #1312, opened 2019-12-10
service order conversation tab

jarekzet:
Mods according to #1298


app/javascript/stylesheets/_bootstrap-customizations.scss
```diff
@@ -1686,10 +1686,44 @@ h5 {
...
 
+// scss-lint:disable all
 .question-box {
   border-top: 1px solid $palette-gray-10;
+
+  .btn-primary {
+    margin-left: 70px;
+  }
+
+  .message_message {
```
goreck888: Is this automatically created class?

PR #1315, opened 2019-12-11
[#1285] Disable visibility of input type attribute

### LINKED ISSUE #1285: "Number of days" attribute is shown in Technical Parameters in the offer view
**Describe the bug**
"Number of days" is shown in Technical Parameters in the offer view but in fact, this attribute isn't the technical parameter of the service

**To Reproduce**
1. Go to https://marketplace.eosc-portal.eu/services/b2drop
2. Scroll down to the offer, eg. B2DROP For Researchers 
3. See error

**Expected behavior**
We shouldn't display the Number of days attributes in the technical parameters section on the offer view. 

**Screenshots**
image

goreck888:
In the offer view disabled all input attributes, because they haven't minimum and maximum values, so they can be treated the same way like string attribute.
Fixes #1285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Statistics for executive user group (@mkasztelnik)
 
 ### Changed
+- Disable visibility of `input` type attribute on the offer view (@goreck888)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


goreck888:
I'll add a flag changing visibility `show_in_offer` and then we can hide or show any of attributes.

PR #1323, opened 2019-12-16
Project item Independence Day

### LINKED ISSUE #1294: Project items are prone to service and offer updates
#### Problem definition 

Project items are referencing service and offers in many places. As the project item reflects the "order", in theory we should even be able to delete services and offers having the orders intact. 

#### Proposed solution 

Copy all the required fields from offers and services to project_item

#### TODO 

Discuss the solution

mkasztelnik:
Fixes #1294

### DIFF
--- a/db/migrate/20191216141227_add_offer_fields_to_project_item.rb
+++ b/db/migrate/20191216141227_add_offer_fields_to_project_item.rb
@@ -1,0 +1,30 @@
+class AddOfferFieldsToProjectItem < ActiveRecord::Migration[5.2]
+  def up
+    add_column :project_items, :name, :string
+    add_column :project_items, :description, :text
+    add_column :project_items, :offer_type, :string
+    add_column :project_items, :webpage, :string
+    add_column :project_items, :voucherable, :boolean, default: false, null: false
+
+    execute(<<~SQL
+      UPDATE project_items
+      SET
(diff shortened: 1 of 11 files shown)


db/migrate/20191216141227_add_offer_fields_to_project_item.rb
```diff
@@ -0,0 +1,22 @@
...
+    execute(<<~SQL
+      UPDATE project_items
+      SET
+        offer_type = (SELECT offer.offer_type FROM offers offer WHERE offer.id = offer_id),
+        name = (SELECT offer.name FROM offers offer WHERE offer.id = offer_id),
+        description = (SELECT offer.description FROM offers offer WHERE offer.id = offer_id)
+    SQL
+    )
+  end
+
```
goreck888: Schema and data migrations should be separate.

mkasztelnik: I don't agee. This should be invoked in one transaction. Splitting into 2 migration can lead to inconsistent data in the DB.

mkasztelnik:
@bwilk thx for hints. Solved all (I even found more) except ` app/models/jira/client.rb:277`. This should be good enough since we don't give the user the possibility to change the relationship between offer and service. What is more, an offer cannot be removed when it has project items attached.

PR #1337, opened 2020-01-03
Comment update

### LINKED ISSUE #1277: Comments update
Editing Comment in jira is propagated to marketplace . Minimal requirements:
1. If Operator edits the comment in JIRA, the updated comment will be sent to MP and the proper label (edited) will be applied.
2. If Operator changes rights from visible to admin to visible to the user, we will send such comments to MP.
3, We don't want to operate cases if  someone deletes the comment and if Operator changes rights from visible to the user to visible to admin

If someone edits the comment in jira, changes are displayed in marketplace

martaswiatkowska:
Comment update 
@agpul can you please add in comment new subject to email?

Fix #1277


app/controllers/api/webhooks/jiras_controller.rb
```diff
@@ -23,15 +23,31 @@ def create
         Jira::IssueUpdated.new(element_updated, params["changelog"]).call
       when "comment_created"
         Jira::CommentCreated.new(element_updated, params["comment"]).call
+      when "comment_updated"
+        if allowed_for_user || !params["comment"]["visibility"]
```
mkasztelnik: Such logic should not be here but in the service. Take a look into `Jira::CommentUpdated#owned?` method for the inspiration :)

app/views/webhook_jira_mailer/message_updated.html.haml
```diff
@@ -0,0 +1,4 @@
+%p Dear , Sir/Madame
+%p
+  The below message related to your project #{@project_name} has been edited by service provider.
```
mkasztelnik: This message is not clear to me @agpul, @MinLidia can you propose better text?

db/migrate/20200102135056_add_was_updated_to_messages.rb
```diff
@@ -0,0 +1,5 @@
+class AddWasUpdatedToMessages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :messages, :was_updated, :boolean, default: false
```
mkasztelnik: Maybe `edited` is a better name?

mkasztelnik:
I was looking into existing code and it can be simplified a bit. You can try to reuse code from `Jira::CommentCreated` and there add logic responsible to check if we are creating new comment or updating existing one (then you should rename the class to something like `Jira::CommentActivity`). 

You can use [`create_of_find_by`](https://blog.bigbinary.com/2019/03/25/rails-6-adds-create_or_find_by.html) for this purpose

PR #1348, opened 2020-01-13
Services search view refactoring

goreck888:
## What's in this PR?

This PR contains views refactoring and reuses them in the backoffice context.
It makes non-visible changes in services search view, but changes views in the backoffice services view and makes search more similar and consistent with main mp layout.

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -3,6 +3,128 @@
     {
       "warning_type": "Mass Assignment",
       "warning_code": 70,
+      "fingerprint": "0b0b808d8841fc9690e7fcaf7a60b846d6f6f70453b446772bd45cda32f0ec80",
+      "check_name": "MassAssignment",
+      "message": "Parameters should be whitelisted for mass assignment",
+      "file": "app/views/services/_paginate_simple.haml",
+      "line": 8,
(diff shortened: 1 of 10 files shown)


config/brakeman.ignore
```diff
@@ -1,5 +1,127 @@
 {
   "ignored_warnings": [
+    {
```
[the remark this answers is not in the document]
goreck888: Yes, because these warnings are also ignored in different context (main view) and added by yourself.

app/views/services/_banner.html.haml
```diff
@@ -0,0 +1,28 @@
+.col-md-12.mb-2
```
mkasztelnik: Maybe `active_queries` or `active_filters` is a better more meaningful name?

config/brakeman.ignore
mkasztelnik: You should run `brakeman -I` and prune obsolete ignored warnings

app/views/services/_banner.html.haml
goreck888: Ok, I change it.

config/brakeman.ignore
goreck888: Done,

PR #1353, opened 2020-01-14
Help in DB

mkasztelnik:
This is a proof of concept in which the main goal is to check how [trix](https://trix-editor.org/) and [action text](https://edgeguides.rubyonrails.org/action_text_overview.html) can be used to create customizable page content. The idea is quite simple: 2 new models are created (`HelpCategory` and `HelpItem`). Inside `TextItem` rich text field is defined. It stores help item content with all formatting. 

This feature has 2 types of views. The first one is in the admin section, where help can be built. The second view is located at `/help`. For now, this section is only visible for admin, other users are redirected to the old help page. 

The plan for this feature in this release is to validate the idea and check if rich text editing is good enough for the help creation. So after the release @MinLidia and @agpul will try to move already existing static help into help stored in the DB. If everything will be ok we will start phase 2 (planned for next release), where old static help will be removed and the user will see the new help section rendered from the DB.

This kind of views is also a perfect candidate to use [cache](https://guides.rubyonrails.org/caching_with_rails.html). This is also something worth to add in the next help section version. 

Another improvement that can be added here is to have moved up/down capability on the main help builder view instead of manually changing the `position` field. To make it simpler we can use [this](https://github.com/brendon/acts_as_list) gem.

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -8,9 +8,3 @@
 // layout file, like app/views/layouts/application.html.erb
 
 import "app"
-
-
-
-
-
-
--- a/app/views/active_storage/blobs/_blob.html.erb
+++ b/app/views/active_storage/blobs/_blob.html.erb
@@ -1,0 +1,14 @@
+<figure class="attachment attachment--<%= blob.representable? ? "preview" : "file" %> attachment--<%= blob.filename.extension %>">
+  <% if blob.representable? %>
+    <%= image_tag blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ]) %>
+  <% end %>
+
+  <figcaption class="attachment__caption">
+    <% if caption = blob.try(:caption) %>
+      <%= caption %>
+    <% else %>
+      <span class="attachment__name"><%= blob.filename %></span>
+      <span class="attachment__size"><%= number_to_human_size blob.byte_size %></span>
+    <% end %>
+  </figcaption>
+</figure>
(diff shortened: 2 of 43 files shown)


app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+  <figcaption class="attachment__caption">
+    <% if caption = blob.try(:caption) %>
+      <%= caption %>
+    <% else %>
+      <span class="attachment__name"><%= blob.filename %></span>
+      <span class="attachment__size"><%= number_to_human_size blob.byte_size %></span>
+    <% end %>
+  </figcaption>
+</figure>
```
goreck888: What's the reason of using erb instead of haml?

spec/features/help_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(page).to have_current_path(page_path("help"))
+  end
+
+  context "as admin" do
+    let(:admin) { create(:user, roles: [:admin]) }
+
+    before { checkin_sign_in_as(admin) }
+
+    scenario "I can see gemerated help page" do
```
martaswiatkowska: generated?

app/javascript/packs/application.js
```diff
@@ -62,3 +62,6 @@ document.addEventListener('DOMContentLoaded', function () {
     initFlash();
     dom.watch();
 });
+
+require("trix")
+require("@rails/actiontext")
```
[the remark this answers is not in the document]
mkasztelnik: Why the new line should be added here?

app/views/active_storage/blobs/_blob.html.erb
mkasztelnik: This is a template from action text generated while running the generator. This can be converted to haml once we will be customizing this view. Now I don't see a reason to do this.

PR #1355, opened 2020-01-16
Overriding views

martaswiatkowska:
New env variable:
CUSTOM_VIEWS_PATH="app/views/custom"  for custom views 
In the folder app/view/custom, add the files you want to overwrite in the same structure as they are in app/view/

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,3 +219,13 @@ You can read documentation [here](https://github.com/bkeepers/dotenv).
 
 In shourt you can store your env variables in `.env` file in the root of the project.
 
+## View Customization
+
+View can be customized by adding two env variables:
+  * CUSTOM_VIEWS_PATH="path to external folder" to customize views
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -30,4 +30,9 @@ def not_authorized_message(exception)
       policy_name = exception.policy.class.to_s.underscore
       I18n.t "#{policy_name}.#{exception.query}", scope: :pundit, default: :default
     end
+
+    def custom
```
mkasztelnik: Is it necessary since additional directory is defined in `config/application.rb`?

app/views/custom/home/index.html.haml
```diff
@@ -0,0 +1 @@
+CUSTOM HOME
```
mkasztelnik: I propose to remove this customization in final PR and describe in `README.md` how to customize views.

PR #1357, opened 2020-01-16
Fix text mails

martaswiatkowska:
Fixed displaying links in conversation mails and emails preview.

### DIFF
--- a/spec/mailers/previews/webhook_jira_preview.rb
+++ b/spec/mailers/previews/webhook_jira_preview.rb
@@ -5,11 +5,11 @@
 # !!! We are using last created project_item to show email previews !!!
 class WebhookJiraPreview < ActionMailer::Preview
   def new_message_for_project_item
-    WebhookJiraMailer.new_message(ProjectItem.last, "message content")
+    WebhookJiraMailer.new_message(Message.new(messageable: ProjectItem.last, message: "message content"))
   end
 
 
   def project_new_message
-    WebhookJiraMailer.new_message(Project.last, "message content")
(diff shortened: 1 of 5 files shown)


app/models/project_item.rb
```diff
@@ -86,7 +86,7 @@ def new_voucher_change(voucher_id, author: nil, iid: nil)
   end
 
   def to_s
-    "\"#{project.name}##{id}\""
+    "\"#{project.name}##{id}\"".html_safe
```
mkasztelnik: `html_safe` is a view concern. Here we are at the model level which should not know anything about the presentation. Can you explain why `html_safe` is needed in the first place, next if really needed such constructs should be placed in the view layer.

spec/mailers/previews/webhook_jira_preview.rb
```diff
@@ -5,11 +5,11 @@
...
 class WebhookJiraPreview < ActionMailer::Preview
   def new_message_for_project_item
-    WebhookJiraMailer.new_message(ProjectItem.last, "message content")
+    WebhookJiraMailer.new_message(Message.new(messageable: ProjectItem.last, message: "aaaaaaa"))
   end
 
 
   def project_new_message
-    WebhookJiraMailer.new_message(Project.last, "message content")
+    WebhookJiraMailer.new_message(Message.new(messageable: Project.last, message: "aaaaaaa"))
```
mkasztelnik: I believe that `message content` (or other lorem ipsum) is better than `aaaaaaaa`.

PR #1363, opened 2020-01-22
[#1049] Service popularity based on number of orders

### LINKED ISSUE #1049: Service popularity based on number of orders
Right now we show popular services at the start page. Popularity is not under control and service list shown there is based on random sample taken from database.  

To start with something we should base service popularity on number of orders  

**Describe alternatives you've considered**
Google analytics clicks, rating

**Additional context**
none

goreck888:
Change the way of sorting services by popularity
After the fix it is populated based of number of orders
Fixes #1049

After merge and deploy in a production instance I'll ask to run task `rake service_counter` to get proper counter values.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -16,7 +16,7 @@ class HomeController < ApplicationController
       @providers_number = Provider.count
       @services_number = Service.count
       @countries_number = 32
-      @services = Service.published.includes(:providers).order(rating: :asc, title: :desc).limit(6)
+      @services = Service.published.includes(:providers).popular(6)
     end
 
     def load_platforms
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -77,6 +77,7 @@ class Service < ApplicationRecord
   enum status: STATUSES
 
   has_many :offers, dependent: :restrict_with_error
+  has_many :project_items, through: :offers
   has_many :categorizations, dependent: :destroy
   has_many :categories, through: :categorizations
   has_many :service_opinions, through: :project_items
@@ -137,6 +138,10 @@ class Service < ApplicationRecord
   validates :order_target, allow_blank: true, email: true
 
   after_save :set_first_category_as_main!, if: :main_category_missing?
+
+  def self.popular(count)
+    order(project_items_count: :desc, rating: :desc, title: :asc).limit(count)
+  end
 
   def main_category
     @main_category ||= categories.joins(:categorizations).
--- a/db/migrate/20200128142804_add_project_items_count_to_services.rb
+++ b/db/migrate/20200128142804_add_project_items_count_to_services.rb
@@ -1,0 +1,5 @@
+class AddProjectItemsCountToServices < ActiveRecord::Migration[6.0]
+  def change
+    add_column :services, :project_items_count, :integer, null: false, default: 0
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Services search view refactoring and unification with backoffice view (@goreck888)
 - Webpack entry structure (@martaswiatkowska)
 - Sidebar in the service view generic rendering (@goreck888)
+- Popular services main page section based on number of orders (@goreck888)
 
 ### Deprecated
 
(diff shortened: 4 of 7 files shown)


app/controllers/home_controller.rb
```diff
@@ -16,7 +16,8 @@ def load_services
       @providers_number = Provider.count
       @services_number = Service.count
       @countries_number = 32
-      @services = Service.published.includes(:providers).order(rating: :asc, title: :desc).limit(6)
+      @services = Service.published.includes(:providers)
```
mkasztelnik: Here we are doing `n+1` queries (take a look at `Service.project_items` method). This is not good since this is our main page and we can have a lot of DB queries here. This can be solved by:
  1. introducing a counter cache for service orders. This field next can be used in DB query
  2. it can also be good enough for now to have a more advanced DB query which will do the sorting.

goreck888: Check my change

goreck888: Now I use pure SQL query but with outer join, so it returns list of all services and orders by popularity. Tests are passing now.

app/models/service.rb
```diff
@@ -123,6 +123,19 @@ def main_category
                                   find_by(categorizations: { main: true })
   end
 
+  def project_items
+    self.offers.each.flat_map { |o| o.project_items }
+  end
+
+  def self.orders_number
+    Service.find_by_sql("SELECT s.*, COUNT(distinct pi.id)
```
mkasztelnik: This is better but maybe you can use arel instead of pure sql query, because this is not composable.

goreck888: Ok, I did it by counter_culture as you suggested

app/models/service.rb
```diff
@@ -123,6 +124,19 @@ def main_category
                                   find_by(categorizations: { main: true })
   end
 
+  def project_items
```
mkasztelnik: Why do you need this method?

app/models/service.rb
```diff
@@ -123,6 +124,19 @@ def main_category
...
+    self.offers.each.flat_map { |o| o.project_items }
+  end
+
+  def self.orders_number
+    Service.find_by_sql("SELECT s.*, COUNT(distinct pi.id)
```
mkasztelnik: Since counter cache is defined for numer of ordered services you can use it instead of this complex raw sql query.

app/models/service.rb
```diff
@@ -123,6 +124,19 @@ def main_category
                                   find_by(categorizations: { main: true })
   end
 
+  def project_items
+    self.offers.each.flat_map { |o| o.project_items }
+  end
+
+  def self.orders_number
```
mkasztelnik: I propose to change this method into `def popular(count: 6)`. It should improve the readability of this line:

```ruby
Service.includes(:providers).orders_number.first(6)
#vs
Service.includes(:providers).popular(count: 6)
```

db/migrate/20200128142804_add_project_items_count_to_services.rb
```diff
@@ -0,0 +1,9 @@
+class AddProjectItemsCountToServices < ActiveRecord::Migration[6.0]
+  def self.up
```
mkasztelnik: I believe that `change` is enough here since `add_column` is reversible

app/models/service.rb
goreck888: I don't need it, but I didn't push commit amend.

db/migrate/20200128142804_add_project_items_count_to_services.rb
goreck888: I used counter culture generator, fixed.

app/models/service.rb
goreck888: Bad commit, fixed.

app/models/service.rb
goreck888: Bad commit, fixed.

app/controllers/home_controller.rb
```diff
@@ -16,7 +16,7 @@ def load_services
       @providers_number = Provider.count
       @services_number = Service.count
       @countries_number = 32
-      @services = Service.published.includes(:providers).order(rating: :asc, title: :desc).limit(6)
+      @services = Service.includes(:providers).sort_by { |s| [-s.project_items_count, -s.rating] }.first(6)
```
mkasztelnik: Convert it into `order`

goreck888: Done.

PR #1367, opened 2020-01-26
[#1365] Services information for monitoring tool

### LINKED ISSUE #1365: Extraction of the service information for integration with the monitoring tool (ARGO) - part 2

Plan is to have a first easy integration of the EOSC Portal with the monitoring tool (ARGO).  We need to provide information about services in the EOSC Portal & Marketplace to ARGO team to enable a first monitoring.

- [ ]  Extract Service Unique ID and add it to already extracted information ( Doc). ( part 1) 
- [ ] Publishing the output to a public URL

Doc : 

[https://app.zenhub.com/files/135438250/4cbbfda0-3e02-44e7-8fea-138e45bd97fd/download](https://app.zenhub.com/files/135438250/4cbbfda0-3e02-44e7-8fea-138e45bd97fd/download)

goreck888:
Add api endpoint for downloading services information
for ARGO monitoring tool
JSON is available from #{root_path}/api/services
Fixes #1365

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class Api::ServicesController < ActionController::API
+  def index
+    @json = Service.where(status: [:published, :unverified]).map { |s| { "serviceUniqueId": s.id,
+                                                                         "serviceType": "eu.eosc.portal.services.url",
+                                                                         "contactEmail": s.contact_emails,
+                                                                         "sitenameServicegroup": s.title,
+                                                                         "countryName": s.places,
+                                                                         "url": s.webpage_url } }
+    render json: @json
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Add static pages for design system and UI documentation (@mkasztelnik)
 - Comment changes in jira propagate to mp (@martaswiatkowska)
 - Dynamic help sections visible currently only for admin (@mkasztelnik)
+- Service information for a monitoring tool through `api/services` (@goreck888)
 
 ### Changed
 - Move research area and category logos to active storage (@mkasztelnik)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -75,6 +75,7 @@ Rails.application.routes.draw do
   end
 
   namespace :api do
+    get "/services" => "services#index", defaults: { format: :json }, as: :services_api
     namespace :webhooks do
       post "/jira" => "jiras#create", as: :jira
     end


app/controllers/services_controller.rb
```diff
@@ -28,6 +28,19 @@ def show
     @question = Service::Question.new(service: @service)
   end
 
+  def json
```
mkasztelnik: This is not a good place for API controllers. We have dedicated space `app/controllers/api` for API requests. What is more try to use `CRUD` methods instead of custom one. In this case I propose to have something like `app/controllers/api/services_controller.rb` and inside this controller `index` method.

app/controllers/services_controller.rb
```diff
@@ -28,6 +28,19 @@ def show
...
+    @json = Service.where(status: [:published, :unverified]).map { |s| { "Service Unique ID": s.id,
+                                                                 "SERVICE_TYPE": "eu.eosc.portal.services.url",
```
mkasztelnik: I know that the keys you are using here are defined in the document attached to the issue, but this these names are not good... we should stick to one convention and use it everywhere. I propose to use `camelCase` for json since json comes from JS world, where camel case is de-facto a standard.

app/controllers/services_controller.rb
```diff
@@ -28,6 +28,19 @@ def show
...
+                                                                 "CONTACT_EMAIL": s.contact_emails,
+                                                                 "SITENAME-SERVICEGROUP": s.title,
+                                                                 "COUNTRY_NAME": s.places,
+                                                                 "URL": s.webpage_url } }
+    respond_to do |format|
+      format.html { @json }
```
mkasztelnik: Why this should respond to html request?

app/controllers/services_controller.rb
goreck888: Yes I know, but convention used in this case is an external requirement. I ask @MinLidia if we could use camelCase.

app/controllers/services_controller.rb
goreck888: Ok, I supposed this is proper way to do that thing.

app/controllers/services_controller.rb
goreck888: Fixed.

app/controllers/api/services_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Api::ServicesController < ActionController::API
+  def index
+    @json = Service.where(status: [:published, :unverified]).map { |s| { "serviceUniqueId": s.id,
```
mkasztelnik: It seems like the policy defined in `ServicePolicy` is not invoked here. Is this service available for everyone?

goreck888: Yes, it's public.

PR #1369, opened 2020-01-31
JS and SCSS customization

mkasztelnik:
Views, JS and SCSS can be customized by adding `CUSTOMIZATION_PATH` env variable.
This variable should point to the directory with the following structure

```
/my/customization/dif
  - views
  - javascript
```

Every file in these directories will override file from `app/views` or
`app/javascript`.

**Warning**: for the JS and SCSS when a new overriding file is added or removed to
customization directory rails application needs to be restarted.


mkasztelnik:
Closing in favour of #1390

PR #1370, opened 2020-01-31
Lead Section and item

martaswiatkowska:


app/controllers/home_controller.rb
```diff
@@ -4,6 +4,8 @@ class HomeController < ApplicationController
   before_action :load_services, :load_platforms, :load_providers, :load_target_groups, :load_opinion
 
   def index
+    @learn_more_section = LeadSection.find_by(slug: "learn-more")
+    @use_cases_section = LeadSection.find_by(slug: "use-cases")
```
mkasztelnik: You should include leads in the query to avoid additional 2 queries.

db/migrate/20200123105814_create_lead_sections.rb
```diff
@@ -0,0 +1,11 @@
+class CreateLeadSections < ActiveRecord::Migration[6.0]
+  def change
+    create_table :lead_sections do |t|
+      t.string :slug, unique: true
```
mkasztelnik: `null: false` ?

db/migrate/20200123125921_create_lead_items.rb
```diff
@@ -0,0 +1,13 @@
+class CreateLeadItems < ActiveRecord::Migration[6.0]
+  def change
+    create_table :lead_items do |t|
+      t.string :header, null: false
+      t.string :body, null: false
+      t.string :url
```
mkasztelnik: `null: false`?

app/models/lead_item.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class LeadItem < ApplicationRecord
```
mkasztelnik: Maybe simple `Lead` ?

PR #1371, opened 2020-01-31
[#1342] Unification of modal templates

### LINKED ISSUE #1342: Change modal loading
Now we have 3 modals in the mp:
- modal for creating projects by service ordering
- modal for question in services view
- modal for issue report to the xGUS

All of these use different templates for generating modals, and we have generic wizard for controllers which uses one template.

There is a need to refactor all these modals to make every dependent of this one modal template `view/layouts/_modal.html.haml`

goreck888:
Change the way for rendering modals in mp.
Now for all modals we use one template
Fixes #1342


app/controllers/services/questions_controller.rb
```diff
@@ -33,12 +34,13 @@ def create
...
     def render_modal_form
-      render "services/question_modal",
+      render "layouts/show_modal",
              content_type: "text/javascript",
              locals: {
                 title: "Ask provider",
                 action_btn: t("simple_form.labels.question.new"),
-                form: "services/questions/form"
+                form: "services/questions/form",
+                variables: { service: @service, question: @question }
```
mkasztelnik: Maybe `form_locals` is a better name?

app/controllers/reports_controller.rb
```diff
@@ -1,6 +1,15 @@
 # frozen_string_literal: true
 
 class ReportsController < ApplicationController
+  def new
+    @report = Report.new
+
+    respond_to do |format|
+      format.html
```
mkasztelnik: Is `html` needed here?

app/controllers/services/questions_controller.rb
```diff
@@ -3,6 +3,7 @@
 class Services::QuestionsController < ApplicationController
   def new
     @question = Service::Question.new
+    @service = Service.friendly.find(params[:service_id])
 
     respond_to do |format|
       format.html
```
mkasztelnik: Is there a situation when question form is rendered as html?

app/views/reports/new.html.haml
```diff
@@ -0,0 +1,14 @@
+.container
+  .mb-4.pb-4.border-header
+    %h1.mr-auto.my-auto Ask provider
+
+  = render "reports/form", remote: true, report: @report
```
mkasztelnik: It looks like the `remote` variable is not used in the `reports/form` partial.

app/views/reports/new.html.haml
```diff
@@ -0,0 +1,14 @@
+.container
```
mkasztelnik: It this view used anywhere in the application? I do not reckon a place in the marketplace were report is rendered as normal html (only js).

app/views/services/questions/new.html.haml
```diff
@@ -0,0 +1,14 @@
+.container
+  .mb-4.pb-4.border-header
+    %h1.mr-auto.my-auto Ask provider
+
+  = render "services/questions/form", remote: true, question: @question, service: @service
```
mkasztelnik: `remote` variable is not used in `services/questions/form`

app/views/services/questions/new.html.haml
```diff
@@ -0,0 +1,14 @@
+.container
```
mkasztelnik: Similar to the above: I do not reckon a place in MP where the question form is rendered as html (only js).

app/controllers/reports_controller.rb
```diff
@@ -1,6 +1,15 @@
...
+      format.js { render_modal_form }
+    end
+  end
+
   def create
```
mkasztelnik: If report is created only using `js` this method can also be cleaned a bit and all `html` elements can be removed.

app/views/projects/_form.html.haml
```diff
@@ -1,4 +1,4 @@
-= simple_form_for @project, remote: @show_as_modal,
+= simple_form_for project, remote: show_as_modal,
```
mkasztelnik: I propose to introduce one small change here:

```ruby
#change 
remote: show_as_modal
#to 
local_assigns[:show_as_modal]
```

This small change will allow us to have default `show_as_modal` value set to `false`. As a conclusion you will be able to remove it from the normal `projects/{new/edit}.htm.haml` partials.

app/controllers/reports_controller.rb
mkasztelnik: This is also valid for questions controller.

app/controllers/services/questions_controller.rb
goreck888: No problem, I just saved previous naming :)

app/controllers/services/questions_controller.rb
goreck888: No, I delete it.

app/views/reports/new.html.haml
goreck888: No it isn't used, I delete this file.

app/views/services/questions/new.html.haml
goreck888: File deleted.

app/views/services/questions/new.html.haml
goreck888: File deleted.

app/controllers/reports_controller.rb
goreck888: Done.

app/views/projects/_form.html.haml
goreck888: Done.

app/views/reports/new.html.haml
goreck888: File deleted.

app/controllers/reports_controller.rb
goreck888: No, I delete it.

agpul:
I have a question. Is it intentional that on production the template looks nice (graphic, fonts etc.) and in this PR we've lost this nice view?
Production: 
image
PR: 
image

mkasztelnik:
Nope, this is a bug. Most probably CSS was created to specific and should be improved. @goreck888 can you take a look at this issue?

PR #1380, opened 2020-02-10
Display "lead-error" only to admin user

martaswiatkowska:
Displayed lead-error (Cannot find lead_section with slug) for admin user. Plain user cannot see this message.

### DIFF
--- a/app/controllers/admin/lead_sections_controller.rb
+++ b/app/controllers/admin/lead_sections_controller.rb
@@ -4,7 +4,7 @@ class Admin::LeadSectionsController < Admin::ApplicationController
   before_action :find_and_authorize, only: [:edit, :update, :destroy]
 
   def new
-    @lead_section = LeadSection.new
+    @lead_section = LeadSection.new(slug: params[:slug])
     authorize(@lead_section)
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Added
 - Customization for translation files (@goreck888)
 - Customization for scss files (@michal-szostak, @mkasztelnik)
+- Lead section error box displayed to admin user (@martaswiatkowska)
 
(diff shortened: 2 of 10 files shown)


martaswiatkowska:
He cannot. But only admin see this message.

martaswiatkowska:
You must create section with e "slug"  learn-more or use-cases.
image

agpul:
ok, I can see it as the unlogged user. So point number 1 is unrelevant

agpul:
OK, so we need to improve this and add some tips for admin what should be in Slug or maybe we should consider adding a list with names to choose?

MinLidia:
Can't load a picture to use cases section. Tried to load to Learn more and works for me. 

Screenshot 2020-02-12 at 09 38 50

martaswiatkowska:
yes because we waiting for front. For now picture is a part of front

MinLidia:
ok, I see. thanks!

martaswiatkowska:
I don't know what to write as note because for now whiteabels are in progress. In final version can be many special slugs like "learn-more" or "use-cases" or admin will be able to edit these slugs.  For now i can add hint ```"You can add sections with slugs: "learn-more" or "use-cases" to see them on home page"```

martaswiatkowska:
And I think will be created doc about all customisation for final version

app/controllers/admin/lead_sections_controller.rb
```diff
@@ -4,7 +4,7 @@ class Admin::LeadSectionsController < Admin::ApplicationController
   before_action :find_and_authorize, only: [:edit, :update, :destroy]
 
   def new
-    @lead_section = LeadSection.new
+    @lead_section = LeadSection.new(slug: params[:slug])
```
goreck888: Why you added slug parameter for new section?

martaswiatkowska: To fill slug input When you go to new lead_section form from home page

PR #1396, opened 2020-02-18
[#945] automatic offer creating

### LINKED ISSUE #945: Improve the way of adding services without offer in Backoffice
Now, when I try to add the service in Backoffice and this service has no offer I need to add a fake offer. This fake offer is necessary to have "Add to a project" button on the front page of Marketplace.

We would like to apply the temporary solution agreed during spring preparation discussion :  

1. We should create a default offer for imported services from Infracentr
It should have :
-  Name " Offer" 
- Description " Service name Offer" 
- Open access type 
- URl from Service ( Webpage url)

2.  For the services that don't have offers, to show the alert massage to fill the offer details.  Preferably on the top of the page - to be agreed with Andrzej. 

**Additional content**:

goreck888:
Add automatic offer creating by service import from external sources.

Add warnings on the backoffice services view and backoffice service view about offers statuses.
Fixes #945

### DIFF
--- a/lib/import/eic.rb
+++ b/lib/import/eic.rb
@@ -192,6 +192,7 @@ module Import
               if @default_upstream == :eic
                 service.update(upstream_id: service_source.id)
               end
+              create_default_offer!(service, name, eid, url)
             end
           else
             service = Service.find_by(id: service_source.service_id)
@@ -200,6 +201,7 @@ module Import
               log "Updating [EXISTING] service #{service.title}, id: #{service_source.id}, eid: #{eid}"
               unless @dry_run
                 service.update!(updated_service_data.except(:research_areas, :categories, :status))
+                create_default_offer!(service, name, eid, url)
               end
             else
               not_modified += 1
@@ -219,6 +221,20 @@ module Import
           file << JSON.pretty_generate(output)
         end
       end
+    end
+
+    def create_default_offer!(service, name, eid, url)
+      if service&.offers.blank? && !url.blank?
+        log "Adding [NEW] default offer for service: #{name}, eid: #{eid}"
+        Offer.create!(name: "Offer", description: "#{name} Offer", offer_type: "open_access",
+                      webpage: url, status: service.status, service: service)
+      elsif url.blank?
+        log "[WARNING] Offer cannot be created, because url is empty"
+      end
+    rescue ActiveRecord::RecordInvalid => reason
+      log "ERROR - Default offer for #{service.title} (eid: #{eid}) cannot be created. #{reason}"
+    rescue error
+      log "ERROR - Default offer for #{service.title} (eid: #{eid}) cannot be created. Unexpected #{error}!"
     end
 
     def map_category(category)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Please view this file on the master branch, on stable branches it's out of date.
 ## [Unreleased]
 
 ### Added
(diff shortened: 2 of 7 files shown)


goreck888:
Warnings views:
no offers EOSC Marketplace
no published offers EOSC Marketplace
no offers services EOSC Marketplace
no published offers services EOSC Marketplace

agpul:
My first comment before checking further. Change the warnings to:
The service has no offers. Add one offer to make possible for a user to Access the service.
The service is published but has no published offers. Publish one offer to make possible for a user to Access the service.

goreck888:
> My first comment before checking further. Change the warnings to:
> The service has no offers. Add one offer to make possible for a user to Access the service.
> The service is published but has no published offers. Publish one offer to make possible for a user to Access the service.

Done.

lib/import/eic.rb
```diff
@@ -192,6 +192,19 @@ def call
...
+              begin
+                unless url.blank?
+                  log "Adding [NEW] default offer for service: #{name}, eid: #{eid}"
+                  Offer.create!(name: "Offer", description: "#{name} Offer", offer_type: "open_access",
+                                webpage: url, status: "draft", service: service)
+                else
+                  log "[WARNING] Offer cannot be created, because url is empty"
+                end
+              rescue ActiveRecord::RecordInvalid => invalid
+                log "ERROR - #{invalid}! #{service.title} default offer (eid: #{eid}) will NOT be created"
```
mkasztelnik: Maybe `reason` is a better name for an `invalid` variable?

Maybe `Default offer (...) for #{service.title) cannot be created. #{resason.to_s}"

The difference between `#{reason}` and `#{reason.to_s}` is that the first one by default is using `inspect` method to show the details:

```
puts "#{reason}" #=> <ActiveRecord::RecordInvalid: Validation failed: ...
puts "#{reason.to_s} #=> Validation failed: ...
```

lib/import/eic.rb
```diff
@@ -192,6 +192,19 @@ def call
...
+              rescue error
+                log "ERROR - Unexpected #{error}! #{service.title} default offer (eid: #{eid}) will NOT be created!"
```
mkasztelnik: Similar to above

lib/import/eic.rb
```diff
@@ -192,6 +192,19 @@ def call
               if @default_upstream == :eic
                 service.update(upstream_id: service_source.id)
               end
+              begin
+                unless url.blank?
+                  log "Adding [NEW] default offer for service: #{name}, eid: #{eid}"
+                  Offer.create!(name: "Offer", description: "#{name} Offer", offer_type: "open_access",
+                                webpage: url, status: "draft", service: service)
```
mkasztelnik: Does it solve the original problem? The offer is created but it is in draft mode, so still, user will not be able to "order" this service.

@MinLidia can you confirm that `draft` status should be set here?

lib/import/eic.rb
```diff
@@ -192,6 +192,19 @@ def call
               if @default_upstream == :eic
                 service.update(upstream_id: service_source.id)
               end
+              begin
```
mkasztelnik: You should add spec for this feature.

lib/import/eic.rb
goreck888: i got Lint warning
```
W: Lint/RedundantStringCoercion: Redundant use of `Object#to_s` in interpolation.
```
after typing `reason.to_s` so I leave it as it is.

PR #1397, opened 2020-02-18
[#1316] Enhancement of the offer selection

### LINKED ISSUE #1316: Enhancement of the offer selection
**Context**

While selecting the offer, user might not know which one he has selected. Now it's just underlined by the color. 

**Expected behaviour** : 

User should know which offer he has selected. 

**Proposed solution** : 

* To change the text on the button to " Selected offer" --> we choose this one
* To delete the text and leave only arrow

goreck888:
Change text content by offer selection from
'Select an offer' to 'Selected offer'
Fixes #1316

### DIFF
--- a/app/views/services/offers/_offer.html.haml
+++ b/app/views/services/offers/_offer.html.haml
@@ -7,5 +7,8 @@
       %label
         = f.radio_button :offer_id, offer.iid, checked: project_item.offer == offer
         %span.btn.btn-light.border.border-secondary.font-weight-bold
-          = t("services.offer.offer_select")
+          .unchecked
+            = t("services.offer.offer_select")
+          .checked
+            = t("services.offer.offer_selected")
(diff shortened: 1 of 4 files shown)


app/views/services/offers/_offer.html.haml
```diff
@@ -7,5 +7,8 @@
       %label
         = f.radio_button :offer_id, offer.iid, checked: project_item.offer == offer
         %span.btn.btn-light.border.border-secondary.font-weight-bold
-          = t("service.offer_select")
+          #unchecked
```
mkasztelnik: We can have more than 1 offer. As a result, you are creating many HTML elements with the same ID. ID should be unique (https://www.w3schools.com/tags/att_global_id.asp). Please use class instead of ID.

goreck888: Ok.

PR #1398, opened 2020-02-19
[#1099] Dynamic breadcrumbs in the service view

### LINKED ISSUE #1099: Breadcrumbs logic to be updated

The current implementation of the breadcrumbs functionality, which should provide links back to the previous page the user navigated before are wrongly implemented ( because of this ambiguity of double presence of the service in both categories).
Example:
1. Go to: https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services
2. find in the search bar: 100 Percent IT Trusted Cloud and go to it
3. See breadcrumb: 

1. Go to: https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services
2. Go to Storage Category: https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services/c/storage
3. Choose from the list: 100 Percent IT Trusted Cloud
4. See breadcrumb: 

We will improve the breadcrumbs functionality. Established improvements:
- [x] If the user comes through the category list, he will see the category from where he comes from (even if it is not the main category of this service).
- [x] If the user comes from a search bar, he will see the services and the main category.

goreck888:
Fix category parent in the breadcrumbs in the service view
Fixes #1099

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -10,6 +10,7 @@ module Service::Searchable
     before_action only: :index do
       @filters = visible_filters
       @active_filters = active_filters
+      session[:category] = @category&.id
     end
   end
 
--- a/config/breadcrumbs/marketplace.rb
+++ b/config/breadcrumbs/marketplace.rb
@@ -16,7 +16,9 @@ end
 
 crumb :service do |service|
   link service.title, service_path(service)
-  if service.main_category
+  if !session[:category].blank?
+    parent :category, Category.find(session[:category])
+  elsif service.main_category
     parent :category, service.main_category
   else
     parent :services
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Breadcrumbs depending on chosen category in service view (@goreck888)
 
 ### Security
 


app/controllers/concerns/service/searchable.rb
```diff
@@ -10,6 +10,7 @@ module Service::Searchable
     before_action only: :index do
       @filters = visible_filters
       @active_filters = active_filters
+      session[:category] = @category
```
mkasztelnik: You should store in DB as small as possible. Here you are storing the whole object.

config/breadcrumbs/marketplace.rb
```diff
@@ -16,8 +16,9 @@
 
 crumb :service do |service|
   link service.title, service_path(service)
-  if service.main_category
-    parent :category, service.main_category
+  if service.main_category || session[:category]
+    parent :category, session[:category].blank? ? service.main_category : session[:category]
+    session.delete(:category)
```
mkasztelnik: This solution is simple but I'm not sure if good enough since this is working only for the first time. After you hit refresh on the service view default category will be shown. 

@MinLidia, @agpul please let us know if this is good enough or should we search for the more advanced mechanism?

config/breadcrumbs/marketplace.rb
```diff
@@ -16,8 +16,9 @@
 
 crumb :service do |service|
   link service.title, service_path(service)
-  if service.main_category
-    parent :category, service.main_category
+  if service.main_category || session[:category]
```
mkasztelnik: You should revert order here. Now there will be always DB query for `main_category`. If condition would look like that `session[:category] || service.main_category` then the DB query will be done only when the category will not be present in the session.

mkasztelnik:
The other solution to solve this issue is to generate service url with category id attached. Something like: `service_path(category, from: category.id)`

config/breadcrumbs/marketplace.rb
agpul: ok w8 I've added my comment before reading that. It would be nice it the refresh doesn't change the breadcrumb behavior.

goreck888: Done.

goreck888:
> from the "clickable" point of view this functionality works as we agreed, so it is ok for me, but for me it would be nice if the refresh doesn't change the breadcrumbs.

Now it remembers which last category was chosen, so page refreshing doesn't affect on breadcrumbs.

PR #1399, opened 2020-02-19
[#914] Service breadcrumbs allow filters settings

### LINKED ISSUE #914: Breadcrumbs should not clear the filters.
This issue has been discovered when analising the new filtering capabilities (in PR https://github.com/cyfronet-fid/marketplace/pull/846) it needs further discussions in brader scope but it seems that navigation through breadcrumbs should not clear the filters or search. for this purpose we have clear filters functionality (introduced in the first round of filtering enhancements) and clear search funcionality (introduced in PR-846).

This is not a complete ticket specification.

goreck888:
Back to the services view from service using breadcrumbs
keeps filters and query settings
Fixes #914

### DIFF
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -10,6 +10,7 @@ module Service::Searchable
     before_action only: :index do
       @filters = visible_filters
       @active_filters = active_filters
+      store_query_params
     end
   end
 
@@ -93,6 +94,16 @@ module Service::Searchable
 
     def active_filters
       @active_filters ||= all_filters.flat_map { |f| f.active_filters }
+    end
+
+    def store_query_params
+      session[:query] = {}
+      @filters.each do |filter|
+        session[:query][filter.field_name] = params[filter.field_name] unless params[filter.field_name].blank?
+        session[:query]["#{filter.field_name}-all"] =
+            params["#{filter.field_name}-all"] unless params["#{filter.field_name}-all"].blank?
+      end
+      session[:query][:q] = params[:q] unless params[:q].blank?
     end
 
     def filter_classes
--- a/app/views/services/nav/_categories.html.haml
+++ b/app/views/services/nav/_categories.html.haml
@@ -9,7 +9,7 @@
   - view = content_for(:category_view) || "services/nav/category"
   - categories.each do |cid, category|
     = render view, category: category[:category], current: current, counter: category[:counter]
-    - if current == category[:category] && !subcategories.empty?
+    - if current == category[:category] && !subcategories.blank?
       %ul.ml-3
         - subcategories.each do |sid, subcategory|
           = render view, category: subcategory[:category], current: current, counter: subcategory[:counter]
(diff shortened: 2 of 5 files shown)


app/controllers/concerns/service/searchable.rb
```diff
@@ -10,6 +10,8 @@ module Service::Searchable
     before_action only: :index do
       @filters = visible_filters
       @active_filters = active_filters
+      session[:active_filters] = @active_filters
```
mkasztelnik: Why do we need to store this value? I'm asking because as far as I understand the only issue here to solve is to construct services link in breadcrumb with all query params already populated. Basing on this data view should be rendered just fine with active filters calculated.

app/controllers/concerns/service/searchable.rb
```diff
@@ -95,6 +97,20 @@ def active_filters
       @active_filters ||= all_filters.flat_map { |f| f.active_filters }
     end
 
+    def store_query_params
+      session[:query] = { "research_areas-filter"=> params["research_areas-filter"],
```
mkasztelnik: I propose to use `@filters` list to generated dynamically this hash. From each filter, you can grab the filter key and use it to fetch values from `param`. This has this plus that when a new filter will be added this code will not need any changes.

mkasztelnik: What is more I propose to set some timeout for this session value as well? Maybe to 5 minutes?

app/views/services/nav/_categories.html.haml
```diff
@@ -9,7 +9,7 @@
   - view = content_for(:category_view) || "services/nav/category"
   - categories.each do |cid, category|
     = render view, category: category[:category], current: current, counter: category[:counter]
-    - if current == category[:category] && !subcategories.empty?
+    - if current == category[:category] && !subcategories.blank?
```
mkasztelnik: How this change is connected with this PR?

goreck888: After breadcrumbs changes I had `UndefinedMethodError empty?`

app/controllers/concerns/service/searchable.rb
goreck888: I think setting timeout is not necessary, every operations on filters are stored to the session and used only with breadcrumbs, so it's not possible to go to the previous filter state without using browser back button. Other thing now I use `@filters`.

app/controllers/concerns/service/searchable.rb
```diff
@@ -95,6 +96,14 @@ def active_filters
       @active_filters ||= all_filters.flat_map { |f| f.active_filters }
     end
 
+    def store_query_params
+      session[:query] = {}
```
goreck888: This is necessary for tests.
Ruby recognize it's hash but in tests it treat `session[:query]` as a NilClass.

app/controllers/concerns/service/searchable.rb
```diff
@@ -95,6 +96,14 @@ def active_filters
...
+      @filters.each do |filter|
+        session[:query][filter.field_name] = params[filter.field_name]
```
mkasztelnik: You forget about `*-all` fields, which are used to store the filter query string. What is more, you can improve it a bit by storing only not blank values.

PR #1413, opened 2020-02-24
Fix offence detected by rubocop

goreck888:
Fix new offense detected by rubocop in searchable.rb
Needed for the mp dependencies update


martaswiatkowska:
This shouldn't go with gem update?

PR #1414, opened 2020-02-24
Update rubocop

goreck888:
Update rubocop
Fix new offense detected by the new version of rubocop


goreck888:
> This shouldn't go with #1410 this pr?

After consultation with @mkasztelnik he proposed to split rubocop update with other updates and add fix only to this update.

PR #1417, opened 2020-02-25
Fix modal routes

goreck888:
Add redirects for report and question modal,
which don't have pure html view:
- root path for report modal
- service path for question

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Redirects for modals which don't have html view (@goreck888)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


goreck888:
Test:

Try to get `root_path/reports/new` and `service_path/question/new` and see if you get an error.

goreck888:
> I'm not 100% sure if this is needed. Why the user will enter this webpage? Maybe instead of this, we should allow only JS in routes for these paths? Something like:
> 
> ```ruby
> resources :reports, constraints: { format: /(js|json)/ }
> ```

I don't have idea, why user will enter this webpage, but we have 20 events users try to get report or question route via html.
Proposed solution doesn't work - it show the same error like without constraints.
The only thing we can do is render 404 or redirect as in this PR.

PR #1418, opened 2020-02-25
[#1372] Fix displaying messages and statuses in projects

### LINKED ISSUE #1372: Change the order of the message in the Service tab and Project tab

There is a problem to find the latest conversation in the timeline. You need to scroll down to see the latest news. It is confusing from the user perspective. ( based on feedback from the real users gathered on the conference) 

For both : Project and Service view : 

1. In the order History the logic of the order should be changed. We need to show the new conversation first.
2. In the conversation Tab we should scroll  down to the newest line as it is in the Facebook

goreck888:
Reverse order of project and services statuses
Scroll always to the last message in messages
Fixes #1372

### DIFF
--- a/app/views/messages/_messages.html.haml
+++ b/app/views/messages/_messages.html.haml
@@ -1,4 +1,4 @@
-#scrollbar.progress-box.mt-4
+#scrollbar.progress-box.mt-4{ "data-controller": "scroller" }
   %ul.step-progress.pl-5.pb-4
     - if messages.count.positive?
       - messages.each do |message|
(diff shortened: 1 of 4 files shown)


app/views/messages/_messages.html.haml
```diff
@@ -9,7 +9,12 @@
         Conversation not started yet.
 
 :javascript
-  window.onload = function () {
+  function scrollToLastMessage() {
```
mkasztelnik: Can we convert this into stimulus controller and remove JS from view code?

goreck888: Yes :)

PR #1420, opened 2020-02-26
Offers in autocomplete

martaswiatkowska:
Added offers to autocomplete results.
After click on displayed offers users are redirected to offer on service page by anchor.
Only published offers are indexed and if the service has more than two offers.

### DIFF
--- a/app/controllers/concerns/service/autocomplete.rb
+++ b/app/controllers/concerns/service/autocomplete.rb
@@ -4,29 +4,23 @@ module Service::Autocomplete
   extend ActiveSupport::Concern
 
   def autocomplete
-    query = Service.search(params[:q],
-      fields: ["title"],
-      operator: "or",
-      match: :word_middle,
-      limit: 5,
-      load: false,
-      where: { id: scope.ids },
-      highlight: { tag: "<b>" }
-    )
+    query = Searchkick.search(params[:q],
+                           fields: ["name", "title"],
+                           operator: "or",
+                           match: :word_middle,
+                           limit: 5,
+                           load: false,
+                           where: { service_id: scope.ids },
+                           highlight: { multiple: true, tag: "<b>" },
+                           models: [Service, Offer],
+                           model_includes: { Service => [:offers], Offer => [:service] })
+
+    service_titles = Service.where(id: scope.ids).pluck(:id, :title).to_h
 
     if request.xhr?
-      generate_html(query)
+      render(template: "services/autocomplete/list",
+             locals: { results: query.with_highlights, service_titles: service_titles },
+             layout: false)
     end
   end
-
-  private
-    def generate_html(query)
-      html_results = query.with_highlights.map { |s, h| "<li class=\"dropdown-item\" role=\"option\" data-autocomplete-value=\"#{s[:id]}\">#{h[:title]}</li>" }.join
-
-      respond_to do |format|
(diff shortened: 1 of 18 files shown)


app/controllers/concerns/service/autocomplete.rb
```diff
@@ -21,12 +23,33 @@ def autocomplete
 
   private
     def generate_html(query)
-      html_results = query.with_highlights.map { |s, h| "<li class=\"dropdown-item\" role=\"option\" data-autocomplete-value=\"#{s[:id]}\">#{h[:title]}</li>" }.join
+      service_results = "<p class=\"list-header\"> Services </p>"
```
mkasztelnik: This generation starts to be complex. Maybe a good solution would be to move HTML generation to `haml`?

app/models/offer.rb
```diff
@@ -3,11 +3,27 @@
...
     published: "published",
     draft: "draft"
   }
 
+  def search_data
+    {
+      offer_id: id,
+      name: name,
+      description: description,
+      service_id: service_id
```
mkasztelnik: Can we add a published flag to elasticsearch and use it to limit offers found in MP (autocomplete in backoffice should search in all offers).

app/helpers/service_helper.rb
```diff
@@ -59,4 +59,8 @@ def service_logo(service)
       image_tag "eosc-img.png"
     end
   end
+
+  def get_service_title(id)
+    Service.find(id).title
```
mkasztelnik: This is quite a danger and can lead to `n+1` queries problem. Can we use a mechanism similar to service search `scope_results` to have offers already joined with services?

app/views/services/autocomplete/_offer_item.html.haml
```diff
@@ -0,0 +1,6 @@
+%li.dropdown-item{ "role": "option",
+  "data-autocomplete-value": "#{offer.service_id}",
+  "data-autocomplete-anchor": "#{offer.id}" }
+  %span= get_service_title(offer.service_id)
```
mkasztelnik: `span` should be used only for the service title of for the whole line?

mkasztelnik: Can we use joins in the offers result and `offer.service.title`?

app/models/offer.rb
martaswiatkowska: Function should_index? are filtering published offers https://github.com/cyfronet-fid/marketplace/pull/1420/files/e5d87a8fb986d2c019adde45f32c1be5a3fb4c44#diff-a1928e578416256fe9dc55132da41a2bR23

Now I think we don't need offer_id and filtering by offer_id in request to elasticsearch.

app/views/services/autocomplete/_offer_item.html.haml
martaswiatkowska: Only for the service title because its element with different collor

martaswiatkowska:
@wziajka doesn't have to do sth additional because already all models with searchkick are indexed by ```rake searchkick:reindex:all```

app/models/offer.rb
mkasztelnik: Ok, it is good for now. When the offer search will be added to the backoffice we will change this mechanism and only then add this published flag.

Issue #1404, opened 2020-02-21
Add the services comparision functionality


agpul:
**Is your feature request related to a problem? Please describe.**
As in the catalogue (eIC) has the possibility to compare the services we are obligated to add something similar in Marketplace. 

**Describe the solution you'd like**
We would like to compare the services by:
- service access type
- places
- languages
- service phase
- providers
- research area
- dedicated for
- category (??) or maybe we would like to compare the services only within the given category
- version
- tags

Additional features:
- [x] we would like to have the possibility to compare up to 3  services simultaneously
- [x] we would like to have the possibility to Remove the services from the comparison engine (see the Reference link)
- [x] we would like to have the possibility to add to the comparison engine additional service to compare (for eg. if we have 2 services selected we can add another one from the comparison view)

**Describe alternatives you've considered**

**Additional context**
Reference: https://catalogue.eosc-portal.eu/compare;services=West-Life.pdb_redo_server,West-Life.fanten

**Scenario + graphic previews**
User can add service to compare in two ways:
- clicking "add to compare" on services list https://invis.io/KSW7S7BNG2R#/407379557_Item_List_Search_-_Compare_Box
- clicking "add to compare" on service detail https://invis.io/DSP0QZN5ZX3#/330285958_Item-Detail

After that, "compare bar" is shown. 
Grom "compare bar" user can:
- remove one or more element, 
- go to compare page by clicking "compare" button,
- go to added service (clicking on it).

Depends on how many services (max 3) user adds, the right view is shown.
- https://invis.io/79W7WS6CK4H#/407417182_Compare_Page_1_Item
- https://invis.io/NWW7WSR7RSG#/407417183_Compare_Page_2_Items
- https://invis.io/KNW7S7LCMRH#/407379556_Compare_Page
from this view, user can remove service or add new service by clicking "add next service" (link to services list).

mkasztelnik:
I would like to ask how user will end up in a compare webpage? Will we have some add to compare button on every service view, store this in the user session and then use this data to run compare view? Another possibility is to have some checkboxes on the services list. 

To sum up, it would be nice to have a clean story starting from the beginning, so how elements to compare are chosen and how to enter into compare views. Some propositions from @abacz would be nice.

MinLidia:
Please also add limitation text after user added 3 services and want to add the 4th. Example below : 

[...]

PR #1431, opened 2020-03-13
[#1404] Add services comparison

goreck888:
Add controller for comparison handling

### DIFF
--- a/app/controllers/comparisons/services_controller.rb
+++ b/app/controllers/comparisons/services_controller.rb
@@ -6,7 +6,7 @@ class Comparisons::ServicesController < ComparisonsController
     session[:comparison] = session[:comparison].blank? ? [] : session[:comparison]
     session[:comparison].include?(new_service_slug) ?
         session[:comparison].delete(new_service_slug) : session[:comparison] << new_service_slug
-    @compare_services = Service.where(slug: session[:comparison])
+    @services = Service.where(slug: session[:comparison])
     respond_to do |format|
       format.json { render_json }
     end
--- a/app/controllers/comparisons/services_controller.rb
+++ b/app/controllers/comparisons/services_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class Comparisons::ServicesController < ComparisonsController
+  def create
+    new_service_slug = params.fetch(:comparison)
+    session[:comparison] = session[:comparison].blank? ? [] : session[:comparison]
+    session[:comparison].include?(new_service_slug) ?
+        session[:comparison].delete(new_service_slug) : session[:comparison] << new_service_slug
+    @compare_services = Service.where(slug: session[:comparison])
+    respond_to do |format|
+      format.json { render_json }
+    end
+  end
+
+  def destroy
+    session[:comparison].delete(params[:slug])
+    redirect_to comparisons_path(params[:fromc].present? ? { fromc: params[:fromc] } : nil)
+  end
+end
--- a/app/controllers/comparisons_controller.rb
+++ b/app/controllers/comparisons_controller.rb
@@ -2,8 +2,15 @@
 
 class ComparisonsController < ApplicationController
   def show
-    @compare_services = Service.where(slug: session[:comparison])
-    redirect_to_services_path
+    @services = Service.where(slug: session[:comparison])
+    if @services.blank?
+      if params[:fromc]
+        category = Category.find_by(slug: params[:fromc])
+        redirect_to category_services_path(category)
+      else
+        redirect_to services_path
+      end
+    end
   end
 
   def destroy
@@ -15,22 +22,11 @@ class ComparisonsController < ApplicationController
   end
 
   private
-    def redirect_to_services_path
-      if @compare_services.blank?
-        if params[:fromc]
-          category = Category.find_by(slug: params[:fromc])
-          redirect_to category_services_path(category)
-        else
-          redirect_to services_path
-        end
-      end
-    end
-
     def render_json
-      render json: { data: session[:comparison], html: render_bottom_bar }
+      render json: { data: session[:comparison], html: bottom_bar }
     end
 
-    def render_bottom_bar
-      render_to_string(partial: "comparisons/service", collection: @compare_services, formats: [:html])
+    def bottom_bar
+      render_to_string(partial: "comparisons/service", collection: @services, formats: [:html])
     end
 end
--- a/app/controllers/comparisons_controller.rb
+++ b/app/controllers/comparisons_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class ComparisonsController < ApplicationController
+  def show
+    @compare_services = Service.where(slug: session[:comparison])
+    redirect_to_services_path
+  end
+
+  def destroy
+    session[:comparison] = []
+    respond_to do |format|
+      format.html { redirect_to services_path }
+      format.js { render_json }
+    end
+  end
+
+  private
+    def redirect_to_services_path
+      if @compare_services.blank?
+        if params[:fromc]
+          category = Category.find_by(slug: params[:fromc])
+          redirect_to category_services_path(category)
+        else
+          redirect_to services_path
+        end
+      end
+    end
+
+    def render_json
+      render json: { data: session[:comparison], html: render_bottom_bar }
+    end
+
+    def render_bottom_bar
+      render_to_string(partial: "comparisons/service", collection: @compare_services, formats: [:html])
+    end
+end
--- a/app/views/comparisons/_bar.html.haml
+++ b/app/views/comparisons/_bar.html.haml
@@ -1,12 +1,13 @@
 .list-inline.border.bg-light
   .row
-    .col-2
-    .col-7.row{ "data-target": "comparison.box" }
-      - unless services.blank? || current_page?(backoffice_services_path)
-        = render partial: "comparisons/service", collection: services
-    .col-3
-      = link_to "CLEAR ALL", comparisons_path, method: :delete, remote: true,
-        "data-action": "click->comparison#clearAll"
-      = link_to "Compare", comparisons_path(category.present? ? { fromc: category.slug } : nil),
-      class: "btn btn-primary"
+    .container
+      .row
+        .col-lg-9.col-12.row.services-list{ "data-target": "comparison.box" }
+          - unless services.blank? || current_page?(backoffice_services_path)
+            = render partial: "comparisons/service", collection: services
+        .col-lg-3.col-12.text-right.action-box
+          = link_to "CLEAR ALL", comparisons_path, method: :delete, remote: true,
+            "data-action": "click->comparison#clearAll"
+          = link_to "Compare", comparisons_path(category.present? ? { fromc: category.slug } : nil),
+          class: "btn btn-primary ml-3"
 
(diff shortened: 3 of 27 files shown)


goreck888:
> I have a couple of comments/questions:
> 
>     1. I have the message: "You have reached the maximum numbers of items you can compare", I've got this message after selecting the third service. I think that this message should be shown if I want to choose the fourth service, there should be no message if I have selected three services. What do you think? @abacz
> 
>     2. On the service access type, I would like to have something similar that we have on the service view. So if I have the cursor on "internal ordering" it would be nice to have a tip/ explanation

Ok, i can add tooltips, just need the content for it
> 
>     3. I removed 2 services from comparison view and now I have "Add new service" button doubled:
>        image

It's compatible with @abacz prevs

> 
>     4. I removed all the services from the comparison view, I think that now I should be redirected to the service search view. Look what we have now:
>        image

Now it redirects to the service or category view

>        5.**The most important** The comparison feature doesn't work when I am inside category, it works only on All service view. To reproduce:
> 
> 
>     * Go to search view
> 
>     * Choose for eg Compute category (https://marketplace-14.docker-fid.grid.cyf-kr.edu.pl/services/c/compute)
> 
>     * Click on " Add to compare" button

Fixed.

agpul:
@abacz can you add the comments to points 1 and 3? For me, point number 1 should look like I've described. Point number 3, I can agree on your solution but please confirm it.

config/routes.rb
```diff
@@ -22,6 +22,12 @@
       resources :opinions, only: :index
     end
   end
+
+  match "comparisons/update", to: "comparisons#update", via: :post
+  match "comparisons/destroy", to: "comparisons#destroy", via: :get
+  match "comparisons/show", to: "comparisons#show", via: :get
+  match "comparisons/remove_service", to: "comparisons#remove_service", via: :get
```
mkasztelnik: These routings have several issues:
  1. You should **never** use `get` to remove resources
  1. This is not RESTfull

Instead, consider the following routings:
```ruby
resource :compare, only: [:show, :destroy] do
  scope :compare do
    resource :services, only: [:create, :destroy]
  end
end
```

as a result you will have following routes:
````
GET    /compare (show compare view)
DELETE /compare (remove all services from compare)

POST  /compare/services (add service to compare)
DELETE /compare/services/:slug (remove service from compare)
```

app/views/services/_service_details.html.haml
```diff
@@ -1,9 +1,16 @@
+- checked = session[:comparison]&.include?(service.slug)
```
mkasztelnik: Such logic should be defined in the controller, not in view.

app/views/services/_service_details.html.haml
```diff
@@ -1,9 +1,16 @@
...
 .media.mb-3.service-box.shadow-sm
   .media-body.pt-4.pb-2
     .pl-4.pr-4.mb-3
       %h2.mb-3= yield
       %p.mb-3= highlighted_for(:tagline, service, highlights)
       = render "services/categorization", service: service
+      - unless current_page?(backoffice_services_path)
```
mkasztelnik: Such a check should be global (checked only once). A controller is a good place to put such logic.

app/views/services/_service_details.html.haml
```diff
@@ -1,9 +1,16 @@
...
+        %p.mb-3
+          = check_box_tag "comparison", service.slug, checked, id: "comparison-#{service.id}",
```
mkasztelnik: This is a remark for further improvement (and this is also a question to @abacz):
Can we use a simple link instead of a check? Something like "Add to compare" (with some nice color), "Remove from compare" (red?).

app/views/services/_index.html.haml
```diff
@@ -1,3 +1,4 @@
+- more_than_2_checked = (session[:comparison]&.size || 0) > 2
```
mkasztelnik: Logic should be defined in controller. Consider better naming, maybe `compare_enabled`

app/views/layouts/application.html.haml
```diff
@@ -10,7 +10,8 @@
   %body{ class: ("main" if current_page?("/")) }
     - if current_page?("/")
       .main-elements
-    = render "layouts/flash"
+    #flash.fixed-top
+      = render "layouts/flash"
```
mkasztelnik: I don't like the solution with red nav alert visible all the time when 3 services are added to compare. Maybe a better solution would be to hide checkboxes (or `Add to compare` links) when compare is full.

goreck888: Honestly I have no idea how resolve this issue. Now, as we can see after 3rd service in the list checkboxes are disabled. @MinLidia told me, that she want to know why these checkboxes are disabled. So I set this alert. @agpul told the alert should be visible after try to click 4th service and you suggest to hide checkboxes - as I understand without any alert? It's similar to disable checkboxes as it was.

app/views/services/_index.html.haml
goreck888: Ok

app/views/services/_service_details.html.haml
goreck888: Ok

goreck888:
> No changelog entry. Compare bottom navbar and compare view UI will be improved - right? I don't see tests for this feature.

UI will be improved, Jarek is waiting for end of my changes. I'll push tests after requested changes fix.

config/routes.rb
goreck888: Done

app/controllers/comparisons_controller.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class ComparisonsController < ApplicationController
+  def update
+    new_service_id = params.fetch(:comparison)
```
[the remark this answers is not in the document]
goreck888: This variable was called before I change it to the slug and I forget about change its name.

app/views/comparisons/_bar.html.haml
```diff
@@ -0,0 +1,10 @@
+.list-inline.border.bg-light
+  .row
+    .col-2
+    .col-7.row{ "data-target": "comparison.box" }
+      - unless services.blank? || current_page?(backoffice_services_path)
+        = render partial: "comparisons/service", collection: services
+    .col-3
+      = link_to "CLEAR ALL", comparisons_destroy_path, remote: true, "data-target": "comparison.clear"
```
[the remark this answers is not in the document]
goreck888: Done.

agpul:
image
I suggest adding a clickable title here? What do you think? @abacz

abacz:
> I suggest adding a clickable title here? What do you think? @abacz
I talked to @goreck888 yesterday. He will fix this.

app/controllers/comparisons/services_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class Comparisons::ServicesController < ComparisonsController
+  def create
+    new_service_slug = params.fetch(:comparison)
+    session[:comparison] = session[:comparison].blank? ? [] : session[:comparison]
+    session[:comparison].include?(new_service_slug) ?
+        session[:comparison].delete(new_service_slug) : session[:comparison] << new_service_slug
+    @compare_services = Service.where(slug: session[:comparison])
```
mkasztelnik: Maybe simple `@service` - you are already in the compare context (controller name) so it seems for me like a duplication.

app/controllers/comparisons_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class ComparisonsController < ApplicationController
+  def show
+    @compare_services = Service.where(slug: session[:comparison])
+    redirect_to_services_path
```
mkasztelnik: This looks like a magic method. It uses not so obvious class variable. I would move this method content here.

app/controllers/comparisons_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class ComparisonsController < ApplicationController
+  def show
+    @compare_services = Service.where(slug: session[:comparison])
```
mkasztelnik: Similar like above

app/controllers/comparisons_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+          redirect_to services_path
+        end
+      end
+    end
+
+    def render_json
+      render json: { data: session[:comparison], html: render_bottom_bar }
+    end
+
+    def render_bottom_bar
```
mkasztelnik: Maybe simple `bottom_bar` since this is a string payload.

app/views/comparisons/_bar.html.haml
```diff
@@ -0,0 +1,13 @@
+.list-inline.border.bg-light
+  .row
+    .container
+      .row
+        .col-lg-9.col-12.row.services-list{ "data-target": "comparison.box" }
+          - unless services.blank? || current_page?(backoffice_services_path)
+            = render partial: "comparisons/service", collection: services
+        .col-lg-3.col-12.text-right.action-box
+          = link_to "CLEAR ALL", comparisons_path, method: :delete, remote: true,
```
mkasztelnik: Upper letters should be done using css class ([`text-uppercase`](https://getbootstrap.com/docs/4.4/utilities/text/#text-transform)).

app/views/comparisons/show.html.haml
```diff
@@ -0,0 +1,67 @@
...
+          %td= t("services.about.sidebar.fields.phase.#{service.phase}")
+      %tr.lightgrey-row
+        %th Places
+        - @compare_services.each do |service|
+          %td= service.places
+      %tr
+        %th Languages
+        - @compare_services.each do |service|
+          %td= service.languages
+      -# %tr
```
mkasztelnik: Please remove the commented code.

app/views/comparisons/show.html.haml
```diff
@@ -0,0 +1,67 @@
+- i = 3 - @compare_services&.size
+- breadcrumb :comparison
+.container
+  %table.services-comparison.bordered.shadow-sm.table-responsive
+    %thead
+      %tr
+        %th{ colspan: 4 }
+          Comparison
+          %span.float-right.font-weight-light
+            = link_to "CLEAR ALL", comparisons_path, method: :delete
```
mkasztelnik: Please use CSS to create an upper case text.

mkasztelnik: Please check other places in this PR where CSS class can be used instead of manually writing upper case text.

app/controllers/comparisons_controller.rb
goreck888: Done.

app/controllers/comparisons_controller.rb
goreck888: Done.

app/views/comparisons/_bar.html.haml
goreck888: Done.

app/views/comparisons/show.html.haml
goreck888: Done.

app/views/comparisons/show.html.haml
goreck888: Done.

app/controllers/comparisons_controller.rb
goreck888: Done.

app/controllers/comparisons/services_controller.rb
goreck888: Done.

goreck888:
> Commit "Last fixes" say nothing, I think. maybe you can squash commits?

Done.

PR #1432, opened 2020-03-13
Fix eIC error handling

goreck888:
Add necessary assignment operators in rescue

### DIFF
--- a/lib/import/eic.rb
+++ b/lib/import/eic.rb
@@ -212,7 +212,7 @@ module Import
           end
         rescue ActiveRecord::RecordInvalid => invalid
           log "ERROR - #{invalid}! #{service.title} (eid: #{eid}) will NOT be created (please contact catalog manager)"
-        rescue error
+        rescue StandardError => error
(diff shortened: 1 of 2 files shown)


martaswiatkowska:
This shouldn't go to branch 2.x?

lib/import/eic.rb
```diff
@@ -210,7 +210,7 @@ def call
           end
         rescue ActiveRecord::RecordInvalid => invalid
           log "ERROR - #{invalid}! #{service.title} (eid: #{eid}) will NOT be created (please contact catalog manager)"
-        rescue error
+        rescue => error
```
mkasztelnik: Do we want to catch all errors? Maybe `StandardError` should be enough?

lib/import/eic.rb
```diff
@@ -233,7 +233,7 @@ def create_default_offer!(service, name, eid, url)
       end
     rescue ActiveRecord::RecordInvalid => reason
       log "ERROR - Default offer for #{service.title} (eid: #{eid}) cannot be created. #{reason}"
-    rescue error
+    rescue => error
```
mkasztelnik: Similar to above

PR #1433, opened 2020-03-13
[#1422] Fix displaying details in service description

### LINKED ISSUE #1422: Helpdesk email section is shown even when there is not email defined
**Describe the bug**
The marketplace shows the helpdesk email section even when there is not help desk email defined. After clicking on the link new window in my email client is opened with the empty address (not very useful). 

**To Reproduce**
1. Go to marketplace checkin service (https://marketplace.eosc-portal.eu/services/egi-check-in)
2. Click on `Helpdesk e-mail`

**Expected behavior**
This section is not shown when no helpdesk email is given in service description

goreck888:
Fix displaying empty links, in particular helpdesk e-mail
Fixes #1422

### DIFF
--- a/app/views/services/sidebar/_links.html.haml
+++ b/app/views/services/sidebar/_links.html.haml
@@ -1,5 +1,5 @@
 - fields.each do |field|
-  - if !!service.send(field)
+  - if service.send(field).present?
     %li.list-group-item
       %span
         - if field.end_with? "_url"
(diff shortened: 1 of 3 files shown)


app/views/services/sidebar/_links.html.haml
```diff
@@ -1,5 +1,5 @@
 - fields.each do |field|
-  - if !!service.send(field)
+  - unless service.send(field).blank?
```
mkasztelnik: Can we change it into a positive condition? Something like `if  service.send(field).present?`

app/views/services/sidebar/_text.html.haml
```diff
@@ -1,5 +1,5 @@
 - fields.each do |field|
-  - if !!service.send(field)
+  - unless service.send(field).blank?
```
mkasztelnik: Similar as above.

PR #1435, opened 2020-03-17
Added version to import eic and display last update date

### LINKED ISSUE #1403: Add service versioning functionality
In BackOffice we don't have information about the service version - when and by whom it was created, who modified the service etc. This is one of the "missing functionalities" vs eIC.

In Backoffice we want to add the versioning functionality. We want to store information:
- [ ] Service version
- [ ] Date of the latest update of the service (Name of this field: **Updated** format: date dd/mm/yyyy)

For the user, on the main service page :
- [ ] version according to: https://projects.invisionapp.com/share/DSP0QZN5ZX3#/screens/330285958_Item-Detail

**Additional context**

- [ ] take into consideration eIC importer

Graphic preview:
- [x] @abacz will provide the place where this information should be shown #1426

martaswiatkowska:
Added import of version from eic. and displaying updatet_at as "last update" 

Fix #1403

### DIFF
--- a/app/views/backoffice/services/show.html.haml
+++ b/app/views/backoffice/services/show.html.haml
@@ -110,3 +110,9 @@
           - if @service.upstream_id == source.id
             %i.fas.fa-check
           = source
+
+      %h5 Version
+      %p= @service&.version
+
+      %h5 Last update
+      %p= l @service.updated_at, format: :short
--- a/app/views/services/sidebar/_date.html.haml
+++ b/app/views/services/sidebar/_date.html.haml
@@ -1,0 +1,4 @@
+- fields.each do |field|
+  - if service.send(field).present?
+    %li.list-group-item
+      %span= l service.send(field), format: :short
--- a/app/views/services/sidebar/_plain_text.html.haml
+++ b/app/views/services/sidebar/_plain_text.html.haml
@@ -1,0 +1,4 @@
+- fields.each do |field|
+  - if service.send(field).present?
+    %li.list-group-item
+      %span= service.send(field)
--- a/lib/import/eic.rb
+++ b/lib/import/eic.rb
@@ -104,6 +104,7 @@ module Import
         # subcategory_name = service["subCategoryName"]
         phase = service["trl"]
         provider_eid = service["providers"][0]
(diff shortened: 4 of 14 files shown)


agpul:
Is there any service with VERSION?

app/views/services/sidebar/_date.html.haml
```diff
@@ -0,0 +1,4 @@
+- fields.each do |field|
+  - if !!service.send(field)
```
goreck888: Change it to service.send(field).present? as in my PR fixing showing empty elements, because `!!` doesn't work as expected by send method.

app/views/services/sidebar/_plain_text.html.haml
```diff
@@ -0,0 +1,4 @@
+- fields.each do |field|
+  - if !!service.send(field)
```
goreck888: Similar as above.

app/views/services/sidebar/_date.html.haml
```diff
@@ -0,0 +1,4 @@
+- fields.each do |field|
+  - if service.send(field).present?
+    %li.list-group-item
+      %span= service.send(field).strftime("%D")
```
mkasztelnik: Try to use `I18n.l` here.

lib/import/eic.rb
```diff
@@ -175,7 +176,8 @@ def call
             status: "draft",
             providers: [mapped_provider],
             categories: map_category(category),
-            research_areas: [@research_area_other]
+            research_areas: [@research_area_other],
+            version: version || ""
```
mkasztelnik: Why not `nil` (I'm asking because `nil` is used as default in the service factory)?

app/views/backoffice/services/show.html.haml
```diff
@@ -110,3 +110,9 @@
           - if @service.upstream_id == source.id
             %i.fas.fa-check
           = source
+
+      %h5 Version
+      %p= @service&.version
+
+      %h5 Last update
+      %p= @service.updated_at
```
mkasztelnik: Please use `I18n.l` here.

PR #1442, opened 2020-03-20
Fix casting parameters variables in range parameter

goreck888:


mkasztelnik:
See the last changes in !1425 where attributes API is used. This is simpler solution than created in the scope of this PR.

PR #1445, opened 2020-03-23
Select and multiselect parameter

martaswiatkowska:


app/models/parameter/multiselect.rb
```diff
@@ -1,6 +1,25 @@
...
 class Parameter::Multiselect < Parameter
-  attr_accessor :values, :min, :max
-end
+  attribute :values, :string_array
+  attribute :min, :integer, default: 2
+
+  validates :values, presence: true
+  validates :min, presence: true
+  validate do
+    if min && values && min >= values.length
```
mkasztelnik: You should also take into account situation when `values` is `nil`

app/views/services/configurations/_attributes.html.haml
```diff
@@ -6,6 +6,6 @@
         - project_item.property_values.each do |attr|
           = form.simple_fields_for :property_values, attr do |field|
             = render "services/configurations/attributes/#{attr.type}", attr: attr, field: field
-          - rescue
+          - rescue => e
```
mkasztelnik: We should not change the attribute in project item behavior. Please rollback this change.

PR #1447, opened 2020-03-23
Quantity price parameter migration

goreck888:
Migrate type names to underscore.
Especially change type from 'quantity-price' to 'quantity_price'


mkasztelnik:
@agpul checked on production and it seems like there is no quantity parameter defined right now. As a conclusion, this migration is not needed.

PR #1455, opened 2020-03-25
Fix existing tests

goreck888:
One test is failing because select parameter is not yet implemented


spec/models/offer_spec.rb
```diff
@@ -27,8 +27,8 @@
   end
 
   context "#parameters" do
-    it "should disallow null" do
-      expect(build(:offer, parameters: nil).valid?).to be_falsey
+    it "should allow null" do
```
mkasztelnik: It seems like this test is not valid anymore. I propose to remove it.

PR #1475, opened 2020-04-02
Add matomo tracker for marketplace

goreck888:
This PR contains integration with PC matomo tracking service.
Service pages visits, orders and ratings are marked as matomo events:
- Category of events => "Service"
- Order/AddToProject => "AddToProject"
- Service rate => "Rate"
- Page visit => "Visit"
With event name => service ID recognizable by PC applications
For rating event value is `ServiceOpinion.service_rating`

For test case you need:
- visit a few services pages
- add to project service
- rate a service

Then I'll check if all cases make events for matomo

### DIFF
--- a/app/views/layouts/_head.html.haml
+++ b/app/views/layouts/_head.html.haml
@@ -9,5 +9,7 @@
   = favicon_pack_tag "favicon-180x180.png", rel: "apple-touch-icon", type: "image/png"
   = stylesheet_pack_tag "application"
   = javascript_pack_tag "application"
-  = render "layouts/google_analytics"
-  = render "layouts/hotjar_trackingcode"
+  - if Rails.env.production?
+    = render "layouts/matomo"
+    = render "layouts/google_analytics"
+    = render "layouts/hotjar_trackingcode"
--- a/app/views/layouts/_matomo.html.haml
+++ b/app/views/layouts/_matomo.html.haml
@@ -1,0 +1,28 @@
+%script
+:javascript
+  let _paq = window._paq || [];
+  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+  _paq.push(['trackPageView']);
+  _paq.push(['enableLinkTracking']);
+  (function() {
+    let u="#{Mp::Application.config.matomo_url}";
+    _paq.push(['setTrackerUrl', u+'matomo.php']);
+    _paq.push(['setSiteId', '1']);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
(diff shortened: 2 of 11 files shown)


app/views/layouts/_head.html.haml
```diff
@@ -9,5 +9,6 @@
   = favicon_link_tag asset_path("favicon-180x180.png"), rel: "apple-touch-icon", type: "image/png"
   = stylesheet_pack_tag "application"
   = javascript_pack_tag "application"
+  = render "layouts/matomo"
```
mkasztelnik, line 12: I propose one small improvement here. 

```haml
- if Rails.env.production?
  = render "layouts/matomo"
  = render "layouts/google_analytics"
￼  = render "layouts/hotjar_trackingcode"
```

app/views/layouts/_matomo.html.haml
```diff
@@ -0,0 +1,13 @@
+%script
+:javascript
+  var _paq = _paq || [];
+  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+  _paq.push(['trackPageView']);
+  _paq.push(['enableLinkTracking']);
+  (function() {
+    var u="//www.einfracentral.eu/matomo/";
```
mkasztelnik, line 8: Maybe this can be extracted to ENV variable?

app/views/services/_header.html.haml
```diff
@@ -30,6 +30,7 @@
     - if policy(service).order?
       = link_to t(".order"),
                     service_offers_path(service),
-                    class: "btn btn-primary d-block mb-3"
+                    class: "btn btn-primary d-block mb-3",
+                    onclick: "_paq.push(['trackEvent', 'Service', 'Order']);"
```
mkasztelnik, line 34: This does not mean that service was ordered, use only starts ordering process - is it ok?

app/views/layouts/_head.html.haml
goreck888: I want to add it when everything will works fine.

app/views/services/_header.html.haml
goreck888: After consultation with @agpul we need to count real orders, so it's bad placed trigger.

app/views/layouts/_matomo.html.haml
goreck888: Done.

agpul:
I've tested. Please check and report if anything changes.

app/helpers/order_helper.rb
```diff
@@ -14,4 +14,16 @@ def order_nav_link(title, path, state)
       end
     end
   end
+
+  def order_submit_options(step_key, orderable, id)
```
martaswiatkowska, line 18: Where do you use this?

goreck888: I forget about deleting it. Thanks.

PR #1482, opened 2020-04-08
Email signature separator should be two dashes and a space

gwarf:
Email signature separator should be two dashes *and* a space, as per 
 https://tools.ietf.org/html/rfc3676#section-4.3
I'm not sure how HAML deals with withe space at the end of a line, so please check, but this is required so that the signature can be identified as a signature.
I'm not reading HTML messages but `mailer.html.haml` may also need a similar change.

### DIFF
--- a/app/views/layouts/mailer.text.haml
+++ b/app/views/layouts/mailer.text.haml
@@ -1,6 +1,6 @@
 = yield
 
 :plain
-  ---
+  -- 
   Kind regards,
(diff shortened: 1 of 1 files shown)


goreck888:
We did it in another PR, and we had to do it another way because RuboCop and HAML specification doesn't allow a whitespace in the end of line. Thanks for your involvement!

PR #1484, opened 2020-04-14
1439 places visualization

### LINKED ISSUE #1439: Visualize offering of EOSC Resources on maps
 The User can visualize offerings of EOSC Resources on (interactive) maps on the EOSC Portal.

**Describe alternatives you've considered**

**Additional context**
eIC reference:

Invisio project: https://projects.invisionapp.com/share/B6WM3272EXC#/screens/411186439_Item-Detail_Copy_2

martaswiatkowska:
Geant Open(sample with single country):
image
Amber(Europa)
image
Zenodo(World)
image

Fix #1439


app/helpers/service_helper.rb
```diff
@@ -59,4 +59,16 @@ def service_logo(service)
       image_tag "eosc-img.png"
     end
   end
+
+  def getCountries(places)
```
mkasztelnik, line 63: I have 2 issues connected with this method:
  1. the name is javaish rather than rubyish - use `_` instead of camel case notation, try to avoid unnecessary words like `get`, `set`
  1. I would move this method to `Country` class with more/less this signature `Country.find_by`. Example usage:
```ruby
  Country.find_by(place: "Europe")
  Country.find_by(name: "Poland")
```

app/javascript/app/controllers/map_controller.js
```diff
@@ -0,0 +1,52 @@
...
+import maps from 'highcharts/modules/map'
+maps(Highcharts)
+
+import europe from "@highcharts/map-collection/custom/europe.geo.json"
+import world from "@highcharts/map-collection/custom/world.geo.json"
+
+export default class extends Controller {
+  static targets = []
+
+  initialize(){
```
mkasztelnik, line 12: If empty maybe we can remove it?

app/javascript/app/controllers/map_controller.js
```diff
@@ -0,0 +1,52 @@
...
+
+  connect() {
+    this.generateChart(this.element);
+  }
+
+  generateChart(element) {
+    const region = element.dataset.region;
+    var map = europe;
+    if(region == "World") {
+      map = world
```
mkasztelnik, line 23: Please try to be consistent - sometimes there are `;` at the end (:+1:), other times `;` is missing.

app/javascript/app/controllers/map_controller.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    Highcharts.mapChart(element, {
+      chart: {
+        map: map,
+        height: "100%",
+        width: '223'
+      },
+      title: {
+        text: ''
```
mkasztelnik, line 33: Isn't it default?

martaswiatkowska: No, default is "Chart title"

app/models/country.rb
```diff
@@ -28,6 +34,30 @@ def all
...
+      elsif places == "World"
+        countries = Country.world
+      else
+        countries = [Country.find_by_name(places)]
+      end
+      countries
+    end
+
+    def european_union
+      EUROPEAN_UNION.map { |c| self.for(c) }
```
goreck888, line 51: What do you think about call here method like `ISO3166::Country.all.select { |c| c.in_eu? }`? I suppose it might be better solution, because we don't have to remember if something will change, e.g. Brexit will come true.

app/models/country.rb
```diff
@@ -28,6 +34,30 @@ def all
       @all ||= (ISO3166::Country.find_all_countries_by_region("Europe") +
                 [ISO3166::Country.new("N/E")]).sort
     end
+
+    def countries_for_places(places)
```
mkasztelnik, line 38: Let's try to read this method invocation `Country.countries_for_places`. It looks like a duplication for me, maybe simple: `Country.for_region("Europe")`?

mkasztelnik: What is more, I see this method is doing 2 things:
  1. Looks for the countries in the region
  1. Looks for the country by given name
Maybe we can simplify this?

app/models/country.rb
martaswiatkowska: Ok, I didn't notice this function.

PR #1485, opened 2020-04-16
Fix research area counter issue

mkasztelnik:
This is another try to fix the research area issue in the filter zone. The issue
is connected with the research area tree-like structure. The intuition says
that parents should count also its children.

Here I'm trying to do exactly this by adding all parents to service
research areas elaticsearch record.

Closes #1230


agpul:
My intuition is similar. For me, this is a good solution.

mkasztelnik:
:dancers: :dancers: :dancers: :dancers: :dancers: :dancers: :dancers: :dancers:

PR #1521, opened 2020-05-21
Fix project timestamp

martaswiatkowska:
Project creation_at isn't automatically overridden because have default on db.


goreck888:
I think we should make a migration deleting default from timestamps in projects and then we don't need to make any method to overwrite timestamps.

Issue #1506, opened 2020-05-11
Add instance badges 


bwilk:
**Is your feature request related to a problem? Please describe.**
When a user opens our development or PR instance they can have an impression of viewing the production instance

**Describe the solution you'd like**
Add "Development instance" (master), "Beta instance" (staging), "Test instance" (PR)

abacz:
I propose to show instance label in the top, left corner with possibility to hide.
1. Visible -> https://invis.io/VGXBRHAQJNA#/418030864_Instance_Badges_Show
2. Hidden -> https://invis.io/NMXBRHEQC9D#/418030865_Instance_Badges_Hide

PR #1524, opened 2020-05-23
[#1506] Instance badges

goreck888:
By setting `MP_INSTANCE` environmental variable
with proper value badge should be visible.
Values for `MP_INSTANCE` variable:
- `pr` for pr instances
- `master` for master instance
- `staging` for staging instance
It's important to not set `MP_INSTANCE` on production


agpul:
Can I check this PR in some way?

PR #1542, opened 2020-06-10
Get first element if places/termOfUse/language is an array

martaswiatkowska:
Save first element if places/termOfUse/language is an Array. 

I added some infos about jms to README.md


goreck888:
Are you sure it's a valid solution? Maybe we should change type of places to an array?

martaswiatkowska:
This is only fix to jms. Places will be changed to array in RDT

goreck888:
Ok

PR #1545, opened 2020-06-15
[#1544] change languages to language availability

### LINKED ISSUE #1544: Change languages to language_availability and type to array

goreck888:
Change field `languages`:
- type from `string` to `string[ ]`
- name `languages` to `language_availability`

Fixes #1544

### DIFF
--- a/app/services/service/pc_create_or_update.rb
+++ b/app/services/service/pc_create_or_update.rb
@@ -68,7 +68,7 @@ class Service::PcCreateOrUpdate
                       data["userBase"]].join("\n"),
         tagline: data["tagline"].blank? ? "NO IMPORTED TAGLINE" : data["tagline"],
         places: map_places(data["places"]["place"]) || "World",
-        languages: data["languages"] || ["English"],
+        language_availability: [data["languages"]["language"]] || ["English"],
         dedicated_for: [],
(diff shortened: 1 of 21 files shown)


goreck888:
Now layout in service view is wrong. We are waiting for merge PRs with places and languages and then we make a fix for displaying these fields - mentioned in #1546

app/services/service/pc_create_or_update.rb
```diff
@@ -68,7 +68,7 @@ def map_service(data)
                       data["userBase"]].join("\n"),
         tagline: data["tagline"].blank? ? "NO IMPORTED TAGLINE" : data["tagline"],
         places: map_places(data["places"]["place"]) || "World",
-        languages: data["languages"]["language"] || "English",
+        language_availability: [data["languages"]["language"]] || ["English"],
```
martaswiatkowska, line 71: I think  this wouldn't work because when data["languages"]["language"] are the array "[]" you will get [["sth", "sth"]]. Use Array()

goreck888: Ok, thanks.

agpul:
Still empty fields.

goreck888:
> Still empty fields.

After task run everything is ok.

PR #1551, opened 2020-06-17
[#1549] change service type to order type

### LINKED ISSUE #1549: Change service_type to order_type

goreck888:
Fixes #1549


goreck888:
> I've checked Marketplace and I think that it is OK. Should I check the JIRA transition? Did you change something there?

You can check it but everything should working fine.

Issue #1448, opened 2020-03-25
improve project view in contact tab


abacz:
This view look not correctly:
![Screenshot 2020-03-25 at 08.42.50.png](https://images.zenhubusercontent.com/5b4c669f624eef429c4400c8/e9713fd4-9299-4ba0-b241-cb8dc4d21f44)

Please improve box "Conversation not started yet".

Correct preview:
https://invis.io/Y7V6A0UDMSK#/396859681_Projects_Sidebar_Contact_With_Support_Empty

agpul:
What about this ticket - is it done?

agpul:
I think it was done so I close this issue.

PR #1562, opened 2020-06-22
contact tab improvements

jarekzet:
Contact tab layout mods according to #1448 
image


jarekzet:
added mods in contact provider form also + removed required star icon

jarekzet:
added change log entry

PR #1570, opened 2020-06-25
[#1559] New associations in service

### LINKED ISSUE #1559: Add new associations in service model
Add new associations:
- funding_body
- funding_program

goreck888:
Add new associations mentioned as API vocabularies:
- FundingBody
- FundingProgram

Fixes #1559

### DIFF
--- a/app/models/funding_body.rb
+++ b/app/models/funding_body.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class FundingBody < Vocabulary
+end
--- a/app/models/service_vocabulary.rb
+++ b/app/models/service_vocabulary.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class ServiceVocabulary < ApplicationRecord
+  belongs_to :service
+  belongs_to :vocabulary, polymorphic: true
+  before_create :set_vocabulary_type
+
+  validates :service, presence: true
+  validates :vocabulary, presence: true
+
+  def set_vocabulary_type
+    self.vocabulary_type = vocabulary.type
+  end
+end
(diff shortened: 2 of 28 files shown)


agpul:
I can't reach this PR, @wziajka can you check?

app/models/funding_body.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class FundingBody < Vocabulary
+  has_many :service_vocabularies, as: :vocabulary, source: "FundingBody"
```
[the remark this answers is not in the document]
goreck888: You are right. Done.

app/models/service_vocabulary.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ServiceVocabulary < ApplicationRecord
+  belongs_to :service
+  belongs_to :vocabulary, polymorphic: true
```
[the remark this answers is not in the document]
goreck888: It has to be polymorphic. We have one table for different child objects. When you remove `polymorphic: true` option, it try to find different associations table.

.rake_tasks~
```diff
@@ -0,0 +1,95 @@
+about
```
martaswiatkowska, line 1: What's that?

goreck888: Sorry, this file creates always when something goes wrong with rake task. I delete it now.

goreck888:
> Add also loading vocabularies to dev:prime

Done.

lib/tasks/dev.rake
```diff
@@ -5,6 +5,7 @@ namespace :dev do
   task prime: "db:setup" do
     yaml_hash = YAML.load_file("db/data.yml")
 
+    create_vocabularies
```
martaswiatkowska, line 8: Add To service.update! some examples of this new attributes

goreck888: Done.

PR #1577, opened 2020-07-03
Rdt Phase field

martaswiatkowska:
Added TRl, and Life_cycle_status as vocabularies with validation(maximum 1)
Deleted Phase


examples.txt
```diff
@@ -0,0 +1,808 @@
+example_id                                                         | status  | run_time        |
```
goreck888, line 1: What is that?

db/migrate/20200703124752_remove_phase_from_service.rb
```diff
@@ -0,0 +1,30 @@
...
+        elsif  d["phase"] == "beta"
+          execute(
+            <<~SQL
+              INSERT INTO service_vocabularies(service_id, vocabulary_id, vocabulary_type, created_at, updated_at)
+              VALUES ( #{d["id"]}, #{beta.id}, 'LifeCycleStatus', '#{Time.now.to_s}', '#{Time.now.to_s}')
+              SQL
+          )
+        end
+      }
+    remove_column :services, :phase
```
goreck888, line 24: Don't mix data and schema migrations into one. You should do `remove_column` in separate migration.

Issue #1546, opened 2020-06-15
Fix display for array fields for places and languages


goreck888:
**Describe the bug**
We change places and languages fields to arrays. Now it use `to_s` method which causes it looks bad (e.g. `["Engilsh"]` instead of `English`). After merge #1545 and places PR we have to do this fix

**To Reproduce**
1. Go to any service view
2. See `Places and languages` section in the sidebar

**Expected behavior**
Expect it shows an unordered list with places and languages.
 
**Screenshots**
![languages](https://user-images.githubusercontent.com/25662749/84663439-79c3aa00-af1d-11ea-9bfe-d867d5e7202e.png)

**Additional context**

PR #1579, opened 2020-07-08
Fix languages and geographical_availabilities

martaswiatkowska:
Add nested option for a specific field

#1546


goreck888:
> image
> If there is no data in languages I think that this second dot should not be visible.

There is a need to run task `rdt:repair_language_data`. I ran this task, after complete check it once more.

PR #1586, opened 2020-07-15
Fixed service sorting by name in backoffice

martaswiatkowska:
Forgotten backoffice name service sorting fixes


agpul:
Everything works fine but look:
image

Why OpenAire is the first one?

PR #1589, opened 2020-07-17
Fixed migration rewriting contacts

martaswiatkowska:


app/models/service.rb
```diff
@@ -91,8 +91,8 @@ class Service < ApplicationRecord
   has_many :service_target_users, dependent: :destroy
   has_many :target_users, through: :service_target_users
 
-  has_one :main_contact, as: :contactable, class_name: "MainContact", dependent: :destroy, autosave: true
-  has_many :public_contacts, as: :contactable, source: :service_contacts, class_name: "PublicContact",
+  has_one :main_contact, as: :contactable, dependent: :destroy, autosave: true
+  has_many :public_contacts, as: :contactable, source: :service_contacts,
```
goreck888, line 95: Source is also bad.

goreck888: And don't remove `class_name` because it makes database issues.

db/migrate/20200703170548_rewrite_contact_emails_to_public_contacts.rb
```diff
@@ -5,8 +5,8 @@ def up
       emails = service["contact_emails"].tr("{}", "").split(",").map
       emails.each do |email|
         execute(<<~SQL
-          INSERT INTO service_contacts(email, service_id, type, created_at, updated_at)
-          VALUES ('#{email}', '#{service["id"]}', 'PublicContact', '#{Time.now.to_s}', '#{Time.now.to_s}')
+          INSERT INTO contacts(email, type, contactable_id, contactable_type, created_at, updated_at)
+          VALUES ('#{email}', '#{service["id"]}', 'PublicContact', 'Service', '#{Time.now.to_s}', '#{Time.now.to_s}')
```
goreck888, line 9: Now you have wrong order of columns, you save `contactable_id` in column `type` and vice versa

martaswiatkowska: eh..

PR #1595, opened 2020-07-20
[#1098] Apply service access model

### LINKED ISSUE #1098: Apply service access model

There is no description of the service access type, so the user doesn’t know if the below service is free/ open for all. How to access it, what should be done? 

The service access model ( the correct naming to be agreed) should be more visible to the user. 
We need to plan how to show it to the user, things to project:

- [ ]  Information about the Service model should be visible in the Service list 

- [ ] Information about the Service model should be visible in the Service view 

- [ ] Be able to Filter by Service model ( to decide on the Titles that should be shown to the user).

goreck888:
Add backend and frontend markups for service type
Add service type filter written as order type
for future compatibility
Fixes #1098


jarekzet:
Added styling for access type badges on service list and service details views.

app/helpers/service_helper.rb
```diff
@@ -48,6 +48,10 @@ def providers_text(service)
     service.providers.map { |target| target.name }
   end
 
+  def order_type(service)
+    service&.service_type || service&.offers&.first&.offer_type || nil
```
bwilk, line 52: For the services having offers with different access types "VARIOUS ACCESS TYPES" should be displayed both on the service list and service view

goreck888: Ok, I'll wait for @jarekzet css mods. All my changes are stashed locally now.

MinLidia:
@goreck888 In the searching fields the name Service type should be changed to Access Type

MinLidia:
I have bug when filtering for External access type 
Screenshot 2020-07-23 at 12 11 00

app/models/filter/order_type.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class Filter::OrderType < Filter
+  def initialize(params = {})
+    super(params: params.fetch(:params, {}),
+          field_name: "service_type",
```
martaswiatkowska, line 6: service_type? not order_type?

app/models/service/search.rb
```diff
@@ -26,6 +26,7 @@ def search_data
       providers: providers.map(&:id),
       platforms: platforms.map(&:id),
       target_groups: target_groups.map(&:id),
+      service_type: [service_type] << offers.map(&:offer_type),
```
martaswiatkowska, line 29: service doesn't have service_type in my scheme

PR #1601, opened 2020-07-23
JMS service deletion

martaswiatkowska:
Service can be deleted only by JMS
Service deletion mean that Service status is set to "deleted"
Deleted service cannot be edited, published and offers cannot be edited, deleted, published/draft or added.
Deleted service isn't shown for users. In bacoffice are displayed with badge "deleted"

### DIFF
--- a/app/policies/backoffice/offer_policy.rb
+++ b/app/policies/backoffice/offer_policy.rb
@@ -15,31 +15,32 @@ class Backoffice::OfferPolicy < ApplicationPolicy
   end
 
   def new?
-    service_portfolio_manager? || record.service.owned_by?(user)
+    (service_portfolio_manager? || record.service.owned_by?(user)) &&
+      !service_deleted?
   end
 
   def create?
-    managed?
+    managed? && !service_deleted?
   end
 
   def edit?
-    managed?
+    managed? && !service_deleted?
   end
 
   def update?
-    managed?
+    managed? && !service_deleted?
   end
 
   def destroy?
-    managed? && orderless?
+    managed? && orderless? && !service_deleted?
   end
 
   def publish?
-    service_portfolio_manager? && record.draft?
+    service_portfolio_manager? && record.draft? && !service_deleted?
   end
 
   def draft?
-    service_portfolio_manager? && record.published?
+    service_portfolio_manager? && record.published? && !service_deleted?
   end
 
   def permitted_attributes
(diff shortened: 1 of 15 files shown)


app/policies/backoffice/offer_policy.rb
```diff
@@ -15,27 +15,28 @@ def resolve
...
   end
 
   def update?
-    managed?
+    managed? && !service_deleted?
   end
 
   def destroy?
-    managed? && orderless?
+    managed? && orderless? && !service_deleted? && !service_deleted?
```
goreck888, line 35: Why did you add `!service_deleted?` twice?

app/policies/backoffice/service_policy.rb
```diff
@@ -65,7 +65,9 @@ def create?
   end
 
   def update?
-    service_portfolio_manager? || (record.draft? && owned_service?)
+    (service_portfolio_manager? ||
+     (record.draft? && owned_service?)) &&
+    !record.deleted?
   end
 
   def publish?
```
goreck888, line 73: I think you should add `!record.deleted?` here and in `publish_unverified` too. I mean we shouldn't publish deleted services.

spec/policies/backoffice/offer_policy_spec.rb
```diff
@@ -72,4 +72,22 @@
...
+  context "When offer service is deleted" do
+    let(:service) { create(:service, owners: [owner], status: :deleted) }
+    let(:offer) { build(:offer, service: service, status: :published) }
+
+    permissions :create?, :edit?, :update?, :destroy? do
+      it "danies access to service portfolio manager" do
+        expect(subject).to_not permit(service_portfolio_manager, offer)
+      end
+
+      it "danies access to service owner" do
```
goreck888, line 85: Add tests for publish

app/policies/backoffice/service_policy.rb
martaswiatkowska: ok but this status should be removed when rdt will be released I think

spec/policies/backoffice/offer_policy_spec.rb
martaswiatkowska: But offer cannot be published twice. Look at offer status

goreck888: Ok, I didn't notice.

PR #1602, opened 2020-07-24
[#1424] Remove duplicated ids on services page

### LINKED ISSUE #1424: Two elements with the same ID on the services list page
**Describe the bug**
Found 2 elements with non-unique id #sort

**To Reproduce**

1. Go to http://localhost:5000/services
2. Open web inspector (ctr+shift+i in chrome/chromium)
3. Open `Console` tab

**Expected behavior**
We don't need to duplicate id to make form works. It uses name value to generate form key no id.

goreck888:
Remove duplicated "sort" ids in services index
Fixes #1424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ Please view this file on the master branch, on stable branches it's out of date.
 - instance badge display on backoffice section (@goreck888)
 - contact form tab layout modifications (@jarekzet)
 - Service preview (@goreck888)
- 
+- Duplicated "sort" ids on services page (@goreck888)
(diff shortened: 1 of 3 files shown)


goreck888:
This PR has no influence on application, it's just cleaning html code and avoid duplicating ids.

goreck888:
> I propose another solution:
> You can just delete "#sort" from https://github.com/cyfronet-fid/marketplace/blob/master/app/views/services/_order.html.haml#L6
> As I checked it isn't styled

We can't do this. After remove `#sort` sorting won't work.

PR #1607, opened 2020-07-28
Update message order 

martaswiatkowska:
Only accept the latest changes from jms.
Changed eic import of last_update to date format because we receiving int 
For provider this changes can be appied if Porvider model will be 

To merge after  #1601

### DIFF
--- a/lib/import/eic.rb
+++ b/lib/import/eic.rb
@@ -63,7 +63,6 @@ module Import
           .each do |service_data|
         service = service_data["service"]
         eid = service["id"]
-
         output.append(service_data)
         # TODO this should be moved to Offer.webpage
         url = service["webpage"]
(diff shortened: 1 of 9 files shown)


lib/import/eic.rb
```diff
@@ -189,13 +188,14 @@ def call
             scientific_domains: [@scientific_domain_other],
             version: version || "",
             target_users: map_target_users(target_users),
-            last_update: last_update,
+            last_update: last_update ? Time.at(last_update&./1000) : Time.now,
```
goreck888, line 191: Do you handle case when `last_update` is `nil`?

martaswiatkowska: Yes then is set as Time.now

PR #1609, opened 2020-07-29
Added target_users page to home

martaswiatkowska:
Added page with listed target_users:
image
Changed "target groups" to "target users" in link


agpul:
> Added page with listed target_users:
> image
> Changed "target groups" to "target users" in link

This picture is old or new?

PR #1617, opened 2020-08-11
fix for home page on windows

jarekzet:
This is a fix for an issue on homepage on windows, where main tabs were not clickable.


jarekzet:
added CSS mods and moved welcome modal to different position (above header)

PR #1622, opened 2020-09-02
Split basic configuration

goreck888:


agpul:
Can you provide me a short scenario of how this PR can be tested?

agpul (approved):

app/controllers/pages_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class PagesController < ApplicationController
   def about
+    ab_finished(:recommendations)
```
JanKapala, line 5: There is no ab_test funstion usage. Maybe should it be added?

app/helpers/application_helper.rb
```diff
@@ -47,4 +47,8 @@ def meta_robots
   def yield_content!(content_key)
     view_flow.content.delete(content_key)
   end
+
+  def placeholder(variant, text = "Placeholder - link to about", link = about_path)
+    render "layouts/placeholder", text: text, link: link, variant: variant
+  end
```
JanKapala, line 53: There is no usage of this placeholder. did you mean to use it with ab_test?

config/experiments.yml
```diff
@@ -0,0 +1,8 @@
+recommendations:
+  alternatives:
+    - name: top
+      percent: 25
+    - name: middle
+      percent: 25
+    - name: none
+      percent: 50
```
JanKapala, line 8: This experiment isn't displayed in the Split Dashboard

app/helpers/application_helper.rb
JanKapala, line 53: There is no usage of this placeholder. Did you mean to use it with ab_test?

app/controllers/pages_controller.rb
JanKapala, line 5: There is no ab_test function usage. Maybe should it be added?

JanKapala (changes requested):
The ab_finished function is used but ab_test isn't so an experiment is not displayed in the Split Dashboard. Maybe should it be added to make example of usage A/B testing more end-to-end? @goreck888

config/experiments.yml
goreck888: Yes, I removed experiment used for tests in the last commit, just forget about remove it in config files.

app/helpers/application_helper.rb
goreck888: I wanted to keep it for future implementations of tests but I can remove it.

app/controllers/pages_controller.rb
goreck888: I forgot about it

wujuu (approved):

JanKapala (approved):

PR #1626, opened 2020-09-14
Fix multimedia and use_cases_url fields and logo

goreck888:
If logo cannot be downloaded or is bad type,
then is not saved


app/services/service/pc_create_or_update.rb
```diff
@@ -164,11 +164,12 @@ def save_logo(service, image_url)
         logo.write(img.to_blob)
         logo.seek(0)
         logo_content_type = "image/png"
-        logo
       end
-      unless logo.nil?
+      unless logo.blank? || logo_content_type == "text/html"
```
michal-szostak, line 168: Are you sure that `logo_content_type == "text/html"` is good enough? Isn't it better to whitelist accepted types?
(there might be cases of `"application/json"` responses or others.

spec/factories/jms_service.rb
```diff
@@ -110,9 +110,10 @@
         "accessPolicy" => "http://openminted.eu/pricing/",
         "statusMonitoring" => "http://openminted.eu/monitoring/",
         "maintenance" => "http://openminted.eu/maintenance/",
-        "multimedia" => [
+        "multimedia" => {
+            "multimedia" => [
```
michal-szostak, line 114: Who decided this latest API to have two nested `multimedia` keys?

app/jobs/jms/manage_message_job.rb
```diff
@@ -3,7 +3,7 @@
 require "nori"
 
 class Jms::ManageMessageJob < ApplicationJob
-  queue_as :jms
+  queue_as :pc_subscriber
```
michal-szostak, line 6: Why is the name of the job changed?

agpul (changes requested):
1. Field use_cases, on the provider component side we have two different fields:
image
on the MP side, we treat is as a one link:
image
2. The same situation is if it comes to the Multimedia field
3.  Filed categories:
- provider component:
image
-MP:
empty one:
image

4. Check mapping for scientific domains, I chose natural science on the provider component side, in Marketplace we map it as "Other"
5.  Target user:
Provider component: researcher
MP: empty
5. Language Availability (*)
Provider component: English
MP: empty
6. Helpdesk email and Security contact email
Provider component: these fields are filled
MP: empty

spec/factories/jms_service.rb
goreck888: This is eIC schema issue.

app/jobs/jms/manage_message_job.rb
goreck888: Because queue name should tell us what is for and not the tool we use.

app/services/service/pc_create_or_update.rb
goreck888: Ok, I'll change it.

agpul (approved):

martaswiatkowska (approved):

PR #1636, opened 2020-10-08
Include external ids for services paths

goreck888:
Allow to use eIC ids for service search in all scopes


goreck888:
Way to test - after jms service update try to use link with eIC id and check if it loads in all scopes (backoffice, ordering, etc.)

bwilk (approved):

michal-szostak (approved):
The code looks alright, if someone tested it manually and it works correctly then it's good to go

PR #1641, opened 2020-10-14
New resource presentation page fixed

michal-szostak:


roksanaer (commented):
General comments:
PR presents itself well. Info propagation between the backoffice and ux is ok. New order types are working OK (along with the order paths mapping), systems seems to work in its every element

Issues I've came across while testing (service presentation and propagation of a service detail from the backoffice). The 5th one NEEDS TO BE FIXED before official release

1: Services without a tagline have a 'NO IMPORTED TAGLINE' comment. While in the backoffice it seems OK, this behaviour should be changed in the user view (e.g. blank space).
image

2. if a service does not have any language chosen, a bullet is shown anyway to the user.
image
Both of the remarks should be assessed wether they should be handled.
On one hand, this situation should not be possible in RDT 3.0 as language is a mandatory field. same applies to the 1st remark. so maybe it's not so urgent for now but definitely worth taking care of in the future.
Last thing is, are we sure we do not have an empty field situation in any of our services, in case we do not change our current database. Pleas assess.
3. tags (in the detail view) should be live links. would be very beneficial if we could change it before the release
image
4. same with categories but it's not as urgent.
5. Order type options need a label
image
image

PR #1666, opened 2020-10-29
Sort resource form fields in the backoffice

goreck888:
Sort resource form similar to the PC component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Added
 
 ### Changed
+- Resource form fields sort similar as PC (@goreck888) 
 
 ### Deprecated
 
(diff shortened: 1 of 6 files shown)


abacz:
What i will do ;)

- clickable block (toggle), not only name + some arrow (up/down)
- remove some gaps (e.g. mt-5 here -> 
Screenshot 2020-10-30 at 11 24 01)
- some inputs/selects have green tic which overlap inputs text. More padding from right or different input width (if possible)
- .is-disable in choices__list--multiple i would set bg-color to #c1c1c1 and color: #717070 (more or less;)) to look more "disable" ;) 
Screenshot 2020-10-30 at 11 31 01
- remove borders in contact section

agpul (changes requested):
Moreover:
1. 
image
2. 
image

goreck888:
> Moreover:
> 1.
> image
> 2.
> image

Everything should be now ok

agpul (approved):

michal-szostak (approved):

PR #1670, opened 2020-11-03
Handle parse errors in jms-subscriber

goreck888:


lib/jms/subscriber.rb
```diff
@@ -33,7 +33,7 @@ def run
         log "Arrived message"
         Jms::ManageMessage.new(msg, @eic_base_url, @logger).call
         @client.ack(msg)
-      rescue StandardError => e
+      rescue ResourceParseError, JSON::ParserError, StandardError => e
```
wujuu, line 36: Why is ResourceParseError defined in the subscriber class instead of Jms:ManageMessage?

wujuu: It's good that we are catching this exception here, but I think it would be clearer if the error was decalred in the class/module that is raising this error.

wujuu (approved):

JanKapala (approved):

PR #1674, opened 2020-11-05
Fix rake dev:prime task

goreck888:


wujuu:
With this fixed, do we still need to run ```import:eic``` task when seeding?

goreck888:
> With this fixed, do we still need to run `import:eic` task when seeding?

Not really. If you don't need actual productional data, then you can just run `dev:prime`. Otherwise use `import:eic`

wujuu (approved):

PR #1675, opened 2020-11-09
Add `remove array element` in forms

goreck888:
In the backoffice service form added remove link for array fields: multimedia, use_cases_url, changelog, related_platforms etc.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -9,6 +9,8 @@ class Service < ApplicationRecord
 
   acts_as_taggable
 
+  before_save :remove_empty_array_fields
+
   has_one_attached :logo
 
   enum order_type: {
@@ -308,6 +310,15 @@ class Service < ApplicationRecord
 
 
   private
+    def remove_empty_array_fields
+      array_fields = [:multimedia, :use_cases_url, :certifications,
+                      :standards, :open_source_technologies,
+                      :changelog, :related_platforms, :grant_project_names]
+      array_fields.each do |field|
+        send(field).present? ? send(:"#{field}=", send(field).reject(&:blank?)) : send(:"#{field}=", [])
+      end
+    end
+
     def open_access_or_external?
       open_access? || external
     end
--- a/app/views/backoffice/services/_form.html.haml
+++ b/app/views/backoffice/services/_form.html.haml
@@ -49,17 +49,11 @@
         = f.input :multimedia, multiple: true, input_html: { class: "form-control" },
           wrapper_html: { "data-target" => "service.multimedia" },
           disabled: cant_edit([multimedia: []]), as: :array
-        %a.text-primary{ "data-action" => "click->service#addNewArrayField",
-          "data-wrapper" => "service_multimedia",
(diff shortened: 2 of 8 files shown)


agpul:
Can I test this PR? How? Can you add a description?

bwilk:
@agpul this is about removing array items. Looks ugly but works. Maybe the frontend logic could be better if the empty fields weren't visualised except after intentionally using "Add new .... "

app/models/service.rb
```diff
@@ -308,6 +310,16 @@ def orderable?
...
+    def remove_empty_array_fields
+      multimedia.reject!(&:blank?) if multimedia.present?
+      use_cases_url.reject!(&:blank?) if use_cases_url.present?
+      certifications.reject!(&:blank?) if certifications.present?
+      standards.reject!(&:blank?) if standards.present?
+      open_source_technologies.reject!(&:blank?) if open_source_technologies.present?
+      changelog.reject!(&:blank?) if changelog.present?
+      related_platforms.reject!(&:blank?) if related_platforms.present?
+      grant_project_names.reject!(&:blank?) if grant_project_names.present?
+    end
```
wujuu, line 322: Can't we refactor this into a for loop?

agpul (changes requested):
For me, it works in a random way. Sometimes when I try to "Save" changes I need to remove one item a couple of times.

app/views/backoffice/services/_form.html.haml
```diff
@@ -278,7 +278,7 @@
         %a.text-primary{ "data-action" => "click->service#addNewArrayField",
           "data-wrapper" => "service_related_platforms",
           "data-name" => "service[related_platforms[]]",
-          "data-class" => "form-control text optional" } Add additional entry
+          "data-class" => "form-control text optional" } Add new related platform
```
wujuu, line 281: Shouldn't we use I18n translations?

wujuu (changes requested):

wujuu:
Also, shouldn't the behaviour to be to not show the 'Remove' button unless it's a second (and further) resource?
image

JanKapala (commented):
Maybe some usage of locales would be nice

app/models/service.rb
goreck888: Done

goreck888:
> Also, shouldn't the behaviour to be to not show the 'Remove' button unless it's a second (and further) resource?
> image

Honestly I think it's a bad idea. When service have eg. 3 entries and want to remove first then he couldn't do this. This solution is simple, working, and doesn't need unnecessary js code to handle this. And it fits to the logic in this form - look at PublicContacts, they're working in the same way.

agpul (changes requested):
I still can remove only one element during one action.

wujuu (approved):

app/views/backoffice/services/_form.html.haml
goreck888: Done.

PR #1676, opened 2020-11-10
950 improve front in backoffice

jarekzet:
a lot of modifications in all backoffice views - start page , list views, edit forms etc.

### DIFF
--- a/app/views/backoffice/categories/show.html.haml
+++ b/app/views/backoffice/categories/show.html.haml
@@ -1,23 +1,24 @@
 - breadcrumb :backoffice_category, @category
 
 %h1= @category.name
-%h2 Description:
-%p= @category.description
-%h2 eID:
-%p= @category.eid
-%h2= @category.parent ? "Parent: #{@category.parent.name}" : "Root element"
-- if @category.logo.attached?
-  = image_tag @category.logo
+.service-box.shadow-sm.p-4.mt-3.backoffice-form
+  %h2 Description:
+  %p= @category.description
+  %h2 eID:
+  %p= @category.eid
+  %h2= @category.parent ? "Parent: #{@category.parent.name}" : "Root element"
+  - if @category.logo.attached?
+    = image_tag @category.logo
 
-%hr
-.btn-group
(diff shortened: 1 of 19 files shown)


abacz:
Scenario:
Scientific Domains list -> edit eg. Clinical Medicine.
Clicking "cancel" didn't back me to previous page (list) but to scientific domain detail.

jarekzet:
Added mods of backlinks in edit forms.

abacz (approved):

bwilk:
Please take a look at the providers edit as well. It seems odd, but this may not be the case of the current changes, but older bug

bwilk:
> Please take a look at the providers edit as well. It seems odd, but this may not be the case of the current changes, but on older bug
> 
> Screenshot 2020-11-12 at 15 48 36

ok, this isn't really an issue. Looks about the same in production.

bwilk (approved):

app/views/backoffice/categories/show.html.haml
```diff
@@ -1,23 +1,24 @@
...
+  .btn-group
+    - if policy([:backoffice, @category]).edit?
+      = link_to "Edit",
+                edit_backoffice_category_path(@category),
+                class: "btn btn-primary"
+    - if policy([:backoffice, @category]).destroy?
+      = link_to "Delete",
+                backoffice_category_path(@category),
+                method: :delete,
+                data: { confirm: "Are you sure you want to remove this category" },
```
wujuu, line 23: Shouldn't we use I18n translations? That goes for all of the explicit strings in this PR's view files.

wujuu (changes requested):

app/views/backoffice/categories/show.html.haml
goreck888: Yeah but I think we will refactor that. Michał is looking for a gem making translations similar like in python or try to find other solution but now we don't have to change everything because we don't know the way we'll refactor things

goreck888 (approved):

agpul (approved):

PR #1686, opened 2020-11-17
Fix popular resources

goreck888:
Add unverified and published resources to popular resources query

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -23,7 +23,7 @@ class HomeController < ApplicationController
       @providers_number = Provider.count
       @services_number = Service.count
       @countries_number = 32
-      @services = Service.published.includes(:providers).popular(6)
+      @services = Service.popular(6)
     end
 
     def load_platforms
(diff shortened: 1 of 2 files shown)


app/controllers/home_controller.rb
```diff
@@ -23,7 +23,7 @@ def load_services
       @providers_number = Provider.count
       @services_number = Service.count
       @countries_number = 32
-      @services = Service.published.includes(:providers).popular(6)
+      @services = Service.where(status: [:published, :unverified]).includes(:providers).popular(6)
```
wujuu, line 26: Do we want to show unverified services on the front page?

goreck888: Yes, now `unverifed` status is used for show imported services, but it's functional in the same way like published.

wujuu (changes requested):
The only thing I would maybe add is to shift the 'popular' logic to Serivce Model, to call it inside this controller like: service.top(n). Shouldn't the buissness logic like that be in the model instead of the controller?

wujuu (approved):

agpul (approved):

PR #1687, opened 2020-11-17
Add `errored` status to the imported resources

goreck888:
When resource has validation errors it saves as resource
with errored state

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Added
 - Add and configure split gem (@goreck888)
+- Errored status for imported resources (@goreck888)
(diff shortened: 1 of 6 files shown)


wujuu:
Question: Why would we persist a service that has been imported with errors?

wujuu (commented):
Question: Why would we persist a service that has been imported with errors? Is to keep track of the overall services (even which failed to be imported into MP)?

goreck888:
> Question: Why would we persist a service that has been imported with errors? Is to keep track of the overall services (even which failed to be imported into MP)?

Because we want to have a sign that service is imported but has errors. This solution is agreed with @bwilk

wujuu (approved):

agpul (approved):

Issue #1682, opened 2020-11-16
Add services api tests


goreck888:
There are missing tests for services api available at link `api/services`. Need to add them.

PR #1693, opened 2020-11-20
[#1682] Add services api tests

goreck888:


agpul (approved):

spec/requests/api/services_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it "shows only published and unverified services with correct data" do
+      body = JSON.parse(response.body)
+      expect(body.size).to eq(2)
+
+      expect(body[0]["Service Unique ID"]).to eq(published_service.id)
+      expect(body[0]["SERVICE_TYPE"]).to eq("eu.eosc.portal.services.url")
+      expect(body[0]["CONTACT_EMAIL"]).to eq(published_service.public_contacts.map(&:email))
+      expect(body[0]["SITENAME-SERVICEGROUP"]).to eq(published_service.name)
+      expect(body[0]["COUNTRY_NAME"]).to eq(published_service.geographical_availabilities.as_json)
+      expect(body[0]["URL"]).to eq(published_service.webpage_url)
```
wujuu, line 30: Maybe we should check if the body[0], contains only and exactly these keys ('Service Unique ID', 'SERVICE_TYPE', etc.). Same goes for body[1] I guess.

spec/requests/api/services_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Api::ServicesController" do
+  let!(:published_service) { create(:service, public_contacts: [build(:public_contact)]) }
+  let!(:unverified_service) { create(:service, public_contacts: [build(:public_contact)], status: :unverified) }
+  let!(:draft_service) { create(:service, public_contacts: [build(:public_contact)], status: :draft) }
+  let!(:deleted_service) { create(:service, public_contacts: [build(:public_contact)], status: :deleted) }
+
```
wujuu, line 10: Should we add also recently introduced 'errored' service here? It shouldn't be visible in the GET request as well

wujuu (changes requested):

spec/requests/api/services_controller_spec.rb
goreck888: Yeah, and that's what I check, if errored services are not visible in the API

wujuu (approved):

PR #1706, opened 2020-11-26
[#1700] Refactoring of the resource view

### LINKED ISSUE #1700: Refactoring of the resource view
## Overview

Now, the resource view (especially details) is defined in the model:

```
class Service < ApplicationRecord
#...
SIDEBAR_FIELDS = [{ name: "scientific_categorisation",
                      template: "classification",
                      fields: ["scientific_domains"],
                      nested: {
                          scientific_domains: "name"
                      }
                    },
#...
                    { name: "resource_availability_and_languages",
                      template: "map",
                      fields: ["languages", "geographical_availabilities"] }]

  HEADER_FIELDS = [{ name: "links",
                      template: "links",
#...
                      fields: ["webpage_url", "helpdesk_url", "helpdesk_email",
                               "manual_url", "training_information_url"] }]

  DETAIL_FIELDS_1 = [{ name: "classification",
                       template: "array",
                       fields: ["target_users", "access_types", "access_modes", "tag_list"],
                       with_desc: true,
                       nested: {
                           target_users: "name",
                           access_types: "name",
                           access_modes: "name",
                           tag_list: "tag"
                       } },
                       with_desc: true }]


  DETAIL_FIELDS_2 = [{ name: "public_contacts",
                       template: "object",
                       fields: ["full_name", "email", "phone", "position_in_organisation"],
#...                   
                     ]

  DETAIL_FIELDS_3 = [{ name: "changelog",
                       template: "changelog",
                       fields: ["changelog"] }]
```

Those definitions should not be part of the model but either controller or some data structure which is around view layer.
I also believe that it might be a good place and time to consider upgrading the code to be more generic than 3 defined columns, and use a little more generic approach - creating an additional / surrogate component or helper can also be a scope of this issue.

wujuu:
Closes #1700


agpul (approved):
From my point of view, for current changes, everything looks fine. If something changes in the future (for eg. after Michał review), I will make a second review.

app/helpers/service/details_helper.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+module Service::DetailsHelper
+  def service_details_columns
+    [
+        column(classification, availability, marketing, dependencies, attribution, order),
+        column(public_contacts, maturity_information, management, financial_information),
+        column(changelog)
+    ]
+  end
```
JanKapala, line 10: Why not just arrays rather than column function?

app/views/services/_details.html.haml
```diff
@@ -1,51 +1,23 @@
...
-                            nested: group[:nested] || "",
-                            with_desc: group[:with_desc] || false
+    - service_details_columns.each do |column|
+      %span{ class: ["col-#{details_column_width_sm(service_details_columns)}",
+                     "col-sm#{details_column_width_sm(service_details_columns)}",
+                     "col-md#{details_column_width_md(service_details_columns)}",
+                     "col-lg-#{details_column_width_lg(service_details_columns)}"] }
+        - column.map do |group|
+          - if (!group[:clazz].present? && any_present?(service,
+                    *group[:fields])) || (group[:clazz] && service.send(group[:clazz]).present?)
```
JanKapala, line 11: This logic looks a bit complicated for the view, maybe should it be moved to the helper?

JanKapala (changes requested):

JanKapala:
In the browser pages look good

app/helpers/service/details_helper.rb
wujuu: I guess we can refactor that - the name of the method gives away that the method is actually returning columns

JanKapala (approved):

Issue #1628, opened 2020-09-24
Styling backoffice


martaswiatkowska:
Create styling for array elements. 
Especially: 
-multimedia
-use cases
-certifications
-standards
-open source technologies
-grant project names

What it looks like now:
![image](https://user-images.githubusercontent.com/17965571/94139036-4ae89880-fe69-11ea-85c8-adf167c5a3a5.png)

PR #1709, opened 2020-11-30
[#1628] styling backoffice

jarekzet:
Added new mechanism to navigate new resource form to make it easier to fill-in all fields.

New left column to scroll between folding-down cards.


goreck888 (approved):
I think everything is ok, but add changelog entry :)

agpul (approved):
From my point of view, this solution is good enough, but if you have enough time you can consider two enhancements: 
- the left menu should be bigger: 
image
- in case of errors you can highlight the section (sections) with potential problems:
image

agpul:
It's fine for me. If there are no bugs in the source code we can merge this PR.

PR #1719, opened 2020-12-11
[#1712] Offers API

### LINKED ISSUE #1712: Offers API

wujuu:
- [x] Add gem 'simple_token_authentication' - for devise-integrated token authentication - changes in User model
- [x] Add gem 'rswag' - generating swagger docs from specs - configuration in spec/swagger_helper.rb
- [x] Proof of concept of everything working together (devise, pundit, simple_token_authentication, rswag, ActionController::API)
- [x] Add swagger docs (/api-docs)
- [x] Add token controller and views, for generating tokens from GUI (/token)
- [x] /token UI - (CSS, design) @jarekzet 
- [x] Update README on how to use rswag and OpenAPI
- [x] Add gem 'active_model_serializers' - custom json serialization
- [x] Add Offers serialization
- [ ] Do research about possible multiple tokens implementation
- [x] Add schemas for all scenarios in rswag specs
- [x] Add breadcrumbs to token view
- [x] Add API overview docs to /token (under token generation window)
- [x] Specify what subsections we need, and what should be in them (@bwilk, @roksanaer )
- [x] Adjust Offer and Service policy conditions in our API (data_administrator role @goreck888 )
- [x] Move Offer status logic to `Offer::Create` and `Offer::Update` (@goreck888 )
- [x] Make it so only `data_administrator` can access `/token` path (@goreck888 )
- [x] Add everything done to CHANGELOG
- [x] Add JSON:Validator to POST methods
- [x] Add JSON schemas to new `config/schemas` directory
- [x] Fix strings in views to incorporate `gettext`

Service endpoints:
- [x] Specify which fields should be visible to the user (JSON schema, ServiceSerializer) (@bwilk, @roksanaer )
- [x] GET /api/v1/services/ - list services owned by user
- [x] GET /api/v1/services/ rswag spec
- [x] GET /api/v1/services/{service_id} - list specific service {service_id}
- [x]  GET /api/v1/services/{service_id} rswag spec

Offer endpoints:
- [x] Specify which fields should be visible to the user (JSON schema, OfferSerializer) (@bwilk, @roksanaer )
- [x] Specify with which fields should the user be able to create a new offer (@bwilk, @roksanaer )
- [x] Specify with which fields should the user be able to update an offer (@bwilk, @roksanaer )
- [x] GET /api/v1/services/{service_id}/offers - list offers belonging to service {service_id}
- [x] GET /api/v1/services/{service_id}/offers rswag spec
- [x] GET /api/v1/services/{service_id}/offers/{offer_id} - list specific offer {offer_id} belonging to service {service_id}
- [x] GET /api/v1/services/{service_id}/offers/{offer_id} rswag spec
- [x] POST /api/v1/services/{service_id}/offers - Add offer to service {service_id}
- [x] POST /api/v1/services/{service_id}/offers rswag spec
- [x] PUT /api/v1/services/{service_id}/offers/{offer_id} - update offer {offer_id} belonging to service {service_id}
- [x] PUT /api/v1/services/{service_id}/offers/{offer_id} rswag spec
- [x] DELETE /api/v1/services/{service_id}/offers/{offer_id} - destroy offer {offer_id} belonging to service {service_id}
- [x] DELETE /api/v1/services/{service_id}/offers/{offer_id} rswag spec

Side jobs:
- [x] Fix parameters_factory

Closes #1712

### DIFF
--- a/app/controllers/api/v1/services/offers_controller.rb
+++ b/app/controllers/api/v1/services/offers_controller.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require "json-schema"
+
+class Api::V1::Services::OffersController < Api::V1::ApiController
+  before_action :find_and_authorize_service
+  before_action :find_and_authorize_offer, only: [:show, :update, :destroy]
+  before_action :validate_payload, only: [:create, :update]
+  after_action :reindex_offer, only: [:create, :update, :destroy]
+
+  def index
+    render json: policy_scope(@service.offers)
+  end
+
+  def show
+    render json: @offer
+  end
+
+  def create
+    offer_temp = offer_template
+    authorize(offer_temp)
+
+    @offer = Offer::Create.new(offer_temp).call
+
+    if @offer.persisted?
+      render json: @offer, status: 201
+    else
+      render json: { error: @offer.errors.messages }, status: 400
+    end
+  end
+
+  def update
+    template = transform_parameters(permitted_attributes(Offer))
+    if Offer::Update.new(@offer, template).call
+      render json: @offer, status: 200
+    else
+      render json: { error: @offer.errors.messages }, status: 400
+    end
+  end
+
+  def destroy
+    Offer::Destroy.new(@offer).call
+    head :ok
+  end
+
+  private
+    def find_and_authorize_service
+      @service = Service.friendly.find(params[:service_id])
+      authorize @service, :administered_by?
+    rescue ActiveRecord::RecordNotFound
+      render json: { error: "Service #{params[:service_id]} not found" }, status: 404
+    end
+
+    def find_and_authorize_offer
+      @offer = @service.offers.find_by!(iid: params[:id])
+      authorize @offer
+    rescue ActiveRecord::RecordNotFound
+      render json: { error: "Offer #{params[:id]} for service #{params[:service_id]} not found." }, status: 404
+    end
+
+    def validate_payload
+      schema_file = (action_name == "create") ? "offer_input.json" : "offer_update.json"
+      JSON::Validator.validate!(Rails.root.join("swagger", "v1", "offer", schema_file).to_s, params["offer"].as_json)
+    rescue JSON::Schema::ValidationError => e
+      render json: { error: e.message }, status: 400
+    end
+
+    def reindex_offer
+      if @service.offers.size > 1
+        @service.offers.reindex
+      end
+    end
+
+    def offer_template
+      temp = transform_parameters(permitted_attributes(Offer))
+      Offer.new(temp.merge(service: @service, status: :published))
+    end
+
+    def transform_parameters(template)
+      unless template["parameters"].blank?
+        template["parameters"] = Parameter::Array.load(template["parameters"])
+      end
+      template
+    end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  acts_as_token_authenticatable
+
   devise :database_authenticatable, :rememberable, :trackable,
          :omniauthable, omniauth_providers: %i[checkin]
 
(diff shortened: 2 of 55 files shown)


agpul (commented):
image
I recommend removing this statement - this is not true, esspecially for unlogged users.

app/models/user.rb
```diff
@@ -35,4 +37,12 @@ def service_owner?
   def to_s
     full_name
   end
+
+  def valid_token?
+    authentication_token != "revoked" && (not authentication_token.nil?)
```
goreck888, line 42: Maybe change (not authentication_token.nil?) to authentication_token.present?

swagger/v1/swagger.yaml
```diff
@@ -0,0 +1,51 @@
...
+          description: lists a service with friendly slug id
+          content: {}
+        '401':
+          description: denies entry for unknown token
+          content: {}
+        '403':
+          description: doesn't show unowned services
+          content: {}
+servers:
+- url: http://{defaultHost}
```
goreck888, line 42: Use anchor, because {defaultHost} returns just a string and not variable. In this case eg.
url: http://*defaultHost

l45: default: &defaultHost localhost:5000

goreck888 (changes requested):
See my comment about yaml config file and add changelog entry.

app/models/user.rb
wujuu: Done

swagger/v1/swagger.yaml
wujuu: Actually, this is how it's suposed to be - this file is generated by rswag. The idea is to introduce variables for the end user to choose from (variable - {defaultHost}, is visible to the user, and then he can interpolate it with whatever he wants (or values provided by us)). You could see this behviour in /api-docs, but actually not anymore, because I deleted the variable and added 3 different servers (localhost, production, master)

app/controllers/api/v1/services/offers_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  def create
+    offer_temp = offer_template
+    authorize(offer_temp)
+
+    @offer = Offer::Create.new(offer_temp).call
+
+    if @offer.persisted?
+      render json: { msg: "Offer #{@offer.iid} created successfully." }
```
michal-szostak, line 26: generally REST message for CREATE (POST) should return `201` and created object.
REST api should avoid returning messages. If it's successful and no content is to be returned just return `200` or `204`

app/controllers/api/v1/services/offers_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+    else
+      render json: { error: @offer.errors.messages }, status: 400
+    end
+  end
+
+  def update
+    template = transform_parameters(permitted_attributes(Offer))
+    if Offer::Update.new(@offer, template).call
+      render json: { msg: "Offer updated successfully." }
```
michal-szostak, line 35: Same as with POST - an object should be returned

config/routes.rb
```diff
@@ -97,6 +96,23 @@
     end
   end
 
+  unless ENV['OFFERS_API_DISABLED'] == 'true'
```
michal-szostak, line 99: I would advice to put this into application configuration. There you can compare it with string, but in the application itself it would be much better if this flag was a boolean.
Also we should refrain from using ENV[...] directly in the application code.
(it's harder to mock in tests, etc)

michal-szostak (changes requested):

app/controllers/api/v1/services/offers_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      render json: @offer, status: 201
+    else
+      render json: { error: @offer.errors.messages }, status: 400
+    end
+  end
+
+  def update
+    template = transform_parameters(permitted_attributes(Offer))
+    if Offer::Update.new(@offer, template).call
+      render json: @offer, status: 201
```
michal-szostak, line 35: update should not return status `201 created`

michal-szostak (approved):

michal-szostak (approved):

bwilk (approved):

PR #1721, opened 2020-12-17
[#1711] Fix filtering by country

### LINKED ISSUE #1711: Fix the tool for filtering resources by country
**Describe the bug**
Allow the end-user to filter the marketplace resource list by country. The feature was requested by users coming from EOSC Pillar.  

**To Reproduce**
1. Go to https://marketplace.eosc-portal.eu/services
2. Scroll down to the "Provider location" filter
3. Change from Provider location to resource availability
4. Fix the filter

kmarszalek:
- change to use geographical_availabilities
- fix filter
Closes #1711 .

### DIFF
--- a/app/models/filter/location.rb
+++ b/app/models/filter/location.rb
@@ -1,20 +1,21 @@
 # frozen_string_literal: true
 
 class Filter::Location < Filter
-  #   TODO finish this filter
-
   def initialize(params = {})
     super(params: params.fetch(:params, {}),
-          field_name: "location", type: :select,
-          title: "Provider location", index: nil)
+          field_name: "geographical_availabilities",
+          title: "Provider location",
+          type: :select,
+          index: "geographical_availabilities")
   end
 
   private
     def fetch_options
-      [{ name: "Any", id: "" }, { name: "EU", id: "EU" }]
+      [{ name: "Any", id: "" }] + Service.where(status: [:published, :unverified])
+             .pluck(:geographical_availabilities).flatten.uniq.map { |s| { name: s, id: s } }
     end
 
     def where_constraint
-      {}  # TODO finish this filter
+      { @index.to_sym => values.map { |c| Country.dump(Country.find_by_name(c)) } }
     end
 end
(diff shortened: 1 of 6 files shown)


spec/controllers/concerns/searchable_spec.rb
```diff
@@ -275,5 +275,20 @@ def basic_test
         basic_test
       end
     end
+
+    context Filter::Location do
+      let!(:collection) { ["PL", "UK"] }
+      let!(:service1) { create(:service).tap { |s| s.geographical_availabilities = Array.wrap(Country.load(collection.first)); s.save } }
```
wujuu, line 281: You can throw the `gegraphical_availabilities` into the `create` , like 
`create(:service, geographical_availabilities: Array...)`

wujuu (changes requested):

wujuu (commented):
Also found a really small bug. How to recreate:
Create a *draft* service with a country that hasn't been used before.
This country now appears in the filters drop down menu(bug?), and if chosen returns 'no results' (this part is actually correct of course).

spec/controllers/concerns/searchable_spec.rb
kmarszalek: good eye :)
done.

wujuu (commented):
Unfortunately I found another bug regarding the draft service:
In the regular service listing the filter works as intended (doesn't show country used in a draft service).
In the backoffice owned resources (services) listing it also doesn't show the country used in a draft service, even though it should (owner of the service should have the abilitiy to filter by conutry on draft services as well)

wujuu (approved):

app/models/filter/location.rb
```diff
@@ -1,20 +1,20 @@
...
+          field_name: "geographical_availabilities",
+          title: "Provider location",
+          type: :select,
+          index: "geographical_availabilities")
   end
 
   private
     def fetch_options
-      [{ name: "Any", id: "" }, { name: "EU", id: "EU" }]
+      [{ name: "Any", id: "" }] + Service.where(status: :published).pluck(:geographical_availabilities).flatten.uniq.map { |s| { name: s, id: s } }
```
goreck888, line 14: I suppose it should be status: [:published, :unverified] :)

goreck888 (changes requested):

goreck888:
Check my comment, moreover it's ok and I'll approve this PR

goreck888 (approved):

PR #1733, opened 2021-01-04
[HOTFIX] Disable split experiments

goreck888:
Disable split experiment in recommendations view.
May fix possible problem with puma memory use increasing and overflow

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -32,8 +32,7 @@ class ServicesController < ApplicationController
       @client = @client&.credentials&.expires_at.blank? ? Google::Analytics.new : @client
       @analytics = Analytics::PageViewsAndRedirects.new(@client).call(request.path)
     end
-
-    ab_finished(:recommendation_panel, reset: false) if params[:from_recommendation_panel]
+    # TODO: add ab_finished to run split recommendations experiment
   end
 
   private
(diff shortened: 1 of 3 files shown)


kmarszalek (approved):
as far as i know it works, but cant you just delete the code? 
im not o fan of using commented out code. 
issue with puma and ab testing is important. we probably won't put it off so deleting code and good commit msg for me is good enough.

app/controllers/services_controller.rb
```diff
@@ -32,8 +32,7 @@ def show
       @client = @client&.credentials&.expires_at.blank? ? Google::Analytics.new : @client
       @analytics = Analytics::PageViewsAndRedirects.new(@client).call(request.path)
     end
-
-    ab_finished(:recommendation_panel, reset: false) if params[:from_recommendation_panel]
+    # TODO: add ab_finished to run split recommendations experiment
```
JanKapala, line 35: Maybe just comment this line - not remove. Who will be remember that there is an if statement and reset parameter

JanKapala (changes requested):

app/controllers/services_controller.rb
kmarszalek: Github will remember.

JanKapala: but someone will have to check old PRs rather than just uncomment XD

JanKapala (approved):

Issue #1789, opened 2021-02-03
Offers sizes don't match in the UI 


bwilk:
**Describe the bug**
Offers sizes don't match in the UI 

**To Reproduce**
1. Go to a Resource presentation page
2. See offers - the size don't match. 

**Expected behavior**
Size should be the same. Some resources use technical Parameters section in the offers display, some don't. This needs to be taken into account. 

**Screenshots**
![image](https://user-images.githubusercontent.com/2674939/106754983-e49d2f00-662d-11eb-84aa-41b8b96de087.png)

PR #1790, opened 2021-02-03
offer card equal height

jarekzet:
This PR resolves issue #1789

### DIFF
--- a/app/views/services/_offer.html.haml
+++ b/app/views/services/_offer.html.haml
@@ -1,4 +1,4 @@
-.col-md-6{ id: "offer-#{offer.id}" }
+.col-md-6.d-flex.align-items-stretch{ id: "offer-#{offer.id}" }
   .card.m-0.mb-5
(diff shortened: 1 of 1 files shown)


bwilk:
@goreck888 @kmarszalek please help @jarekzet to fix the tests in this PR

kmarszalek:
resolved by re-running. should offers width be equal?

roksanaer (changes requested):
Please rebase this PR to the master version. I am finding a major issue when testing: parameters introduced to the offer are not shown at all in the backoffice view:
image

image
 
Parameters are shown in the user view though:
image

 
Tested service is B2find

Master does not bear such a problem. 

Additionally, would it be possible to extend the description if there is still some space left in the tile? like here:
image

can we show more text instead of 'show more' link?

roksanaer (approved):
After some consultation. approval for the PR is granted

PR #1795, opened 2021-02-08
[HOTFIX] Offers API content fixes

wujuu:


bwilk (approved):
Please update the following API description and fix tests:
`Resource eid or MP slug ->  Resource MP slug (id) or eid`
Otherwise it's fine.

bwilk:
I another issue while reviewing this PR (at least I believe this is not the problem caused with this implementation, but with the UI behavior instead) 

I think the discussion between @wujuu and @goreck888 will give us the whole picture. 

https://github.com/cyfronet-fid/marketplace/issues/1796

bwilk:
Alternative 
`Offer id -> Offer identifier
Resource eid or MP slug ->  Resource identifier - id or eid`

Issue #1782, opened 2021-02-02
Implement Probes for recommendation system


michal-szostak:
> Part of the EPIC #1783

This issue is composed from a couple of sub tasks:

* [x] Research how to implement probes (whether Google Analytics / Matomo style, or custom solution based on stimulus)
* [x] Fetch recommended services
  * [x] Create middleware that call recommender module
  * [x] Create service to call recommender
     * [x] Stub recommended products
  * [x] Create client UID like do it Google Analitics for all users to keep track for devices
  * [x] Fetch panel_id from HTML
  * [ ] Create separate Model to keep Recommender data stored in Redis
* [ ] Create service for calling probe API in recommender module
  * [ ] Disable probation call's in dev environment
  * [ ] call on stimulus event (? alternative is 
  * [ ] Create stimulus controller

PR #1814, opened 2021-02-23
[#1782] Implement Probes for recommendation system

danielkryska:
- Fetch recommended services by search state 
  - Send request on search, filters change
- Record user actions
  - Add attribute tags to HTML elements
  - Add event listener on page load and turbo::before


bwilk:
During the review check whether we are able to operate without recommender system in production

app/controllers/probation_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class ProbationController < ApplicationController
```
michal-szostak, line 3: I'm not sure about the name of the controller - I'd rather change it to ProbeController...

app/javascript/app/probes.js
```diff
@@ -0,0 +1,67 @@
+import Rails from '@rails/ujs'
+
+export default function initProbes(scope = window) {
+    window.document.querySelectorAll("[data-probe]")
+        .forEach(function(element) {
+            element.addEventListener('click', async function () {
+                const visit_time = new Date().getTime();
+                const is_ordered = scope.location.pathname.includes("summary")
```
michal-szostak, line 8: This is a minor, may be omitted - but It's better to have consistent style with string literals in code - I'd advise to pick `'<string>'`

michal-szostak (approved):
Two minors, but I strongly recommend fixing the naming scheme.
I'd like to have @JanKapala to checkout this build and test it manually with recommender instance

Marcelinna:
when choosing a domain, several queries come out at once and there is a problem with execution
image
image

Marcelinna:
image

event is added only to button <_CLEAR ALL>, should it also be added to the trash icon next to the service on the bottom bar? is added to the functionality <_Remove from comparison> which does the same as the trash can icon, where there is no functionality.

Marcelinna:
lack of action in 
image

Marcelinna:
lack of action in 
image
image

Marcelinna:
lack of action in 
image

Marcelinna:
lack of request to /probation in Helpdesk-email
image

PR #1817, opened 2021-02-24
[#1813] fixes for 3.7.0

### LINKED ISSUE #1813: Fixes for 3.7.0
* [x] for the "Recently added resources" query please broaden the scope so it contains services that either have a given provider listed in the "Resource Organisation" or "Resource Providers" field
 * [x] at the moment in the main resource bar:
image
under "provided by" has "resource providers" field
This should be broaden to  the sum of Resource Organisation and Resource Providers. First of all, Resource Providers field is not mandatory.
Second, I've added as a newly created provider a new resource. Underpinned CESNET on the providers list. My newly created provider (listed as a resource organistation) is not connected anywhere on the resource presentation page. for now, until we come up with a better idea, I suggest to underpin the Resource Organistaion field under this link
image
* [x] map works fine now. it includes participating countries field from the provider's profile and it's OK
* [x] not all of the provider's profile fields are visualised in the MP. e.g.
From Other' tab: Hosting Legal Entity, Structure Type, MERIL Scientific Categorisation, Area of Activity, Societal Grand Challenges, National Roadmaps are missing
From basic: legal status
From contact: position of the public contact
From maturity: lifecycle status

this is what is visible:
image

while all fields are filled in in the PP
* [x] restrict "recently added services" quesry to published (or unverified) reslurces only.

goreck888:
[#1813] Fixes for 3.7.0

Fixes #1813


roksanaer (approved):

kmarszalek (approved):
Works, maybe some additional tests?

michal-szostak (approved):

PR #1821, opened 2021-02-26
[#1749] Change title for related_resources in details

### LINKED ISSUE #1749: Change the title "Suggested compatible resources" at the resource "Details" page
### Current behaviour
Currently on the MAIN resource page and the DETAILS  page we see "Suggested compatible resources" heading and the "Related resources" content (the set of resources come from the Related Resources field in the Resource Profile) 

### Expected result:
"Suggested compatible resources" should be displayed at the MAIN page only and the set of resources should remain the same it is right now. Related Resources set should be at the DETAILS page, and the heading should be "Related resources"

### Solution 
Keep the generic parts of the partial generic. If needed split the partial in to two separate ones. Keep the layout of the partials the same.  

#### Current
image

#### Expected 
image

#### What should remain unchanged 
image

goreck888:
Change the title of related resources section
in details and opinions
from "Suggested compatible resources"
to "Related resources"

Closes #1749

### DIFF
--- a/app/views/backoffice/services/preview.html.haml
+++ b/app/views/backoffice/services/preview.html.haml
@@ -27,4 +27,4 @@
   = render "services/tags", service: @service
 
 .container.mt-4.pt-4.related-container
-  = render "services/related", related_services: @related_services
+  = render "services/related", related_services: @related_services, title: @related_services_title
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Changed
 - Geographical availabilities order in the Resource details view (@kmarszalek, @jarekzet)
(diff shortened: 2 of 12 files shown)


app/views/backoffice/services/preview.html.haml
```diff
@@ -27,4 +27,8 @@
   = render "services/tags", service: @service
 
 .container.mt-4.pt-4.related-container
-  = render "services/related", related_services: @related_services
+  - if @related_services&.count&.positive?
+    %h2.mb-4.mt-3.font-weight-light
+      = _("Suggested compatible resources")
+
+    = render "services/related", related_services: @related_services
```
kmarszalek, line 34: add a title: "Suggested compatible resources" here and then use it in removed if from partials

kmarszalek (approved):

roksanaer (approved):

roksanaer (approved):

PR #1835, opened 2021-03-02
[#1796] Differences in the behaviour of the offer UI and API

### LINKED ISSUE #1796: Differences in the behaviour of the offer UI and API
**Describe the bug**
When deleting offers which have orders via API, they are still shown in backoffice / ordering configuration UI 

**To Reproduce**
1. Create 2 offers (via API or UI)
2. Make an order for the second offer
3. Delete second offer with the use of API
4. **See both offers in the ordering configuration UI / backoffice** 

More sophisticated path:
1. Create 2 offers (via API or UI)
2. Make an order for the second offer
3. Delete second offer with the use of API 
4. See both offers in the ordering configuration UI / backoffice 
5. Create another offer (3rd one)
6. Make orders for the 3rd offer
7. With the use of UI (**!**) delete the 3rd offer
8. **See that now the state in the UI and API are synced again**
9. Create another offer 
10. **See the same problem again (more offers in the backoffice then expected**

**Expected behavior**
Same behaviour for API and UI -> offers with orders are archived instead of deletion  

**Screenshots**
Exactly the same time:
image
image

```
[ 
  {
    "id": 1,
    "name": "Opportunistic access",
    "description": "Based on scheduler ",
    "parameters": [
    ....
    ],
    "order_type": "other",
    "webpage": "http://www.cyfronet.pl",
    "internal": false,
    "order_url": ""
  },
  {
    "id": 3,
    "name": "test",
    "description": "terst",
    "parameters": [],
    "order_type": "order_required",
    "webpage": "",
    "internal": false,
    "order_url": ""
  }
]
```

**Additional context**
It seems that most of the app would work fine even if we consider deleting offers possible. Is the current "archiving" behaviour really needed?

kmarszalek:
- show only published offers
- change the status of the offer when deleting to deleted if there is an order
- migrate drafts to deleted

Closes #1796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Differences in the behaviour of the offer UI and API (@kmarszalek)
 
 ## [3.7.0] 2021-02-19
 
--- a/app/controllers/backoffice/services_controller.rb
+++ b/app/controllers/backoffice/services_controller.rb
@@ -23,7 +23,7 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
 
   def show
     @offer = Offer.new(service: @service, status: :draft)
-    @offers = @service.offers.order(:created_at)
+    @offers = @service.offers.published.order(:created_at)
     if current_user&.executive?
       @client = @client&.credentials&.expires_at.blank? ? Google::Analytics.new : @client
       @analytics = Analytics::PageViewsAndRedirects.new(@client).call(request.path)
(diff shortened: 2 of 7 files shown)


roksanaer (approved):

db/schema.rb
```diff
@@ -473,6 +473,7 @@
     t.datetime "last_update"
     t.string "related_platforms", default: [], array: true
     t.datetime "synchronized_at"
+    t.boolean "external", default: false
```
goreck888: Remove changes in schema.rb
External is an old value and now we don't use it.

goreck888 (approved):

PR #1844, opened 2021-03-04
[#1826] Include `resource_organisation` in provider's filter

### LINKED ISSUE #1826: Providers filter include both fields: RO and RP
Problem:
When I've created new providers and a resource for it (in the resource my provider is listed as the RO), my provider does appear as the element in the Providers filter but it has no resources connected:

Solution:

Broaden the elasticsearch  filtering mechanism for the Providers filter so it includes both "Resource Organisation" and "Resource Providers" field 

Please note: the list of elements in the provider filter should include providers registered and approved in the provider component (regardless of the number of underpinned resources)

goreck888:
Filtering by providers includes `resource_organisation`
and `providers` associations

Fixes #1826

### DIFF
--- a/spec/controllers/concerns/searchable_spec.rb
+++ b/spec/controllers/concerns/searchable_spec.rb
@@ -11,10 +11,10 @@ RSpec.describe ApplicationController, type: :controller do
   let!(:providers) { create_list(:provider, 2) }
   let!(:categories) { create_list(:category, 2) }
 
-  let!(:service1) { create(:service, name: "duper super name", providers: providers,
-                           categories: [categories.first], tag_list: "tag1, tag2") }
-  let!(:service2) { create(:service, name: "very different title", providers: [providers.first],
-                           categories: categories, tag_list: "tag2") }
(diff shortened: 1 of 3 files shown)


spec/controllers/concerns/searchable_spec.rb
```diff
@@ -11,10 +11,10 @@
   let!(:providers) { create_list(:provider, 2) }
   let!(:categories) { create_list(:category, 2) }
 
-  let!(:service1) { create(:service, name: "duper super name", providers: providers,
```
kmarszalek, line 14: add to context "filter type tests" section tests filtering by provider and resource organisation

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_02_26_164811) do
```
kmarszalek, line 13: necessary?

kmarszalek (changes requested):
it works on PR instance.

wujuu (approved):

kmarszalek (approved):

PR #1852, opened 2021-03-11
[#1803] Display sync error for errored resources

### LINKED ISSUE #1803: Display information about sync error on the resource presentation page (provider admin only)
Currently we don't give provider admins the possibility to check the PC - UC (MP) sync errors. There is no way for the provider admin to check whether the resource has been successfully propagated to the MP

A notification on the resource presentation page dedicated for the provider admin when the state is errored. 
On top of the error it'd be good to have the reason of the failure recorded and displayed

**Additional context**
Other roles in the PC should also be able to see the error. Some alignement of the authorisation mechanisms should be considered. Maybe:

 1. Sync of the user roles?
 2. The same session storage?

goreck888:
Change logic, now errored resources
are displayed on resources list
Errors are saved in service_source as errored field.

Fixes #1803

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Filtering by providers includes both `resource_organisation` and `providers` associations (@goreck888)
+- Displaying synchronization issues (if any occurs) in resource page for `data_administrators` (@goreck888)
 
 ## [3.8.0] 2021-03-03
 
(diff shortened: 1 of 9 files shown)


Marcelinna:
It is ok, I only wonder where the error description should be displayed,
in /services/egi-cloud-compute or /backoffice/services/egi-cloud-compute, because now is  /services/egi-cloud-compute. 

image
image

Marcelinna (approved):

goreck888:
Added info in the backoffice

kmarszalek (approved):

roksanaer:
the ideal solution is:
the error appears in /services/egi-cloud-compute for logged in users who have admin role or is associated with the provider underpinned as Resource organsiation.

Of course, displaying this info and error in the backoffice as an addition is totally fine

roksanaer (changes requested):
please see the comment above

roksanaer (approved):

PR #1861, opened 2021-03-16
[#1859] Fix externally ordered resources

### LINKED ISSUE #1859: far too many resources claim 'orderable via MP' access type (access wizard related)
**Describe the bug**
statistic panel shows that we have 16 (if i remember correctly) resources orderable via MP. 
Acting as a user in the MP, I find many more resources claiming to br 'possible to order vie MP' when starting resource access wizard
it is possible that resource access path types are not matched properly per newly introduced resource order types.

**To Reproduce**
1. List orderable services (access wizard-wise it will list resources orderable internally(MP)+externally) 
https://marketplace.eosc-portal.eu/services?scientific_domains-filter=&scientific_domains-all=&providers-filter=&providers-all=&target_users-all=&related_platforms-filter=&related_platforms-all=&rating=&order_type=order_required&geographical_availabilities=
61 resources listed 15th March 2021 14:45
2. start accessing each one of them
3. more then 16 will will claim to be orderable via MP 

please check more thoroughly 

**Expected behavior**
Resource access wizard should map properly to newly introduced resource order types and implement the logic connected to it 
internally orderable = orderable order type + no URL
externally orderable = orderable order type + URL exists
open access = fully open access or open access 

@goreck888 please add the logic for 'other' type here as well

**Screenshots**

**Additional context**

goreck888:
Fix conditions for external ordering type
Fix query in executive panel
Add hidden `internal` field to the Offer's form
Fixes #1859

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Functional recommendation enhancements for the sake of better user experience (@wujuu)
 - Removal of duplicated provider entries in the Resource Provider field in the MP (@kmarszalek)
 - Feature highlight opinions panel in the admin section (@goreck888)
+- `Internal ordering` checkbox in offer's form (@goreck888)
 
 ### Changed
 - `Provided by` field in `Popular resources`, `Suggested compatible resources` and `Recently added resources` to `Organisation` (@kmarszalek, @jarekzet)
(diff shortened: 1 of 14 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_02_26_164811) do
```
kmarszalek, line 13: :)

wujuu (approved):

michal-szostak (approved):

bwilk:
@goreck888 please change the button "Order externally" to "Go to the order website"
 
This:
image

Should look that way:
image

bwilk (approved):

Marcelinna:
I checked how many resource can be ordered via EOSC Marketplace from (Order required and other) 75 services. I counted 12, so it is posiible that Marketplace ordering 54 is correct number?
image

Marcelinna (changes requested):
.

Marcelinna (approved):

roksanaer (approved):

Issue #1858, opened 2021-03-15
Filters sent to recommendations endpoint need additonal data


wujuu:
**Describe the bug**
Filters sent to recommender system (to `/recommendations`) bugs:
- Country filter - country sent by itself. We need to also send all of the regions this country is in. Also we need country to be sent in alpha2 format.
- Scientific domains filter - chosen scientific domain is sent by itself. We need to also send all of the scientific_domains descendants (Ancestry gem tree structure)

**To Reproduce bug**
1. Go to '/services'
2. Select 'Natural Sciences' in the 'Scientific Domain' filter
2. Select 'Poland' in the 'Resource Availability' filter
3. Json sent to `/recommendations` looks like this:
`[...]
{'filters': {'geographical_availabilities': 'Poland',
              'scientific_domains': [3],
              'sort': '_score'},
[...]
`

**Expected behavior**

Json sent to `/recommendations` looks like this:
`[...]
{'filters': {'geographical_availabilities': ["PL", "WW", "EU", "AH", "EO"],
              'scientific_domains': [3, 9, 18, 29, 36, 40, 42, 52, 64, 65, 66, 67, 68],
              'sort': '_score'},
[...]
`

**Additional context**
To the country filter problem possible fix is to include `Country.regions_for_country` method
To the scientific domain filter problem possible fix is to include `ScientificDomain.find().descendant_ids`

PR #1862, opened 2021-03-16
[#1858] Send additional filters context data to recommender endpoint

danielkryska:
- Extend resource availability filter with regions
- Extend tree structure like filters to send also child nodes

### DIFF
--- a/app/controllers/concerns/service/recommendable.rb
+++ b/app/controllers/concerns/service/recommendable.rb
@@ -5,6 +5,12 @@ require "net/http"
 module Service::Recommendable
   extend ActiveSupport::Concern
 
+  @@filter_param_transformers = {
+    geographical_availabilities: -> name { Country.convert_to_regions_add_country(name) },
+    scientific_domains: -> ids { ids.map(&:to_i) + ids.map { |id| ScientificDomain.find(id).descendant_ids }.flatten },
+    category_id: -> slug { Category.find_by(slug: slug).id }
+  }
+
   included do
     before_action only: :index do
       @params = params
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/service/recommendable.rb
```diff
@@ -5,6 +5,11 @@
 module Service::Recommendable
   extend ActiveSupport::Concern
 
+  @@filter_param_transformers = {
+    geographical_availabilities: -> name { Country.convert_to_regions_add_country(name) },
+    scientific_domains: -> ids { ids + ids.map { |id| ScientificDomain.descendants_of(id).map(&:id) }.flatten }
```
wujuu, line 10: I think you can replace `map(&:id)` with `pluck(:id`). It may be more efficient or use `descendant_ids` method

app/models/scientific_domain.rb
```diff
@@ -19,6 +19,11 @@ def self.names
     all.map(&:name)
   end
 
+  def self.descendants_of(id)
```
wujuu, line 22: You can just use `descendant_ids` method from ancestry. https://github.com/stefankroes/ancestry#tree-navigation. I think there is no need to rewrite this method

app/controllers/concerns/service/recommendable.rb
wujuu, line 10: You can just use `descendant_ids` method from ancestry. https://github.com/stefankroes/ancestry#tree-navigation. Like `ScientificDomain.find(id).descendant_ids

wujuu, line 10: Example  format of the scientific_domains is ['3', 9, 18, 29, 36, 40, 42, 52, 64, 65, 66, 67, 68] - Every item should be an `int`

app/controllers/concerns/service/recommendable.rb
```diff
@@ -81,27 +86,12 @@ def get_filters_by(params)
...
-            filters[key] = value
-            next
-          end
+          is_empty_value = value.nil? || value == "" || value == []
+          next if key == "controller" || key == "action" || is_empty_value
 
-          next if value.nil? || value == "" || value == []
-          if !key.match?(/[a-z_]-filter/) && !key.match?(/[a-z_]-all/)
-            filters[key] = value
+          filters[key] = value
```
wujuu, line 92: `category_id` filter is a string, it should be an int as before

wujuu (changes requested):

wujuu (approved):

app/controllers/concerns/service/recommendable.rb
```diff
@@ -81,27 +87,12 @@ def get_filters_by(params)
...
-
-          if key.match?(/[a-z_]-filter/)
-            key = key.sub "-filter", ""
-            value = params[key]
-
-            next if value.nil? || value == "" || value == []
-            filters[key] = value
-            next
-          end
+          is_empty_value = value.nil? || value == "" || value == []
```
goreck888, line 90: You can use just value.blank?

goreck888 (dismissed):

goreck888 (approved):

danielkryska:
Will be contiuned in PR #1866

PR #1906, opened 2021-03-23
[HOTFIX] Fix logic of resource order wizards

goreck888:
Change logic:
external if offer is `order_required && !internal && order_url.present?`
internal in any other case of `order_required` offer
(it means `!internal && order_url.blank?` or just `internal`)
Executive statistics have unified query

In PRODUCTION data there should be only 3 services with external flag:
["InAcademia", "GÉANT Clouds Service ", "GÉANT Testbeds Service"]

### DIFF
--- a/app/models/usage_report.rb
+++ b/app/models/usage_report.rb
@@ -56,13 +56,13 @@ class UsageReport
     def service_count_by_order_type(*types, internal: false)
       unless internal
         Service.joins(:offers).where(offers: { order_type: types, status: :published },
-                                     status: [:published, :unverified]).uniq.count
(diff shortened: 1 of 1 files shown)


goreck888:
@wziajka please load anonymized production data on this instance, we have to check if behavior and statistics are ok.

Marcelinna (approved):

kmarszalek (approved):
generally, i would like to see more spec to test all conditions. but as it is a hotfix..
@michal-szostak is it ok?

michal-szostak (approved):

PR #1912, opened 2021-03-24
[#1879] Add OMS to DB

### LINKED ISSUE #1879: Add OMS to DB
Create a table with OMSes (Order Management Systems) and populate it with SOMBO, i.e. type=global.

The table should have the following columns: id, type=(global|provider), provider_id (nullable), webhook_url (nullable).
For type=global the provider_id=null.
There can be only a single global OMS.

Apart from that, each OMS should have a set of custom_offer_fields. These are custom fields, which have to be specified for each offer which uses this OMS, and they are passed on to the project item's primary OMS.
Such field should have the following columns: name, description, default.

wujuu:
Closes #1879 

**What has been changed**: 

- Add OMS, OmsAministration, OmsProviders and OmsOffer models
- Add primary_oms to Offer, add validations
- Add default OMS logic
- Add SOMBO OMS (along with its admin) creation migration
- Add OMS/Offer model specs

### DIFF
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -33,15 +33,15 @@ class Offer < ApplicationRecord
     OfferSerializer.new(self).as_json
   end
 
+  counter_culture :service, column_name: proc { |model| model.published? ? "offers_count" : nil },
+                  column_names: {
+                    ["offers.status = ?", "published"] => "offers_count"
+                  }
+
   enum status: STATUSES
 
   belongs_to :service
-
-  counter_culture :service, column_name: proc { |model| model.published? ? "offers_count" : nil },
-                  column_names: {
-                      ["offers.status = ?", "published"] => "offers_count"
-                  }
-
+  belongs_to :primary_oms, class_name: "Oms", optional: true
   has_many :project_items,
            dependent: :restrict_with_error
 
@@ -49,6 +49,14 @@ class Offer < ApplicationRecord
   validates :service, presence: true
   validates :iid, presence: true, numericality: true
   validates :status, presence: true
+
+  validates :oms_params, absence: true, if: -> { current_oms.blank? }
+  validate :check_oms_params, if: -> { current_oms.present? }
+
+
+  def current_oms
+    primary_oms || Oms.find_by(default: true)
+  end
 
   def to_param
     iid.to_s
@@ -62,4 +70,24 @@ class Offer < ApplicationRecord
     def offers_count
       service && service.offers.maximum(:iid).to_i || 0
     end
+
+    def oms_params_match?
+      if (Set.new(oms_params.keys) - Set.new(current_oms.custom_params.keys)).length > 0
+        errors.add(:oms_params, "additional unspecified keys added")
+        return
+      end
+
+      missing_keys = Set.new(current_oms.custom_params.keys) - Set.new(oms_params.keys)
+      if (missing_keys & Set.new(current_oms.mandatory_defaults.keys)).length > 0
+        errors.add(:oms_params, "missing mandatory keys")
+      end
+    end
+
+    def check_oms_params
+      if current_oms.custom_params.present?
+        oms_params.blank? ? errors.add(:oms_params, "can't be blank") : oms_params_match?
+      else
+        errors.add(:oms_params, "must be blank if primary oms' custom params are blank") if oms_params.present?
+      end
+  end
 end
--- a/app/models/oms.rb
+++ b/app/models/oms.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+class Oms < ApplicationRecord
+  has_many :oms_administrations, dependent: :destroy
+  has_many :administrators,
+           through: :oms_administrations,
+           source: :user,
+           class_name: "User"
+  has_many :oms_providers, dependent: :destroy
+  has_many :providers, through: :oms_providers
+  has_many :offers, foreign_key: :primary_oms_id, dependent: :nullify
+  belongs_to :service, optional: true
+
+  self.inheritance_column = nil
+  enum type: {
+    global: "global",
+    provider_group: "provider_group",
+    resource_dedicated: "resource_dedicated"
+  }
+
+  validates :name, presence: true, uniqueness: true
+  validates :type, presence: true, inclusion: { in: types }
+
+  validates_associated :service, if: :resource_dedicated?
+  validates :service, presence: true, if: :resource_dedicated?
+  validates :providers, absence: true, if: :resource_dedicated?
+
+  validates_associated :providers, if: :provider_group?
+  validates :service, absence: true, if: :provider_group?
+  validates :providers, presence: true, if: :provider_group?
+
+  validates :service, absence: true, if: :global?
+  validates :providers, absence: true, if: :global?
+
+  validate :single_default_oms?, if: :default?
+
+  validate :validate_custom_params, if: :custom_params?
+
+  def mandatory_defaults
+    custom_params&.filter { |_, v| v["mandatory"] }&.transform_values { |v| v["default"] }
+  end
+
+  private
+    def single_default_oms?
+      if Oms.all.pluck(:default).any?
+        errors.add(:default, "there can't be more than one default OMS")
+      end
+    end
+
+    def validate_custom_params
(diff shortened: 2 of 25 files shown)


db/schema.rb
```diff
@@ -474,6 +506,7 @@
     t.datetime "last_update"
     t.string "related_platforms", default: [], array: true
     t.datetime "synchronized_at"
+    t.boolean "external", default: false
```
jswk, line 509: Probably for removal

jswk (approved):

app/models/offer.rb
```diff
@@ -50,6 +51,28 @@ def as_json(options = nil)
   validates :iid, presence: true, numericality: true
   validates :status, presence: true
 
+  validates_associated :primary_oms
+  validates :primary_oms, presence: true
+  validate :check_oms_params, if: -> { primary_oms.present? }
```
jswk, line 56: When primary_oms is required then the if here isn't necessary.

app/models/offer.rb
```diff
@@ -50,6 +51,28 @@ def as_json(options = nil)
...
+
+  before_validation :set_sombo # TODO: Should we do it here or at all?
```
jswk, line 58: Setting SOMBO before_validation seems fine by me. Nevertheless, I wouldn't create it in the set_sombo method.
When you need it set in tests you could have it created in some before_each hook (maybe in support)? In production it should always be present, and no optional creation should be necessary.

app/models/offer.rb
```diff
@@ -36,6 +36,7 @@ def as_json(options = nil)
   enum status: STATUSES
 
   belongs_to :service
+  belongs_to :primary_oms, class_name: "Oms", foreign_key: :oms_id
```
jswk, line 39: Only thing is I would consider renaming this column in DB to primary_oms_id, so that you don't have to specify foreign_key explicitly here and it would make schema.rb a little clearer.

app/models/offer.rb
```diff
@@ -62,4 +84,12 @@ def set_iid
     def offers_count
       service && service.offers.maximum(:iid).to_i || 0
     end
+
+    def set_sombo
+      if primary_oms.blank?
+        self.primary_oms = Oms.find_by(name: "SOMBO")
```
bwilk, line 90: I don't like this solution. Whitelabel Marketplaces will hate this. Are you able to move such custom logic to the app configuration and make it self-explanatory though generic?

bwilk: Sth like "default", "catch-all" OMS per MP instance would be the proper term to describe this use case in a generic way

app/models/oms.rb
```diff
@@ -0,0 +1,54 @@
...
+  validates :providers, presence: true, if: :provider_group?
+
+  validates :service, absence: true, if: :global?
+  validates :providers, absence: true, if: :global?
+
+  validate :custom_params_unique?, if: :custom_params?
+  validate :custom_params_valid_keys?, if: :custom_params?
+
+
+  # TODO: before_destroy: ensure associated offers have set their primary_oms to SOMBO!!!!
```
bwilk, line 39: Same here as above - programming with magic strings https://en.wikipedia.org/wiki/Magic_string

app/models/oms.rb
```diff
@@ -0,0 +1,54 @@
...
+  validates :providers, absence: true, if: :resource_dedicated?
+
+  validates_associated :providers, if: :provider_group?
+  validates :service, absence: true, if: :provider_group?
+  validates :providers, presence: true, if: :provider_group?
+
+  validates :service, absence: true, if: :global?
+  validates :providers, absence: true, if: :global?
+
+  validate :custom_params_unique?, if: :custom_params?
```
bwilk, line 35: Maybe add "just a little" documentation about "custom_params" what are they for. This is obvious at this stage but we'll loose that knowledge (especially the use of "oms_params" with offers)

lib/ordering_api/add_sombo.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module OrderingApi
+  class AddSombo
+    def initialize; end
+
+    def call
+      # TODO: Setting uid as a constant for now to take advantage of find_or_create_by method
```
bwilk, line 8: Are you sure this should be put in the module initializer?

lib/ordering_api/add_sombo.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      sombo = Oms.find_or_create_by(name: "SOMBO") do |oms|
+        oms.type = :global
+        oms.custom_params = ["order_target"]
+        oms.administrators = [sombo_admin]
+      end
+
+      Offer.all.each do |o|
+        o.update(primary_oms: sombo, oms_params: { order_target: o.service.order_target }) if o.primary_oms.blank?
```
bwilk, line 22: this looks like a migration candidate

bwilk (commented):
In general the code looks neat. Let's discuss the comments mentioned as my feedback and I think we're ready to go.

lib/ordering_api/add_sombo.rb
wujuu: OrderingApi::AddSombo.new.call is in fact used in migrations. I implemented it as a task just in case, and it can be run as task after every deploy (I don't think all of the migrations are run every time the code is deployed, just the new ones - correct me if I'm wrong)

lib/ordering_api/add_sombo.rb
wujuu: By this you mean this comment? Or setting the uid as a constant?  Or adding the SOMBO admin user?

app/models/oms.rb
wujuu: Where should we put this documentation? As a comment? Or into readme?

lib/ordering_api/add_sombo.rb
bwilk: as a task it should land in /lib/taks

bwilk: yeah "migration" was unprecise. Task is the way to go. I didn't understand the reason for this class as it was directly in the /lib/xyz

app/models/oms.rb
bwilk: common approach is to document with a cooment before accessor 
https://rubydoc.info/gems/yard/file/docs/GettingStarted.md#documenting-attributes

app/models/offer.rb
jswk: It seems sensible to have a single designated OMS. Wouldn't it be sufficient to designate it by name, such as it's done now? The name could be made configurable by a property for example.
A column "default" or "catch_all" would only be true for a single record.

app/models/oms.rb
jswk: Configuring the protected default OMS by a property may be the way to go?

spec/models/offer_spec.rb
```diff
@@ -26,6 +29,29 @@
...
+  context "OMS dependencies" do
+    it "validates properly against primary oms" do
+      oms = build(:oms, custom_params: { a: { mandatory: false }, b: { mandatory: true, default: "XD" } })
+
+      expect(build(:offer, oms_params: { a: 1, b: 2 }, primary_oms: oms)).to be_valid
+      expect(build(:offer, oms_params: { b: 1 }, primary_oms: oms)).to be_valid
+
+      expect(build(:offer, oms_params: { a: 1 }, primary_oms: oms)).to_not be_valid
+      expect(build(:offer, oms_params: { a: 1, b: 1, c: 1 }, primary_oms: oms)).to_not be_valid
+      expect(build(:offer, oms_params: { b: 1, c: 1 }, primary_oms: oms)).to_not be_valid
```
jswk, line 41: I'd verify against `nil` and  `{ }` `oms_params` too.

app/models/oms.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  private
+    def single_default_oms?
+      if Oms.all.pluck(:default).any?
+        errors.add(:default, "there can't be more than one default OMS")
+      end
+    end
+
+    def validate_custom_params
+      unless custom_params.values.map { |param| JSON::Validator.validate(CUSTOM_PARAMS_SCHEMA, param) }.all?
```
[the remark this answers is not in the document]
wujuu: I had no idea you can do that, good call

jswk (approved):

PR #1915, opened 2021-03-25
Ordering API

wujuu:
This PR contains changes from [EOSC MP ordering API](https://github.com/orgs/cyfronet-fid/projects/11).

In particular:
- ordering API endpoints
- OMS table and its relation to offers, etc.
- Events table
- generalization of the project_item.status
- addition of message scope and author_role
- addition of project_item.user_secrets (with voucherable handling)

The functionality region to test for regressions is mostly around service ordering (status handling, voucherable).
See that the API description is correct too (it should be accessible under a group different than the offerings API).

This PR is to be rebased (or merged) into the master, not squashed.
Note that the tests may fail despite retry working as expected. It's only a single failing test case though and can be disregarded.


bwilk:
Swager documentation changes proposal:
"Ordering API" -> "API for Order Management Systems to integrate with EOSC Marketplace ordering process"

Marcelinna:
GET https://marketplace-10.docker-fid.grid.cyf-kr.edu.pl/api/v1/oms/1/projects

If I created project Customer typology  = Representing a research community, or Representing a project, request return 500 error.
image

Marcelinna:
I wonder if API shouldn't take into account fields like: Company name *, Company website url *, when project has type: Representing a private company.
image
image

Marcelinna:
I have project id 1421 with 3 added services (2 open access and 1 order required)
image

When I try execute GET ​/api​/v1​/oms​/{oms_id}​/projects​/{p_id}​/project_items
I have 500 error
image


It works when in the project I have only required order service
image

Marcelinna:
Something that is not obvious to me: I create project (id = 1425) with two services.
After GET ​/api​/v1​/oms​/{oms_id}​/projects​/{p_id}​/project_items, I see that project has two project_Items and I see two id 1711, 1712.
image

image

but If I want execute GET ​/api​/v1​/oms​/{oms_id}​/projects​/{p_id}​/project_items​/{pi_id} for fist 100 Percent IT Trusted Cloud I have to use id = 1, I think my first attempt was to id=1711
image

jswk:
> I wonder if API shouldn't take into account fields like: Company name *, Company website url *, when project has type: Representing a private company, there arent visible in body response. They are to be fields equal to fields in JIRA?

Possibly yes, but for now the fields mostly just match the ones which were generated and passed to Jira in `app/models/jira/client.rb`, and the company fields aren't mapped there.

Marcelinna:
Api save only one category, but services can belong to several categories
image
image

jswk:
> Api save only one category, but services can belong to several categories

It's expected. This is how this mechanism works in legacy Jira integration.
image

Marcelinna (approved):

Issue #1881, opened 2021-03-19
Generalize project_item status


jswk:
As different OMSes can have different order handling workflows, don't enforce a particular set of statuses.
SOMBO will anyways take part in ordering process as operations team, so a way of mapping statuses is nevertheless required.
To facilitate it, store both a status and its type (current set of possible statuses, i.e. rejected, waiting_for_response, registered, in_progress, ready, closed, approved) for current and historical statuses.

PR #1917, opened 2021-03-25
[#1881] Generalize project_item status

jswk:
Make project_item have a status and a status_type.
The current status is renamed to status_type and it is kept as enum,
so that SOMBO-JIRA can relate to it.
Other than that, a status will now be managed by the primary OMS - it
just has to specify the associated status_type as well.

For now, keep the current SOMBO-JIRA integration, mapping
the status_type to status directly, ensuring it continues to work.

Purge forgotten project_item_changes model from the DB and factories.

Closes: #1881.

### DIFF
--- a/app/models/project_item.rb
+++ b/app/models/project_item.rb
@@ -7,7 +7,7 @@ class ProjectItem < ApplicationRecord
   include Iid
   include Offerable
 
-  STATUSES = {
+  STATUS_TYPES = {
     created: "created",
     registered: "registered",
     in_progress: "in_progress",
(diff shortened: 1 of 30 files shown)


app/models/project_item.rb
```diff
@@ -39,6 +39,7 @@ class ProjectItem < ApplicationRecord
 
   validates :offer, presence: true
   validates :status, presence: true
+  validates :status_type, presence: true
```
wujuu, line 42: Can add `validates :type, inclusion: { in: status_types }`

wujuu (approved):

app/models/project_item.rb
wujuu: And then the validation spec for it

jswk: It isn't necessary, when an enum with a matching name is defined on the class then rails will validate the column against it.

PR #1932, opened 2021-03-29
Temporary fix for Capybara random tests failure

danielkryska:
- repeat test cases 3 times until they pass or return failure

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -17,6 +17,7 @@ require "rspec/rails"
 require "action_dispatch/system_testing/server"
 ActionDispatch::SystemTesting::Server.silence_puma = true
 require "action_dispatch/system_test_case"
+require "rspec/repeat"
 
 # Requires supporting ruby files with custom matchers and macros, etc, in
 # spec/support/ and its subdirectories. Files matching `spec/**/*_spec.rb` are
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -76,6 +77,12 @@
 
   # Use mock of middleware
   config.include Devise::Test::ControllerHelpers, type: :controller
+
+  # Re-run tests on failure to prevent random failures
+  config.include RSpec::Repeat
+  config.around :each do |example|
```
wujuu, line 83: Rerunning undeterministic tests and hoping they would pass seems kind of sloppy but I guess it's OK as a quick fix for now

spec/rails_helper.rb
```diff
@@ -76,6 +77,12 @@
...
+    repeat example, 4.times, verbose: true
```
wujuu, line 84: Also, are you running EVERY example 4 times? Not just the capybara ones?

wujuu (dismissed):

spec/rails_helper.rb
danielkryska: As in the title `Temporary fix for Capybara random tests failure`, it's only an ad hoc solution. Although I suppose that the fix is going to stay with us for longer.

spec/rails_helper.rb
danielkryska: It's running strictly 3 times with normal capybara run.

spec/rails_helper.rb
danielkryska: I understand your worries about the made solution, but please notice that we're not living in a perfect world. Running failing test cases multiple times is a common solution of companies, even when those have perfectly working automatic QA pipes.

michal-szostak: I do agree @wujuu , but we desperately need some kind of fix for this. We'll be working on a better solution in the next sprint.

michal-szostak (approved):

PR #1936, opened 2021-03-29
Add author_role and scope to Message

jswk:
The author_role extends the Message typology, so that Messages
which come from providers and ops-team can be differentiated
(author_role=user|provider|mediator).
On the other hand, scope allows the providers and ops-team to exchange
Messages private to user between OMSes (scope=internal) and allows
providers to send Messages privately to user (scope=user_direct).

Before, when a Message wasn't sent by an MP user, it only had the
author_id==nil. I take it into account when migrating data.

Yet unused, but I add author_name and author_email columns, so that
OMSes can provide them when creating messages, making it easier for user
and other participating providers and ops-team members to discern the
Message origin.

Update the conversation_controllers, so that only scope_public? and
scope_user_direct? are shown to users. Also, make sure the messages are
always ordered by creation time (for project_items the order was by
updated time, inconsitent with the project level).

### DIFF
--- a/app/controllers/projects/conversations_controller.rb
+++ b/app/controllers/projects/conversations_controller.rb
@@ -15,8 +15,8 @@ class Projects::ConversationsController < ApplicationController
       permitted_attributes(Message)
         .merge(
           author: current_user,
-          author_role: "user",
-          scope: "public",
+          author_role: :user,
+          scope: :public,
           messageable: @project
         )
     )
--- a/app/controllers/projects/conversations_controller.rb
+++ b/app/controllers/projects/conversations_controller.rb
@@ -6,8 +6,8 @@ class Projects::ConversationsController < ApplicationController
   def show
     @message = Message.new(messageable: @project)
 
-    load_messages
-    load_projects
+    load_messages!
+    load_projects!
   end
 
   def create
@@ -25,18 +25,18 @@ class Projects::ConversationsController < ApplicationController
       flash[:notice] = "Message sent successfully"
       redirect_to project_conversation_path(@project)
     else
-      load_messages
-      load_projects
+      load_messages!
+      load_projects!
       render :show, status: :bad_request
     end
   end
 
   private
-    def load_projects
+    def load_projects!
       @projects = policy_scope(Project).order(:name)
     end
 
-    def load_messages
+    def load_messages!
       @messages = Message.where(
         messageable_id: @project.id,
         scope: %w[public user_direct],
--- a/app/controllers/projects/conversations_controller.rb
+++ b/app/controllers/projects/conversations_controller.rb
@@ -37,9 +37,6 @@ class Projects::ConversationsController < ApplicationController
     end
 
     def load_messages!
-      @messages = Message.where(
-        messageable_id: @project.id,
-        scope: %w[public user_direct],
-      ).order(:created_at)
+      @messages = policy_scope(@project.messages).order(:created_at)
     end
 end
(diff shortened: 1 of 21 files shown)


app/controllers/projects/conversations_controller.rb
```diff
@@ -4,23 +4,42 @@ class Projects::ConversationsController < ApplicationController
...
   def create
-    @message = Message.
-                new(permitted_attributes(Message).
-                    merge(author: current_user, messageable: @project))
+    @message = Message.new(
+      permitted_attributes(Message)
+        .merge(
+          author: current_user,
+          author_role: "user",
+          scope: "public",
```
bwilk, line 19: please use symbols instead of strings:
- ` :this_is_a_symbol`
- "this is a string" :p

app/controllers/projects/conversations_controller.rb
```diff
@@ -4,23 +4,42 @@ class Projects::ConversationsController < ApplicationController
...
-      @messages = @project.messages
-      @projects = policy_scope(Project).order(:name)
+      load_messages
+      load_projects
       render :show, status: :bad_request
     end
   end
+
+  private
+    def load_projects
```
wujuu, line 35: I've noticed a pattern in MP when we name these kind of methods `find_and_authorize_projects` or `find_projects`

app/controllers/projects/conversations_controller.rb
```diff
@@ -4,23 +4,42 @@ class Projects::ConversationsController < ApplicationController
...
+      @projects = policy_scope(Project).order(:name)
+    end
+
+    def load_messages
```
wujuu, line 39: As above

app/controllers/projects/conversations_controller.rb
```diff
@@ -4,23 +4,42 @@ class Projects::ConversationsController < ApplicationController
...
+      @messages = Message.where(
+        messageable_id: @project.id,
+        scope: %w[public user_direct],
+      ).order(:created_at)
```
wujuu, line 43: This should be in policy_scope - especially the `scope` argument filtering

app/controllers/projects/services/conversations_controller.rb
```diff
@@ -31,6 +37,9 @@ def load_projects
     end
 
     def load_messages
-      @messages = @project_item.messages.order(:updated_at)
+      @messages = Message.where(
```
wujuu, line 40: This should be in policy scope

app/controllers/projects/conversations_controller.rb
jswk: Can do! Why preference of symbols over strings though? It's not like they're enums anyway (i.e. you can still assign incorrect value).

app/controllers/projects/conversations_controller.rb
jswk: The `find_sth` pattern is only used when you load a single object, i.e. `find_project`, but not `find_projects`.
The `load_sth` name is actually used throughout the project:
image
Nevertheless, it seems that its name should end with exclamation mark, as it changes object state (if it only returned the projects it would have no ! at the end).

app/controllers/projects/conversations_controller.rb
jswk: See https://github.com/cyfronet-fid/marketplace/pull/1936#discussion_r603905098.

wujuu (approved):

PR #1945, opened 2021-03-30
Upgrade rails to 6.0.3.6

jswk:
In the previous version 6.0.3.2, there was problematic dependency to
marcel-0.3.3, which was resolved in 6.0.3.6.

The problem manifested itself when doing bundle install with no
prior-cache for the package:
 > Downloading marcel-0.3.3 revealed dependencies not in the API or the
 > lockfile (mimemagic (~> 0.3.2)).


kmarszalek (approved):
wait for CI with merging

PR #1954, opened 2021-04-01
[HOTFIX] Fix set upstream for providers

goreck888:
Fixes REST and JMS importers should set upstream for providers
exactly like it is in the services


db/schema.rb
```diff
@@ -572,6 +572,16 @@
     t.index ["user_id"], name: "index_user_scientific_domains_on_user_id"
   end
 
+  create_table "user_services", force: :cascade do |t|
```
kmarszalek, line 575: ?

kmarszalek (changes requested):
faulty migration

db/schema.rb
goreck888: Done.

Marcelinna (approved):

kmarszalek (approved):

PR #1974, opened 2021-04-08
[#1966] Expose user_secrets in ordering API

jswk:
The user_secrets hash values are obfuscated, excluding the ones which
were updated by the call (i.e. in method update).

Setting `user_secrets: {}` in policy's permitted_attributes allows any
key.

The obfuscation is done in controller on purpose, to allow
non-obfuscation of some keys in case of update.

Closes: #1966.

### DIFF
--- a/app/controllers/api/v1/oms/projects/project_items_controller.rb
+++ b/app/controllers/api/v1/oms/projects/project_items_controller.rb
@@ -6,28 +6,22 @@ class Api::V1::Oms::Projects::ProjectItemsController < Api::V1::Oms::ApiControll
   before_action :validate_payload, only: :update
 
   def index
-    # TODO: Obfuscate user secrets
     @from_id = params[:from_id].present? ? params[:from_id] : 0
     @limit = params[:limit].present? ? params[:limit] : 20
     load_project_items
-    render json: { project_items: @project_items.map { |pi| OrderingApi::V1::ProjectItemSerializer.new(pi) } }
+    render json: { project_items: @project_items.map { |pi| serialize_with_obfuscation(pi) } }
   end
 
   def show
-    # TODO: Obfuscate user secrets
-    render json: OrderingApi::V1::ProjectItemSerializer.new(@project_item).as_json
+    render json: serialize_with_obfuscation(@project_item)
   end
 
   def update
-    # TODO: add user_secrets update option
     attributes = permitted_attributes(@project_item)
 
-    transformed = Hash.new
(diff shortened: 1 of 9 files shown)


app/controllers/api/v1/oms/projects/project_items_controller.rb
```diff
@@ -60,4 +54,31 @@ def validate_payload
...
+      transformed
+    end
+
+    def serialize_with_obfuscation(project_item, excluded = [])
+      obfuscate_user_secrets(OrderingApi::V1::ProjectItemSerializer.new(project_item).as_json, excluded)
+    end
+
+    def obfuscate_user_secrets(payload, excluded)
+      payload[:user_secrets] = payload[:user_secrets].map { |k, v|
+        if excluded.include?(k)
```
wujuu, line 76: I guess we could use the https://www.rubyguides.com/2019/10/ruby-ternary-operator/ operator here - just a syntax sugar though

app/controllers/api/v1/oms/projects/project_items_controller.rb
```diff
@@ -60,4 +54,31 @@ def validate_payload
...
     end
+
+    def transform(attributes)
+      transformed = Hash.new
+      if attributes[:status].present?
+        transformed[:status] = attributes[:status][:value]
+        transformed[:status_type] = attributes[:status][:type]
+      end
+      if attributes[:user_secrets].present?
+        transformed[:user_secrets] = @project_item.user_secrets.merge(attributes[:user_secrets])
```
wujuu, line 65: It would be cool if we could update only the keys that were actually sent by the OMS - but user_secrets is a jsonb field and I think the update method would just overwrite the old values. Google also says it's kinda risky: https://mikerogers.io/2019/06/11/when-to-use-jsonb-columns-in-rails. If it can't be done any other way I would just leave it as is

wujuu:
**IMPORTANT** Additionally we should check if the user secrets field is obfuscated in the events (action: update, eventable: message)

app/controllers/api/v1/oms/projects/project_items_controller.rb
jswk: Yup, if I assign only the sent hash then the old values get deleted. I imagined that activerecord does some kind of locking in such concurrent case... If it doesn't then to be safe we should have changes all throughout the app to avoid such risk.

wujuu (approved):

Issue #1885, opened 2021-03-19
E2E ordering API + adapter test


jswk:
Create a set of integration (cypress?) tests, which are going to verify full user path during the ordering process.

---
**Basic cypress setup - app quality improvement #1978**
App instance: https://marketplace-beta.docker-fid.grid.cyf-kr.edu.pl
JIRA instance: https://jira.eosc-hub.eu/projects/EOSCSOSTAGING/summary

- [x] setup cypress
ref: https://docs.cyfronet.pl/display/FID/Proces+QA?focusedCommentId=88244635#comment-88244635
  - [x] setup profiles in code
  - [x] Add possibility to create typescript commands
  - [x] provide element coverage (highlight elements on which have been made action)
  - [x] on failure, output error logs to terminal 
  - [ ] ~~add the possibility to check if async assertions have been made~~ (only for JS version)
- [x] Github actions setup for e2e (if cases will take too long add a possibility to trigger it manually on PR)
---
**General**
- [x] Move e2e catalog to test/system due to [Ruby guides](https://guides.rubyonrails.org/testing.html#system-testing)

**Utilities**
- [x] get UUID v4
- [x] get rand string
- [x] get rand email
- [x] get rand url
- [x] check captcha
- [x] force visit in another origin
- [x] check if element exists in body without test failure

**Auth commands**
- [x] Login with auth mock
- [ ] Login with AAI
- [x] Logout
- [x] Expand login with firstName, lastName

**JIRA commands**
- [x] Login
- [x] Logout
- [x] Get ticket
- [x] Find ticket
- [x] Get ticket detail (i. e. type)
- [x] Get ticket comments
  - [x] Get comment date
  - [x] Get comment text

**MP app tests**
- [x] Create new project
  - [x] Check project details after creation (prevent cookies clear)
  - [x] Check epic in JIRA

--- 

- [x] implement basic tests based on function provided commands
Important fields in JIRA tickets: SO-1, ServiceOrderTarget
  - [x] Add projects with other types of resources (open access, fully open access, ...)
  - [x] New comment in JIRA epic should be displayed in the app created project x 2
  - [x] Add your first source when in the project.
  - [x] User should order resource.
  - [x] Change the state of the ticket on JIRA, so it can be visible in the order history.
- [ ] Mock production services responses
- [ ] Use `factory.ts`
- [ ] Refactor offers tests
- [ ] Create function getCookies(...keys)

---
**Master/beta env setup**
[...]

PR #1978, opened 2021-04-09
[#1885] Basic cypress setup - app quality improvement

danielkryska:
- Developer can use profiles depend on a server instance
- Developer can create typescript commands
- Developer will see touched elements on site: have a pink border
- On failure, print all logs in terminal: including XHR body and response
- Run e2e in Pull Request Github actions
  - Cache npm packages


.gitignore
```diff
@@ -4,6 +4,12 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile '~/.gitignore_global'
 
+# Ignore e2e miscs
+/e2e/.tool-versions
```
michal-szostak, line 8: I'm relatively sure that `.tool-versions` should be in the repository

michal-szostak (approved):
There is one minor case with ignoring `.tool-versions` which I think should be left out of `.gitignore` (but it can be disputed)

PR #1980, opened 2021-04-10
Fix Jira messages callbacks

wujuu:


jswk (commented):
This is actually to be done alongside executing OMS triggers, i.e. when any event is created, a delayed_job should be registered to send the calls. Apart from running these triggers, it should also send the legacy info about a new message through the legacy Jira integration.

In the previous implementation the direction MP->Jira is not implemented, so I wouldn't implement it now, only leave it for the new adapter.

PR #1981, opened 2021-04-10
Fix AddSombo task

wujuu:


lib/ordering_api/add_sombo.rb
```diff
@@ -18,9 +18,11 @@ def call
...
       end
 
-      Offer.all.each do |o|
-        if o.current_oms == sombo && o.service.order_target.present?
-          o.update(oms_params: { order_target: o.service.order_target })
+      Service.all.each do |s|
+        s.offers.each do |o|
+          if o.current_oms == sombo && s.order_target.present?
+            o.update(oms_params: { order_target: s.order_target })
+          end
```
jswk, line 25: What does it fix? It seems to me that it has the same logic as before, just the loop is structured differently.

If there was a particular problem with it then give the reason for change in the commit message preferably.

spec/lib/ordering_api/add_sombo_spec.rb
```diff
@@ -26,7 +26,7 @@
     expect(Oms.first.administrators.first.first_name).to eq("SOMBO admin")
   end
 
-  it "if updates offers' oms_params properly" do
+  it "it updates offers' oms_params properly" do
```
jswk, line 29: The `it` syntax actually is used without the repeated "it", therefore it should read `it "updates offers' oms_params properly"`

lib/ordering_api/add_sombo.rb
wujuu: Well it fixed a problem in which offers which didn't have a service (deleted offers, some old offers in the beta database). By restructuring a loop like this we are ensuring that `s.order_target` doesn't throw an exception.

jswk (approved):

PR #1996, opened 2021-04-14
[#1885] Ordering API happy paths and cypress customization

danielkryska:
- Added cypress new commands
  - generate UUID4
  - get random: string, email, URL
  - check captcha
  - force visit URL
- Added cypress types support for custom commands
- Added tests for offers
  - Sanity check of add new project path
  - Add project
  - Edit project
  - Add resources to project: open access, order required
  - Check project sync with JIRA
  - Check if a message from the app has arrived to JIRA epic
- Create auth mock endpoint (replacement of AAI auth)
- Create README.md
- Move e2e catalog to test/system due to [Ruby guides](https://guides.rubyonrails.org/testing.html#system-testing)

---

`Auth mock` can be run by setting env variable `AUTH_MOCK=true`.


jswk (commented):
Please fix the typo in the commit message too (not just PR title):
image

Other than that, the `Users::AuthMockController` seems fine. The condition for its activation is `Mp::Application.config.auth_mock && Rails.env.development?`, so it shouldn't be possible to activate it in production by mistake or otherwise.

jswk (approved):

test/system/cypress/integration/offers.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+
+                        const newProject = ProjectFactory.create();
+                        cy.fillFormProject(newProject);
+                        cy.hasProjectDetails(newProject);
+
+                        cy.clearCookie('project');
+                        cy.setCookie(
+                            'project',
+                            JSON.stringify(newProject),
+                            {domain: Cypress.env('MP_JIRA_URL').replace(/(^\w+:|^)\/\//, '')}
```
michal-szostak, line 99: This replace also occurs quite a few times in the code, It would be a good idea to extract it into a function

test/system/cypress/integration/offers.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+                        cy.get("button.btn-primary")
+                            .contains("Add to a project", {matchCase: false})
+                            .first()
+                            .click();
+                        cy.getCookie('resources')
+                            .then(cookie => {
+                                let resources: IResource[] = [];
+                                if (!!cookie) {
+                                    resources = JSON.parse(cookie.value);
+                                }
```
michal-szostak, line 174: What about tertiary operator?

test/system/cypress/integration/offers.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+                            .should('include.text', 'N/A');
+                        cy.jiraGetTicketDetailFrom('cp-customerCountry')
+                            .should('include.text', Utilities.capitalizeAll(project.countryOfOrigin));
+                        project.scientificDomains
+                            .forEach(scientificDomain => {
+                                cy.jiraGetTicketDetailFrom('cp-scientificDiscipline')
+                                    .should('include.text', Utilities.capitalizeAll(scientificDomain));
+                            });
+                        cy.jiraGetTicketDetailFrom('eosc-hub test')
+                            .should('include.text', 'No');
```
michal-szostak, line 307: This is ok for now, but if we want to test more cases in the future this whole block could be encapsulated into cypress command or a function

test/system/cypress/support/jira.ts
```diff
@@ -0,0 +1,130 @@
...
+Cypress.Commands.add('jiraGetTicketsBy', (type: findTicketBy, value) => {
+    switch (type) {
+        case null:
+            return cy.get(".issue-list > li");
+        case 'name':
+            return cy.get(".issue-list").find('li[title="' + value + '"]');
+        case 'key':
+            return cy.get(".issue-list").contains(".issue-link-key", value);
+        default:
+            throw "Not supported filtration type, please update support/jira.ts file";
```
michal-szostak, line 49: This is also non issue, but throwing string literals is not advised in JS

michal-szostak (approved):
I have some minor comments. Also I have a question why cookies are a primary vehicle of information in some tests?
Is it required workaround?

test/system/cypress/factories/project.factory.ts
```diff
@@ -0,0 +1,30 @@
...
+    customerTypology: 'Single user' | 'Representing a research community';
+    email: string;
+    organization: string;
+    department: string;
+    countryOfOrigin: string;
+    webpage: string;
+}
+
+export const ProjectFactory = {
+    create: (args: {[field: string]: string} = {}): IProject => ({
```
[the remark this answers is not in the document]
danielkryska: I'll consider it in the next iteration of e2e.

test/system/cypress/integration/offers.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+    // In that case "it" is used to run only once
+    it('Setup cookies data', () => {
+        const user = UserFactory.create();
+        cy.clearCookie('user');
+        cy.setCookie(
+            'user',
+            JSON.stringify(user),
+            {domain: Cypress.env('MP_JIRA_URL').replace(/(^\w+:|^)\/\//, '')}
+        );
+        cy.setCookie(
```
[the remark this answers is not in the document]
danielkryska: Clearing cookies has its purpose. 

Cypress:
- On `setCookie` always add a new Cookie (without overriding), such feature will provide a lot of duplicated data in store with random placement.
- `setCookie` domain always point to window.location.hostname. This behaviour leads us to visit the site before setting a cookie. 
- `visit` different domains in this same test is bugged and disallowed in many ways in cypress.

Knowledge source: https://docs.cypress.io/api/commands/setcookie#Arguments

Conclusion:
Getting knowledge about features of `setCookie` and `visit` functionalities:
- setting cookie twice for different domains has a purpose to fetch the same data in JIRA and MP
- clearing old cookies assure us that there aren't duplicates with same key

test/system/cypress/integration/offers.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+            'project',
+            JSON.stringify(project),
+            {domain: Cypress.config().baseUrl.replace(/(^\w+:|^)\/\//, '')}
+        );
+    });
+    it('Should go to new project', () => {
+        cy.visit("/");
+        cy.getCookie('user')
+            .then(cookie => JSON.parse(cookie.value) as IUser)
```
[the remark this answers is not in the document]
danielkryska: Such a function will have the same length of code (1 line). I think it's simple enough.

test/system/cypress/integration/offers.spec.ts
danielkryska: It's already a function `setCookie`. :P But seriously this code chunk will be used only in specific integrations like one with JIRA. I think it's simple enough to keep it like this.

test/system/cypress/integration/offers.spec.ts
danielkryska: Good point. I'm planning to refactor tests in the next steps, where I'll fix it up.

test/system/cypress/integration/offers.spec.ts
danielkryska: For now, I'm not considering any `jiraGetTicketDetailFrom` usage except `Should create epic in JIRA on new project` case.

test/system/cypress/integration/offers.spec.ts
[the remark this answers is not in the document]
michal-szostak: True but:
1) It would REQUIRE typing (when using `JSON.parse` it's optional)
2) You could wrap it into cypress command to make something like `cy.getJsonFromCookie<Type>("cookieName").then(...)`

test/system/cypress/integration/offers.spec.ts
michal-szostak: In this case I'd at least extract Regexp literal

Issue #2012, opened 2021-04-21
Add OMS to the Resource Offer


roksanaer:
1. Offer model 
     1. Introduce OMS and offer `custom_params` - consult the status with @jswk 
          1. introduce assignment `offer.primary_oms` to the default OMS when `internal == true` (rethink this together with @jswk)
          1. migrate `service.order_target` to offer `custom_params`
    1. get rid of `service.order_target`

1. Offer frontend
     1. In all offer configuration panels of `order_required` type - default offer and offers in general (like technical parameters but higher 😛) (fixed by #2059)
         1. Show `internal` checkbox
         1. Show OMS selection field when `internal == true` 
            2. select field with the only value "SOMBO"   
         1. Show `custom_params` when `internal == true` 
1. Offer API **(fixed by https://github.com/cyfronet-fid/marketplace/pull/2033)**
   1. GET - When `internal == false` ignore (do not return) `primary_oms_id` and `oms_params`
   2. PATCH/POST - When `internal == false` ignore `primary_oms_id` and `oms_params`
   3. Maybe "strong params" are the right place to handle this

jswk:
Notice that the scope of this ticket partly overlaps with #1880. Both may need merging.

bwilk:
https://github.com/cyfronet-fid/marketplace/pull/2018 partially introduces needed enhancements

wujuu:
**IMPORTANT**

> 1. Offer API 
>     1. GET - When `internal == false` ignore (do not return) `primary_oms_id` and `oms_params`
>     2. PATCH/POST - When `internal == false` ignore `primary_oms_id` and `oms_params`
>     3. Maybe "strong params" are the right place to handle this

Issue #1988, opened 2021-04-13
Resource offer metadata changes


roksanaer:
![image.png](https://images.zenhubusercontent.com/5b0fef1989d5dc41604fb4c3/327e8190-785f-4ea8-ab08-3fe3b64583a7)

### General assumptions 
- we don't change PC-MP integration logic  
- `internal` overrides any logic based on `order_url` existence  

### Current
1. Offer model 
     1. `website` 
     1. `order_url` 
     1. `internal` 
2. Offer frontend 
     1. `website` 
     1. `order_url` 
     1. `internal` 

### Desired
1. Offer model 
    1. Using db migration assign `order_url = order_url || website `
    1. Get rid of `website`
    2. Reset `internal` to false in the model when `order_type != order_required` to avoid problems when the field is hidden **(fixed by https://github.com/cyfronet-fid/marketplace/pull/2033)**
    2. In the db migration set `internal = false` when `order_type != order_required` **(fixed by https://github.com/cyfronet-fid/marketplace/pull/2033)**
    
1. Offer frontend
    1. Get rid of `website` (order_url should be used throughout the system to avoid dychotomy - see the sidenote in the comment below)
    2. `order_url` is hidden when `internal = true` (in "default offer" layout it is in the hidden section anyway) (fixed by #2059)
    2. Label for `order_url`: "Order / access URL" - to avoid confusion between open access / order required services (fixed by #2059)
    3. Hide `internal` when `order_type != order_required` (fixed by #2059)
    3. Label for `internal` checkbox: "Use EOSC Portal as the order management platform" (fixed by #2059)

1. Offer API
    1. GET - When `order_type != order_required` ignore (do not return) `internal` **(fixed by https://github.com/cyfronet-fid/marketplace/pull/2033)**
    2. PATCH/POST - When `order_type != order_required` ignore `internal` **(fixed by https://github.com/cyfronet-fid/marketplace/pull/2033)**
    3. Maybe "strong params" are the right place to handle this **(fixed by https://github.com/cyfronet-fid/marketplace/pull/2033)**
    4. Remove Offer.webpage field.

### Sidenote

In the Service Access Wizard (frontend) we should use `order_url` field to create access button regardless the service type (obviously the button does not exist for the internally orderable services). When `order_url` is nil we should have a fallback (frontend) to the `offer.service.webpage_url`

bwilk:
Current:
1. Offer model 
     1. website 
     1. order_url 
     1. internal 
 
[...]

Issue #1880, opened 2021-03-19
Add OMS to service offer view


jswk:
When provider edits/creates an offer, she should have to choose a primary OMS from the global OMS and the service provider's OMSes.

Add an extra option in the UI to assign OMS to an offer.
When creating and editing, allow a provider admin to choose from available OMSes for that resource (service): these will include global OMSes and OMSes registered for the resource's provider.

If an OMS specifies custom_offer_details, then require their input (for now, text fields without validation).

Consider #1964 and https://docs.cyfronet.pl/pages/viewpage.action?spaceKey=FID&title=Studium+przypadku+Oms.custom_params+i+Offer.oms_params.

PR #2013, opened 2021-04-21
Improve Offerable logic and use it in OnStatusTypeUpdated

jswk:
In conversation with @Marcelinna we decided that improving on what is in this PR is something out of scope of #2002 (#1941) but worth doing anyway, as currently hardly any emails are sent for project_item changes (i.e. most mails are only sent when `service.order_url.blank?`, and it's mostly populated (at least on PR DB) and it's populated for AoD services too, so no emails for voucherable are sent **AT ALL**).

----------------------------

The legacy Jira adapter email sending logic was based on
service.external, which was in some place in project's history migrated
from service to offerable (i.e. offer and project_item).
This caused an inconsitency between mailer and ordering wizard behaviour
(i.e. the mailer hardly sent any mails).

To correct this, switch the ProjectItem::OnStatusTypeUpdated to the
offerable methods.
Also, refactor the current code around offerable, simplifying the
expressions used to enforce the existing logic and prune some unused
methods.

--------------------------

What are your thoughts on this @bwilk? I reckon that @goreck888, you also were correcting this logic recently, so it may be worth a look by you as well.

---------------------------

# How to test

API-key: `W8jzMHDeahxT__rmVGdZ`
oms_id: `1`

Below are services/offers to go through and verify emails are sent.
For each service disregard the values in parens (it's for @jswk not to go crazy).
For each service, as an ordinary user go and order an offer from it (for voucherable don't provide a voucher).
Then set the status_type of such project_item through API to values in list and see if the emails appear, for example for DEEP, set the status_type to "ready", and then you should receive the mail with content [from](https://github.com/cyfronet-fid/marketplace/blob/improve-offerable-and-project-item-mailer-logic/app/views/project_item_mailer/ready_to_use.html.haml).

To set the status_type of project_item: `PATCH /projects/{project_id}/project_item/{project_item_id}` with payload `{"status":{"value": "whatever", "type": "<status_type>"}}`.

DEEP training facility: (order_type=order_required, order_url="")
 - waiting_for_response: :waiting_for_response
 - approved: :approved
 - ready: :ready_to_use
 - rejected: :rejected
 - closed: :closed

100: (order_type=order_required, order_url!="")
 - waiting_for_response: :waiting_for_response
 - approved: :approved
 - ready: :ready_to_use
 - rejected: :rejected
 - closed: :closed

EC3 non-voucherable offer: (order_type=order_required, order_url!="", aod, non-voucherable)
 - waiting_for_response: :waiting_for_response
 - approved: :approved
 - ready: :aod_accepted
 - rejected: :rejected
 - closed: :closed

EC3 voucherable offer: (order_type=order_required, order_url!="", aod, voucherable)
 - waiting_for_response: :waiting_for_response
 - approved: :approved
 - ready: :aod_voucher_accepted
 - rejected: :aod_voucher_rejected
 - closed: :closed

## How to prepare PR instance

```ruby
OrderingApi::AddSombo.new.call
Oms.first.id
User.find_by(uid: "iamasomboadmin").authentication_token
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Resource Organisation field is required for creating a resource (@kmarszalek)
+- Mail delivery for orderable offers (@jswk)
 
 ## [3.11.0] 2021-04-21
 
--- a/app/components/services/inline_order_url_component.rb
+++ b/app/components/services/inline_order_url_component.rb
@@ -18,7 +18,7 @@ class Services::InlineOrderUrlComponent < ApplicationComponent
   end
 
   def link_name
-    if @offerable.external
+    if @offerable.external?
       _("Go to the order website")
     else
       _("Go to the resource")
@@ -26,6 +26,6 @@ class Services::InlineOrderUrlComponent < ApplicationComponent
   end
 
   def render?
-    !url.blank? && (!@offerable.order_required? || @offerable.external)
+    !url.blank? && !@offerable.orderable?
   end
 end
--- a/app/helpers/project_items_helper.rb
+++ b/app/helpers/project_items_helper.rb
@@ -18,7 +18,7 @@ module ProjectItemsHelper
   end
 
   def webpage(project_item)
-    if project_item.external
+    if project_item.external?
       project_item.order_url
     else
       project_item.webpage
--- a/app/helpers/service_helper.rb
+++ b/app/helpers/service_helper.rb
@@ -95,7 +95,7 @@ module ServiceHelper
   end
 
   def map_view_to_order_type(service_or_offer)
-    if service_or_offer.external
+    if service_or_offer.external?
       "external"
     else
       service_or_offer.order_type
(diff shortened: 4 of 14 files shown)


bwilk:
I think the direction is good and corrects some clearly erroneous behavior. I think that I might be hepful to discuss the "belief" behind this PR but otherwise based on the code analysis I find the state after the changes improved. What might be helpful as well is a definition "what this PR does not solve" - especially the things that are not solved by https://github.com/cyfronet-fid/marketplace/issues/1880, https://github.com/cyfronet-fid/marketplace/issues/2012, https://github.com/cyfronet-fid/marketplace/issues/1988

jswk:
> What might be helpful as well is a definition "what this PR does not solve" - especially the things that are not solved by #1880, #2012, #1988

That's right, this PR doesn't modify the existing logic of `orderable?` as it's posited in some of these tickets but only homogenizes the behaviour between mailer and UI.

The logic improvement is in the scope of its formulation mostly (so that it isn't as convoluted as it was).

jswk:
The test for OnStatusTypeUpdated has been added

Marcelinna (approved):

wujuu (approved):

PR #2021, opened 2021-04-22
[#2020] Fix Field Resource Organisation is required

### LINKED ISSUE #2020: Problem with creating new resource in BackOffice
Tested on v3.11.0-beta.0
**Describe the bug**
after filling all fields correctly, it is not possible to create new Resource in Backoffice

**To Reproduce**
1. Go to '/backoffice'
2. Click on "Owned resource"
3. Click button "Create new resource"
4. Fill in required field 
5. Click button "Create resource"
6. See error : 
So sorry!
An unexpected error occurred. A notification has been sent to our development team for investigation. We are sorry for the inconvenience. 

**Expected behavior**
New resource is created correctly

**Screenshots**

**Additional context**

kmarszalek:
Field Resource Organisation is required, so this PR fixes this problem in BO create Resource form.
Zrzut ekranu 2021-04-22 o 14 38 56

Closes #2020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Resource Organisation field is required for creating a resource (@kmarszalek)
 
 ## [3.11.0] 2021-04-21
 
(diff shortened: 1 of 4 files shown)


jswk:
Should it be actually required? I'm not familiar with its semantics.

If so, then the column `project.resource_organisation_id` should be marked as `null: false` too.
Aaand existing services without it should be given a sensible default (in a migration preferably).

Marcelinna (approved):

jswk:
Ok, so the field is required on the side of PC and we don't have any nulls in DB, so it will suffice to mark the column as not-null.

Marcelinna (approved):

roksanaer (approved):

jswk (approved):

Issue #1998, opened 2021-04-15
Don't recommend unpublished services


JanKapala:
In the recommendation panel there should be only published services. Services having any other status (like draft or deleted) shouldn't appear there.

This issue demand adjustments on both sides: marketplace and recommender.

# On the recommender side

- Only published services ids should be sent as a response by /recommendations endpoint

# On the marketplace side:

- Database dump should contain services statuses.

- Database dump is sent periodically so there can be a time when services statuses in MP database and in the recommender database are inconsistent (for example because status of some service has been modified after sending a dump and before sending a new one) - to prevent MP from displaying unpublished services from recommender
, the appropriate exception handling and unpublished services filtering should be implemented

PR #2023, opened 2021-04-22
Don't recommend unpublished services

JanKapala:
This PR solves issue #1998 on the marketplace side. Should be considered together with cyfronet-fid/recommender-system#51 (that solves this issue on the recommender side).

Now, services sent by a recommender are checked for status before displaying in recommendation panel. If some of them has status other than "published" the recommendation panel isn't displayed.

### DIFF
--- a/app/controllers/concerns/service/recommendable.rb
+++ b/app/controllers/concerns/service/recommendable.rb
@@ -42,14 +42,15 @@ module Service::Recommendable
     def get_recommended_services_by(body, size)
       url = Mp::Application.config.recommender_host + "/recommendations"
       response = Unirest.post(url, { "Content-Type" => "application/json" }, body.to_json)
-      response_body = response.body.transform_keys(&:to_sym)
-      if response.blank? || response_body[:recommendations].blank?
-        Raven.capture_message("Recommendation service, recommendation endpoint response error")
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/service/recommendable.rb
```diff
@@ -39,14 +39,15 @@ def fetch_recommended
     def get_recommended_services_by(body, size)
       url = Mp::Application.config.recommender_host + "/recommendations"
       response = Unirest.post(url, { "Content-Type" => "application/json" }, body.to_json)
-      response_body = response.body.transform_keys(&:to_sym)
-      if response.blank? || response_body[:recommendations].blank?
-        Raven.capture_message("Recommendation service, recommendation endpoint response error")
-        return []
```
wujuu, line 45: Are you sure the "rescue" later will catch all of the errors? response.blank?, response_body[:recommendations].blank?,
NoMethodError "body" for NilClass, request timeout etc?

JanKapala: Yep, any error will be caught by the rescue.

wujuu (approved):

michal-szostak (approved):

Issue #1836, opened 2021-03-02
problem with provider's logos in the backoffice admin panel


roksanaer:
![image](https://user-images.githubusercontent.com/38095945/109626713-280d8f00-7b41-11eb-8f82-ecd7b24e788f.png)

PR #2053, opened 2021-05-07
[#1836] Fix img background on resize

danielkryska:
- Fix all models containing logo field
  - Providers
  - Services
  - Categories
  - Scientific Domains
- Add task migrating images without extension in a name to have one
- Add the image extension to the file name on load
- Set slug as image name on missing pid field

---

Fixing existing images requires migration. Although it's needed to run it with param `ROOT_URL` as for instance domain URL

Example:
`rake add_extension_to_images["http://localhost:5000"]`

---

Before
image

After
image

- Extention on Auth Mock endpoint to add roles to user

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Change img background at /backoffice/providers/<id> from black to transparent on resize (@danielkryska)
 - Add error note on a scientific domain with services deletion (@danielkryska)
 - Change error message at order summary to general - better UX (@danielkryska)
 - Mail delivery for orderable offers (@jswk)
--- a/app/controllers/users/auth_mock_controller.rb
+++ b/app/controllers/users/auth_mock_controller.rb
@@ -16,6 +16,7 @@ class Users::AuthMockController < ApplicationController
       "#{params[:password]}#{nil}",
       cost: 11
     ).to_s
+
     @user = User.find_by(
       email: params[:email],
       encrypted_password: encrypted_password,
--- a/app/views/backoffice/providers/show.html.haml
+++ b/app/views/backoffice/providers/show.html.haml
@@ -6,7 +6,9 @@
   .logo-wrapper
     %span.helper
     - if @provider.logo.attached? && @provider.logo.variable?
-      = provider_logo(@provider)
+      = image_tag @provider.logo.variant(resize: "100x67"), class: "align-self-center mr-4 float-left img-responsive"
+    - else
+      = image_pack_tag "eosc-img.png", size: "100x67", class: "align-self-center mr-4 float-left img-responsive"
   - unless @provider.sources.empty?
     %h3
       = _("External Sources:")
--- a/app/helpers/provider_helper.rb
+++ b/app/helpers/provider_helper.rb
@@ -27,14 +27,6 @@ module ProviderHelper
     fields.map { |f| record.send(f) }.any? { |v| v.present? }
   end
 
-  def provider_logo(provider, classes = "align-self-center mr-4 float-left img-responsive", resize = "100x67")
-    if provider.logo.attached? && provider.logo.variable?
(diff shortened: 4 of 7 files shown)


kmarszalek (approved):
Some changelog entry?

app/views/backoffice/providers/show.html.haml
```diff
@@ -6,7 +6,7 @@
   .logo-wrapper
     %span.helper
     - if @provider.logo.attached? && @provider.logo.variable?
-      = provider_logo(@provider)
+      = image_tag @provider.logo, class: "align-self-center mr-4 float-left img-responsive"
```
jswk, line 9: If the provider_logo isn't necessary then you should also remove it, there are no other usages of it.
Also, for some reason it was resized before, why is it not necessary anymore? Describe the motivation in commit message.

app/controllers/users/auth_mock_controller.rb
```diff
@@ -30,6 +31,7 @@ def login
         uid: SecureRandom.uuid,
         encrypted_password: encrypted_password
       )
+      @user.roles = params[:roles].blank? ? [] : params[:roles].map(&:to_sym)
```
jswk, line 34: Create a separate PR for this, or at least put it in a separate commit, as it totally unreleated to the main problem.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- Fix img at /backoffice/providers/<id> - now colors of image aren't reversed (@danielkryska)
```
jswk, line 19: This isn't descriptive for the end-user and is repetitive ("Fixed", "Fix"). It is also incorrect, there was no color reversal there.
I'd rather describe it like this: "Improve transparency handling in provider logo".

jswk (changes requested):
The PR does somewhat address the issue, but uses a workaround, which probably introduces regression (no resizing and default logo).

The actual fix should correct the way image variants are produced, and handle transparency correctly, not ditch the mechanism.

db/migrate/20210510114948_migrate_providers_images.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  private
+  def should_rename(attachment)
+    if attachment.blank? || attachment.filename.blank?
+      return false
+    end
+
+    has_non_ext = attachment.filename.to_s.split(".")[1].blank?
```
jswk, line 43: If I understand correctly, the intention is to find cases where `filename="foo."` or `filename="foo.bar."`, but not  `filename="foo"`. Then the line should read something like this to be correct:
```ruby
dot_splitted = attachment.filename.to_s.split(".")
has_blank_extension = dot_splitted.size > 1 && dot_splitted[-1].blank?
```

jswk (changes requested):
First of all, this fix isn't as much as a db migration (db migrations mostly deal with schema changes) but data migration unrelated to schema and as such has strong coupling to implementation, which is undesirable for migrations.
Therefore, it has to be moved to a Rake task. As it relates to import, I'd suggest putting it in `import.rake` under key `reimport_logos_with_no_extension`.

Since it should be run after new version's deploy then the necessary task should be described in `docs/upgrade` too.

jswk:
Another insight, this time from @wojt and I didn't notice.
Don't introduce another envvar just to set the default_url_options. When changing the code placement to task, pass it as a task argument from cmd. The task call then looks as follows: `rails some:task[arg_value,another_arg_value]` (i.e. the args are comma seprated). To define such task (example, beware) with arguments email and role:
```ruby
  task :add_role_to_user, [:email, :role] => :environment do |_, args|
    TestBootstrap::AddRoleToUser.new(args.email, args.role).call
  end
```

Also, when exiting the service call method, in the finally block, restore the initial value of the default_url_options, just to be on the safe side if something throws exception.

jswk (commented):
Mark it ready for review when you take into account things from the call

jswk (approved):

PR #2057, opened 2021-05-10
Auth mock extension

danielkryska:
- Add privileges to Auth Mock endpoint
  - admin
  - data admin
  - service portfolio manager
  - executive


jswk (commented):
I don't think the functionality of adding data admin to user should be in a controller.
If so, maybe dev:prime or something like this should seed it. Change my view.

jswk (approved):

danielkryska:
After a quick talk with @jswk, we get to some conclusions. Those are
- Auth Mock controller is only for e2e or development used to create specific user configuration ad hoc
- Keeping in here simplified business logic of user permissions will help in understanding architecture of app

kmarszalek (approved):

Issue #1997, opened 2021-04-14
Editorial rights for providers are consistent with established editing policy


roksanaer:
**Description**  

Insufficient handling of the `upstream` field for the `Provider` entity (see appropriate handling in the Service / Resource entity). All the form fields should be disabled in case of `upstream` set to external source.

**Context**
 
Backoffice is shown to the users with the `Service Portfolio Manager` role. It should stay as is. We can consider getting rid of this user role for the EOSC Hub members having it as it is no longer meant to be used.

PR #2061, opened 2021-05-11
[#1997] Editorial rights for providers

kmarszalek:
Provider in BO should not be editable when upstream is set to MP.

Closing #1997


goreck888 (approved):

Marcelinna:
image
it isn't possible make categories disable too?

Marcelinna (approved):

PR #2070, opened 2021-05-13
[#1988] Purge offer webpage

jswk:
Set the order_url columns to `null: false, default: ""` to avoid
situations where there are records with order_url="" and order_url=null.

Remove column webpage from tables offers and project_items, applying the
same migration method to both, i.e. `order_url = order_url || webpage`.

Remove webpage from offer form, policies, Offer API and
offer_controller.js.

In webpage helper, return order_url and fall back to
service.webpage_url in case order_url.blank?.

Don't propagate service.webpage_url to offer.webpage, as the former
doesn't exist anymore. There is no need to populate offer.order_url with
it, as the order button url falls back to it.

Remove webpage from data.yml, as it was mirroring service.webpage_url
anyway and it doesn't make sense to move it to order_url field.

Correct `</br`> to `<br/>` in texts.

Closes: #1988.

## How to test

Make sure the logic around `orderable?` (i.e. offers and project_items for which `order_type=order_required && internal`) is correct. For example when link to service is presented to user (e.g. in project_items list), that "Order history" is shown for appropriate project_items, and that "Go to the order website" and "Go to the resource" work ok.

### For an offer with `internal=false` (regardless of order_type)

With `order_url` set, the `order_url` value should override `offer.service.webpage_url` when creating the external link button in the ordering wizard.
Otherwise, when order_url will be blank then the button should fall back to `offer.service.webpage_url`.

### For an offer with `internal=true`

The external link button shouldn't be shown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,13 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Resource provider can navigate to the ordering parameters management from the Resource Presentation Page (@kmarszalek, @jarekzet)
 
 ### Changed
+- For an Offer, an empty order URL doesn't imply internal ordering (@jswk)
 - More consistent form behaviour for ordering configuration (@jswk)
 
 ### Deprecated
 
 ### Removed
+- Offer webpage field (@jswk)
 
 ### Fixed
 - Change img background at /backoffice/providers/<id> from black to transparent on resize (@danielkryska)
--- a/app/helpers/project_items_helper.rb
+++ b/app/helpers/project_items_helper.rb
@@ -18,10 +18,10 @@ module ProjectItemsHelper
   end
 
   def webpage(project_item)
-    if project_item.external?
-      project_item.order_url
+    if project_item.order_url.blank?
+      project_item.service.webpage_url
     else
-      project_item.webpage
+      project_item.order_url
     end
   end
 
--- a/app/javascript/app/controllers/offer_controller.js
+++ b/app/javascript/app/controllers/offer_controller.js
@@ -2,11 +2,10 @@ import {Controller} from 'stimulus'
 import initChoises from "../choises";
 
 export default class extends Controller {
-  static targets = ["parameters", "webpage", "external",
+  static targets = ["parameters", "external",
                     "attributes", "button", "attributeType"];
 
   initialize() {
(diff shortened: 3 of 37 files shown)


kmarszalek (approved):

Marcelinna:
Validation to Order / access URL should be added 
image
image

Marcelinna:
_Url should start with http or https [e.g. http://webpage.org].
On service order information step we will provide this URL. If it isn't specified we will fall back to the service's webpage URL._ 
For me the text is too small, I have a problem reading it, maybe it should be bigger and more visible.
@abacz @roksanaer 


image

roksanaer:
> _Url should start with http or https [e.g. http://webpage.org]. On service order information step we will provide this URL. If it isn't specified we will fall back to the service's webpage URL._
> For me the text is too small, I have a problem reading it, maybe it should be bigger and more visible.
> @abacz @roksanaer
> 
> image
Text should implement:

Order/ access Url should start with the  http or https prefix [e.g. http://webpage.org].
Given URL will be used in the offer order/access mechanism - on the resource access information step this URL will be used under 'Go to the order website' / 'Go to the resource' button (depending on the offer order typer). If it isn't specified it will fall back to the resource's webpage URL.

In my opinion the text could be larger but maybe @abacz should have a final say

jswk:
Things corrected:
- "Order / access URL" validation https://github.com/cyfronet-fid/marketplace/pull/2070#issuecomment-842304454
- The new "Order / access URL" hint https://github.com/cyfronet-fid/marketplace/pull/2070#issuecomment-842323987
- Add information about change in order URL handling to changelog
- Set the internal flag for offers (and project_items) which were `effective_internal?` before, but didn't have the flag set
- Add a label about oms_params, so that the intent of the showed fields is easier to understand for providers
- Add missing breadcrumbs for offers

bwilk:
@roksanaer I've suggested following text: "Given URL will be used to guide the users when trying to access the resource. 'Go to the order website' / 'Go to the resource' button (depending on the offer order type) will navigate to the resource's webpage URL if the Order/access URL is not specified for the offer.". May be ignored at this point. Just for the record.

bwilk (approved):

kmarszalek (approved):

Marcelinna (approved):

PR #2084, opened 2021-05-17
[#2000] Vocabularies import

### LINKED ISSUE #2000: Prepare for sync vocabularies with PC
The list of vocabularies starts to grow. This is related to the new functionality of the Provider Component, that allows to introduce `Other` values for vocabulary objects. 

This is how the list of the pending requirements list looks like. 

image

For now this is only used to get information about the providers needs to include new values of vocabularies in the list of available options. Although, the list will start evolving slowly. We need to address populating the value list using the PC API.  

1. Use PC API to get vocabularies https://api.eosc-portal.eu/vocabulary/byType
2. Keep the seeds for the development purposes (or replace the file with the response of the PC to make it easy to update) https://github.com/cyfronet-fid/marketplace/blob/master/db/vocabulary.yml
3. Address how to manage the list of vocabularies in the UI (especially for the Whitelabel) - should we manage them individually or all vocabularies together? 

**Acceptance criteria**

1. New vocabularies can be added on the PC side
2. Vocabularies meant to be controlled in backoffice can supplement the list of the vocabularies from PC (for Whitelabel)
3. The SPM user role can manage source of the vocabularies

goreck888:
Allow REST to import vocabularies from PC
Closes #2000


lib/import/vocabularies.rb
```diff
@@ -0,0 +1,111 @@
...
+                 filepath: nil,
+                 unirest: Unirest,
+                 logger: ->(msg) { puts msg },
+                 token: nil)
+    @eic_base_url = eic_base_url
+    @dry_run = dry_run
+    @unirest = unirest
+    @token = token
+
+    @logger = logger
```
jswk, line 42: Maybe just assign `Rails.logger` here? You generally want to have logs, and using the native logger instead of puts seems better.

jswk (commented):
If target_user is now parentable, doesn't it need different treatment in forms too? Also, why is this change (parentable etc.) introduced?

Also, from the acceptance criteria of #2000:
> - The SPM user role can manage source of the vocabularies

How is this realised in this PR? Maybe there's an existing code for this but I missed it.

lib/import/vocabularies.rb
```diff
@@ -0,0 +1,111 @@
...
+    @filepath = filepath
+  end
+
+  def call
+    log "Importing vocabularies from eInfraCentral..."
+
+    begin
+      r = Importers::Request.new(@eic_base_url, "vocabulary/byType", unirest: @unirest, token: @token).call
+    rescue Errno::ECONNREFUSED
+      abort("import exited with errors - could not connect to #{@eic_base_url}")
```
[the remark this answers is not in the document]
goreck888: About logger I just left it as it is in other importers (eic, providers). I think we could change it as a refactoring in all importers.

spec/lib/import/vocabularies_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  describe "#standard responses" do
+    before(:each) do
+      response = double(code: 200, body: create(:eic_vocabularies_response))
+      expect_responses(unirest, test_url, response)
+    end
+
+    it "should create and update vocabularies" do
+      esfri_type = create(:esfri_type, name: "TEST", eid: "provider_esfri_type-landmark")
```
jswk, line 49: I would move it to a `let!` block, so that in dry-run test you also have a case with some vocabularies already existing.

jswk (approved):
Apart from the test suggestion, the PR is ok now.

The management by SPM will be done in scope of #2118 and target_user ancestry is handled seamlessly according to @goreck888, so no further modifications are required in forms or such.

spec/lib/import/vocabularies_spec.rb
goreck888: Ok :)

kmarszalek (approved):

PR #2095, opened 2021-05-19
3.12.x merge

danielkryska:


kmarszalek (approved):

kmarszalek (changes requested):
not sure that is the best solution.. @michal-szostak

jswk:
I'd suggest doing it with an actual merge: https://github.com/cyfronet-fid/marketplace/pull/2096.

PR #2098, opened 2021-05-19
[HOTFIX] Links to PC

goreck888:
PROVIDERS_DASHBOARD_URL should be only link to the PC instance


kmarszalek (changes requested):
I think that we should have PROVIDERS_COMPONENT_URL for that?
then there will be no confusion about destination of this env :)

PR #2112, opened 2021-05-24
[HOTFIX] Fix validator messages for MinMax parameter (backport to master)

goreck888:
Min and Max values implemented in parameters
had wrong message: "must be grater/lesser..."
From now it's correct "must be greater/lesser or equal..."

### DIFF
--- a/app/models/parameter/minmax.rb
+++ b/app/models/parameter/minmax.rb
@@ -12,12 +12,12 @@ module Parameter::Minmax
 
     validates :min, numericality: {
       less_than_or_equal_to: ->(p) { p.max },
-      message: "must be less than maximum value"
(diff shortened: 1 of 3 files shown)


Marcelinna:
It is possible for the case when the user does not complete the Values doesnt show message for min:Min must be less than or equal to 0 and max Max must be less than or equal to 0 ?
image

kmarszalek (changes requested):
hotfix - it should be merge to release candidate :) 
msg validation in @Marcelinna's case should be also done in this PR

kmarszalek (changes requested):
first rebase to version, then version rebase to master. this pr you can cancel. if not there will be conflicts because commit in this PR or has diff hash than one in PR to the version.

PR #2125, opened 2021-05-27
[#1931] Remove unwanted search parameters in url

### LINKED ISSUE #1931: remove unwanted parameters in url
When selecting and deselecting filters, the url should not contain unnecessary characters.

Current url addresses:
selecting a filter
https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services?sort=_score&scientific_domains-filter=&scientific_domains-all=&scientific_domains%5B%5D=1&providers-filter=&providers-all=&target_users-all=&related_platforms-filter=&related_platforms-all=&rating=&order_type=&geographical_availabilities=

deselecting filter
https://marketplace.docker-fid.grid.cyf-kr.edu.pl/services?geographical_availabilities=&order_type=&providers-all=&providers-filter=&rating=&related_platforms-all=&related_platforms-filter=&scientific_domains-all=&scientific_domains-filter=&sort=_score&target_users-all=

goreck888:
Remove empty filter parameters in resource search
Closes #1931

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - For an Offer, an empty order URL doesn't imply internal ordering (@jswk)
 - More consistent form behaviour for ordering configuration (@jswk)
 - From now after click "Stop showing in the MP" on the resource page, offers aren't deleted (@goreck888) 
+- Remove empty filter parameters in resources search, which makes urls shorter (@goreck888)
 
 ### Deprecated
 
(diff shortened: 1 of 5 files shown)


jswk (changes requested):
If I go to a path with an empty param (i.e. `https://pr-2125.docker-fid.grid.cyf-kr.edu.pl/services?foo=`, but you have to correct the url after malformed redirection) then you get an error page (nothing in the container logs though).
image

I think it is the code in the `services_controller.rb`, which is liable.
Moreover, I don't think it is necessary. The idea of the ticket was to remove empty params from the query, which you did in the `filter_controller.js`.
If somebody navigates to a path with empty params like I did, it may be good that you are redirected back to a version without them, but I would find it misleading. Also, I'm not sure if the added code actually does that, as activating it triggers an error.

So, I would say to revert the changes in `services_controller.js` whatsover, and rely on the javascript event handler.
If you deem the mechanism important though, the same should be added in backoffice services_controller.rb (that controller wasn't modified here, and it works as intended).

PR #2150, opened 2021-06-08
[#2118] Vocabularies management in the marketplace

### LINKED ISSUE #2118: Create UI management for vocabularies
This is a second part of the #2000 

Add possibility to manage vocabularies in the Backoffice.

The first solution is to have separate links to all vocabulary type.
The second is to have one link to all vocabularies.

goreck888:
Add management panel in the backoffice for vocabularies

Closes #2118

### DIFF
--- a/app/controllers/backoffice/categories_controller.rb
+++ b/app/controllers/backoffice/categories_controller.rb
@@ -41,9 +41,19 @@ class Backoffice::CategoriesController < Backoffice::ApplicationController
   end
 
   def destroy
-    @category.destroy!
-    redirect_to backoffice_categories_path,
-                notice: "Category destroyed"
+    if @category.descendant_ids.present?
+      redirect_back fallback_location:  backoffice_category_path(@category),
+                    alert: "This category has successors connected to it,
+                            therefore is not possible to remove it. If you want to remove it,
+                            edit them so they are not associated with this category anymore"
+    elsif @category.services.present?
+      redirect_back fallback_location:  backoffice_category_path(@category),
+                    alert: "This category has resources connected to it, remove associations to delete it."
+    else
+      @category.destroy!
+      redirect_to backoffice_categories_path,
+                  notice: "Category removed"
+    end
   end
 
   private
--- a/app/controllers/backoffice/scientific_domains_controller.rb
+++ b/app/controllers/backoffice/scientific_domains_controller.rb
@@ -41,16 +41,21 @@ class Backoffice::ScientificDomainsController < Backoffice::ApplicationControlle
   end
 
   def destroy
-    if @scientific_domain.descendant_ids.blank?
-      @scientific_domain.destroy!
-      redirect_to backoffice_scientific_domains_path,
-                  notice: "Scientific domain removed"
+    if @scientific_domain.descendant_ids.present?
+      redirect_back fallback_location:  backoffice_scientific_domain_path(@scientific_domain),
+                    alert: "This scientific domain has successors connected to it,
+                            therefore is not possible to remove it. If you want to remove it,
+                            edit them so they are not associated with this scientific domain anymore"
+    elsif @scientific_domain.services.present?
+      redirect_back fallback_location:  backoffice_scientific_domain_path(@scientific_domain),
+                    alert: "This scientific domain has resources connected to it, remove associations to delete it."
+    elsif @scientific_domain.providers.present?
+      redirect_back fallback_location:  backoffice_scientific_domain_path(@scientific_domain),
+                    alert: "This scientific domain has providers connected to it, remove associations to delete it."
     else
-      redirect_to backoffice_scientific_domain_path(@scientific_domain),
-                  alert: "This scientific domain has resources connected to it,
-                        therefore is not possible to remove it. If you want to remove it,
-                        please go to the service list view
-                        and edit them so they are not associated with this scientific domain anymore"
+      @scientific_domain.destroy!
+      redirect_to backoffice_categories_path,
+                  notice: "Scientific Domain removed"
     end
   end
 
--- a/app/controllers/backoffice/vocabularies_controller.rb
+++ b/app/controllers/backoffice/vocabularies_controller.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+class Backoffice::VocabulariesController < Backoffice::ApplicationController
+  before_action :find_and_authorize, only: [:show, :edit, :update, :destroy]
+  before_action :instantiate_type
+
+  VOCABULARY_TYPES = {
+    "TargetUser": "Target User",
+    "Vocabulary::AccessMode": "Access Mode",
+    "Vocabulary::AccessType": "Access Type",
+    "Vocabulary::FundingBody": "Funding Body",
+    "Vocabulary::FundingProgram": "Funding Program",
+    "Vocabulary::Trl": "TRL",
+    "Vocabulary::LifeCycleStatus": "Life Cycle Status",
+    "Vocabulary::ProviderLifeCycleStatus": "Provider Life Cycle Status",
+    "Vocabulary::AreaOfActivity": "Area of Activity",
+    "Vocabulary::EsfriDomain": "ESFRI Domain",
+    "Vocabulary::EsfriType": "ESFRI Type",
+    "Vocabulary::LegalStatus": "Legal Status",
+    "Vocabulary::Network": "Network",
+    "Vocabulary::SocietalGrandChallenge": "Societal Grand Challenge",
+    "Vocabulary::StructureType": "Structure Type",
+    "Vocabulary::MerilScientificDomain": "MERIL Scientific Domain"
+  }
+
+  def index
+    authorize(Vocabulary)
+    @pagy, @vocabularies = pagy(vocabulary_type.all.order(:name))
+  end
+
+  def show
+  end
+
+  def new
+    @vocabulary = vocabulary_type.new
+    authorize(@vocabulary)
+  end
+
+  def create
+    @vocabulary = vocabulary_type.new(permitted_attributes(vocabulary_type))
+    authorize(@vocabulary)
+
+    if @vocabulary.save
+      redirect_to send("backoffice_#{@vocabulary.model_name.element}_path", @vocabulary),
+                  notice: "New access mode created successfully"
+    else
+      render :new, status: :bad_request
+    end
+  end
+
+  def edit
+  end
+
+  def update
+    if @vocabulary.update(permitted_attributes(@vocabulary))
+      redirect_to send("backoffice_#{@vocabulary.model_name.element}_path", @vocabulary),
+                  notice: "#{@vocabulary.model_name.element.humanize} updated correctly"
+    else
+      render :edit, status: :bad_request
+    end
+  end
+
+  def destroy
+    if @vocabulary.descendant_ids.blank?
+      @vocabulary.destroy!
+      redirect_to send("backoffice_#{@vocabulary.model_name.element.pluralize}_path"),
+                  notice: "#{@vocabulary.model_name.element.humanize} removed"
+    else
+      redirect_to send("backoffice_#{@vocabulary.model_name.element}_path", (@vocabulary)),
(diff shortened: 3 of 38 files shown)


jswk:
If it's blocking current release than it could be merged (if tests pass and so on), but it must be refactored in the upcoming sprint, taking the above into consideration.

goreck888:
> Do we really want to introduce so much repetition with this?
> Each vocabulary type has the same controller and views associated with it, which could be parametrized.
> The routes could include a set of routes `[...]/vocabularies/:vocabulary_type[...]` and you would then have a single controller (with views and policy), which would handle such requests. It could have a static mapping between the `:vocabulary_type` path-param and concrete type. If you represented the type in snake_case then you could generate labels by using appropriate string transformations.
> 
> Is there a reason why it's so duplicated? Maybe each vocabulary will have distinct policy?, or some other editing rules? I may have missed this and it makes sense to keep it this way.

Not all vocabularies are included in `Vocabulary` scope, some are historically implemented as just a non-namespaced class. That's the reason I did it this way.

jswk (approved):
The issues raised above https://github.com/cyfronet-fid/marketplace/pull/2150#pullrequestreview-678592411 should be addressed in the upcoming sprint, but for now it could be left as is, if @Marcelinna approves that it functions ok.

Marcelinna:
Why Eid sometimes is required, sometimes no?
image
image

Marcelinna:
I am on Funding Bodies but see button Add new Target user
image

Marcelinna:
For me the text should wrap earlier, now it isnt looks good
image

Marcelinna:
When I click on one of the Access Mode, I see error
image

goreck888:
> Why Eid sometimes is required, sometimes no?
> image
> image

It's ok, eid is not required only in Target User, because it is a model not included in vocabularies, similar like ScientificDomain or Category.

goreck888:
> Why Eid sometimes is required, sometimes no?
> image
> image

It's ok, eid is not required only in Target User, because it is a model not included in vocabularies, similar like ScientificDomain or Category.

Marcelinna:
- [ ] QA
- [ ] DEV

Header for Life Cycle Statuses is Target User
image

app/controllers/backoffice/vocabularies/provider_life_cycle_statuses_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  def destroy
+    if @vocabulary.descendant_ids.blank?
+      @vocabulary.destroy!
+      redirect_to backoffice_vocabularies_provider_life_cycle_statuses_path,
+                  notice: "Provider life cycle status removed"
+    else
+      redirect_to backoffice_vocabularies_provider_life_cycle_status_path(@vocabulary),
```
Marcelinna, line 49: I can remove element from Provider Life Cycle Statuses (e.g Under Construction ) despite it is used in provider form.

app/controllers/backoffice/vocabularies/provider_life_cycle_statuses_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+                  alert: "This provider life cycle status has resources connected to it,
+                      therefore is not possible to remove it. If you want to remove it,
+                      please go to the provider life cycle status list view
+                      and edit them so they are not associated with this provider life cycle status anymore"
+    end
```
Marcelinna, line 54: I can remove element from Provider Life Cycle Statuses (e.g Under Construction ) despite it is used in provider form.

app/controllers/backoffice/vocabularies/provider_life_cycle_statuses_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+end
```
Marcelinna, line 56: I can remove element from Provider Life Cycle Statuses (e.g Under Construction ) despite it is used in provider form.

Marcelinna: Alert: "This provider life cycle status has resources connected to it..........." it should be changed

goreck888: In this context "resources" doesn't mean exactly a Resource object, but other resources (other vocabularies set as parent, provider, resource etc.). Of course we could change it to "objects" or something else to have a consistency of our nomenclature.

goreck888: I changed it to successors.

app/controllers/backoffice/vocabularies_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    "Vocabulary::ProviderLifeCycleStatus": "Provider Life Cycle Status",
+    "Vocabulary::AreaOfActivity": "Area of Activity",
+    "Vocabulary::EsfriDomain": "ESFRI Domain",
+    "Vocabulary::EsfriType": "ESFRI Type",
+    "Vocabulary::LegalStatus": "Legal Status",
+    "Vocabulary::Network": "Network",
+    "Vocabulary::SocietalGrandChallenge": "Societal Grand Challenge",
+    "Vocabulary::StructureType": "Structure Type",
+    "Vocabulary::MerilScientificDomain": "MERIL Scientific Domain"
+  }
```
jswk, line 24: This is great and all, but to avoid constantizing later on, you could specify the name and klass directly for each path. I.e.
```ruby
  VOCABULARY_TYPES = {
    "target_user": { name: "Target User", klass: TargetUser },
    ...
  }
```
this would remove exact class names from routes, and allow you to only differentiate based on resource name (possibly).

app/controllers/backoffice/vocabularies_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    else
+      alert = @vocabulary.descendant_ids.blank? ? "This vocabulary has resources or providers connected to it,
+        remove associations to delete it." : "This #{VOCABULARY_TYPES[@vocabulary.model_name.name.to_sym]}
+        has successors connected to it, therefore is not possible to remove it. If you want to remove it,
+        please go to the #{VOCABULARY_TYPES[@vocabulary.model_name.name.to_sym]} list view and edit them
+        so they are not associated with this #{VOCABULARY_TYPES[@vocabulary.model_name.name.to_sym]} anymore"
+
+      redirect_to send("backoffice_#{@vocabulary.model_name.element}_path", (@vocabulary)),
+                  alert: alert
+    end
```
jswk, line 77: I would also reorganize this method:
```ruby
    if @vocabulary.descendant_ids.present?
      redirect_back fallback_location: destroy_fallback_location,
                    alert: "This #{@type_name}
        has successors connected to it, therefore is not possible to remove it. If you want to remove it,
        please go to the #{@type_name} list view and edit them
        so they are not associated with this #{@type_name} anymore"
    elsif @vocabulary.services.present?
      redirect_back fallback_location: destroy_fallback_location,
                    alert: "This vocabulary has resources connected to it, remove associations to delete it."
    elsif @vocabulary.providers.present?
      redirect_back fallback_location: destroy_fallback_location,
                    alert: "This vocabulary has providers connected to it, remove associations to delete it."
    else
      @vocabulary.destroy!
      redirect_to send("backoffice_#{@vocabulary.model_name.element.pluralize}_path"),
                  notice: "#{@type_name} removed"
    end
```
The cases for providers and services could be merged too, however, this way it's more specific.

spec/features/backoffice/vocabularies_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+
+      scenario "I can delete existing vocabulary if is not root element" do
+        existing_vocabulary = create(vocabulary)
+
+        visit send("backoffice_#{vocabulary}_path", existing_vocabulary)
+
+        click_on "Delete"
+
+        expect(page).to have_text("#{humanized} removed")
+      end
```
jswk, line 91: These two seem duplicated.

jswk (changes requested):
Sorry for slow processing, these are several possible improvements I've determined.

app/controllers/backoffice/categories_controller.rb
```diff
@@ -41,9 +41,20 @@ def update
...
+                  notice: "Category removed"
+    else
+      alert = @scientific_domain.descendant_ids.blank? ? "This category has resources or providers
+        connected to it, remove associations to delete it." : "This scientific domain
+        has successors connected to it, therefore is not possible to remove it. If you want to remove it,
+        please go to the scientific domain list view and edit them so they are not associated
+        with this scientific domain anymore"
+      redirect_to backoffice_category_path(@scientific_domain),
+                  alert: alert
+    end
```
jswk, line 57: I'd do it the same as it is done in `vocabularies_controller.rb`.
Also, for category you know what associations it has (services, providers), so there is no need to use the syntax with try.

app/controllers/backoffice/scientific_domains_controller.rb
```diff
@@ -41,16 +41,19 @@ def update
...
+        connected to it, remove associations to delete it." : "This Scientific Domain
+        has successors connected to it, therefore is not possible to remove it. If you want to remove it,
+        please go to the Scientific Domain list view and edit them so they are not associated
+        with this scientific domain anymore"
       redirect_to backoffice_scientific_domain_path(@scientific_domain),
-                  alert: "This scientific domain has resources connected to it,
-                        therefore is not possible to remove it. If you want to remove it,
-                        please go to the service list view
-                        and edit them so they are not associated with this scientific domain anymore"
+                  alert: alert
```
jswk, line 56: I'd do it the same as it is done in `vocabularies_controller.rb`.
Also, for scientific_domain you know what associations it has (services, providers), so there is no need to use the syntax with try.

spec/features/backoffice/vocabularies_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    area_of_activity: "Area of Activity",
+    esfri_domain: "ESFRI Domain",
+    esfri_type: "ESFRI Type",
+    legal_status: "Legal Status",
+    network: "Network",
+    societal_grand_challenge: "Societal Grand Challenge",
+    structure_type: "Structure Type",
+    meril_scientific_domain: "MERIL Scientific Domain"
+  }.each do |vocabulary, humanized|
+    context "As a service portolio manager" do
```
jswk, line 26: I'd add another context level here, for example:
```ruby
context "For #{vocabulary}" do
  ...
end
```

spec/features/backoffice/vocabularies_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      scenario "I can delete existing vocabulary if is not root element" do
+        existing_vocabulary = create(vocabulary)
+
+        visit send("backoffice_#{vocabulary}_path", existing_vocabulary)
+
+        click_on "Delete"
+
+        expect(page).to have_text("#{humanized} removed")
+      end
+
```
jswk, line 82: I would keep the test for removing a vocabulary entry with children:
```ruby
      scenario "I cannot remove a vocabulary if it has children" do
        parent = create(vocabulary)
        create(vocabulary, ancestry: parent.id)

        visit send("backoffice_#{vocabulary}_path", parent)

        click_on "Delete"

        expect(page).to have_text("This #{humanized} " +
                                    "has successors connected to it, therefore is not possible to remove it. If you want to remove it, " +
                                    "please go to the #{humanized} list view and edit them so they are not associated " +
                                    "with this #{humanized} anymore")
      end
```

jswk (changes requested):
Thanks for the corrections, it looks much better now.
There are only these several minor cases left and I think the PR is ready to go.

Also, remember about the changelog.

jswk (approved):
Looks good to go for me!

Marcelinna:
Please change notyfiktion:
When user try delete child :
This Legal Status has successors connected to it, therefore is not possible to remove it. If you want to remove it, edit them so they are not associated with this Legal Status anymore

When user try delete vocabularies conntected with resources/providers:

This vocabulary has resources/providers connected to it, remove associations to delete it.

Marcelinna (approved):

Issue #2143, opened 2021-06-08
Log all errors occurred while importing to sentry with specific tag


danielkryska:
## Context
To keep track of errors while importing, we'll need to log them in sentry. Having knowledge of repeating crashes of logo conversion we want to add a special tag to logged data. This will prevent a flood of repeatable data in the console.

michal-szostak:
If at some point in time we decide to run import tasks non interactively (as a cron job for example). This feature will be in high demand. If import tasks are run by hand infrequently this should not be an issue

PR #2078, opened 2021-05-17
[#1977] Improve consistency between MP provider and PC provider

danielkryska:
IMPORTANT!!! 
Threads from now on will be aborted on exception due to lack of tools to influence nested threads pools.
The most common errored place is Image Magic conversion to PNG.

---

API information about fields types: https://providers.eosc-portal.eu/openapi

Rake task to add default logo to providers if doesn't exists: `rake add_providers_default_logo`

- Changes for a Provider model and the Provider form in /backoffice/providers/...
  - Basic
    - Abbreviation should be required
    - Website should be required
    - Legal entities should be required
    - Legal statuses should be a select field
  - Marketing
    - Description should be required
    - Logo should be required
  - Classification
    - Scientific domain should be always chosen with subcategory
    - Tags should be unique items list instead of input
  - Location
    - street should be required
    - postal code should be required
    - city should be required
    - country should be required
  - Maturity
    - Provider life cycle statuses should be select
    - Certifications should be unique items list instead of input
  - Other
    - Merril scientific domains should be always chosen with subcategory
    - ESFRI Type should be select instead of multi-select
- Bugs
  - [fix] on page reload (form error) array type fields has duplicated empty values

Steps to reproduce
1. Edit a provider or add the new provider
2. Make error while filling a form
3. Save the form
4. Page should be refreshed with notification
5. Repair made the error in the form
6. Save the form

What has happened?
The form can't be saved, because unique fields like tags, media, networks have empty duplicates.

What should happen?
The form should update the provider.

  - [fix] add logo value to form if it's exists

Steps to reproduce
1. Edit a provider with a logo
2. save the form

What has happened?
The form throws an error on the empty logo field.

What should happen?
The form should be saved properly.

  - [fix] tags string input should be replaced with a list of separate tags

Steps to reproduce
1. Edit a provider
2. Add tags

What has happened?
Tags can be added as a list of comma-separated values.

What should happen?
Tags should be possible to add as media or networks.

  - [fix] logo fetching should have a timeout

Steps to reproduce
1. Import with providers.beta.com

What is happening?
Imported Providers can have the logo in SVG/XML format. Conversion to PNG can hang workers.

What should happen?
Logo conversion should have a time limit after which should be interrupted. 

  - [fix] handle corrupted non ISO:3166-1 alpha2 codes of countries

Steps to reproduce
1. Import with providers.beta.com

What has happened?
[...]

Issue #1843, opened 2021-03-04
Missing resource logotype in MP, invalid images extensions of providers


lukasz-korzeniowski:
Instance: https://marketplace.eosc-portal.eu

**URLs to reproduce**
https://marketplace.eosc-portal.eu/services/pdb-tools-web 
https://marketplace.eosc-portal.eu/services/pan-learning-org-is-a-e-learning-platform-specifically-designed-to-provide-training-for-users-of-photon-and-neutron-sources
https://marketplace.eosc-portal.eu/services/amber-based-portal-server-for-nmr-structures-amps-nmr
https://marketplace.eosc-portal.eu/services/auto-rickshaw
https://marketplace.eosc-portal.eu/services/d4science-spatial-data-repository
https://marketplace.eosc-portal.eu/services/perfsonar-real-time-multi-domain-performance-monitoring
https://marketplace.eosc-portal.eu/services/edumeet-webbased-videoconferencing-platform
https://marketplace.eosc-portal.eu/services/eduvpn-access-your-institute-s-network-or-the-internet-using-an-encrypted-connection

**What has happened?**
Images don't display correctly.

**What should happened?**
Images should display from active directory.

**Additional context**
Providers images have incorrect extensions (are missing). It may be related with special chars in file names or too long active records path.

Example:
Page: https://marketplace.eosc-portal.eu/services/pdb-tools-web
Image src: https://marketplace.eosc-portal.eu/.../wenmr.pdb-tools_web
Image name: wenmr.pdb-tools_web **Missing image extension**

![image](https://user-images.githubusercontent.com/79973075/109966015-08689900-7cf0-11eb-8547-f795b612647c.png)

---

Some resource logos in MP are not uploaded on the website:

kmarszalek:
local with no db with import from prod
All: 288
1. http://localhost:5000/services/datacite-doi-registration-service
URL: https://support.datacite.org/docs/datacite-logo -> problem
2. http://localhost:5000/services/data-stewardship-wizard
URL: https://purl.org/ds-wizard/logo/full.png
ERROR - there was a unexpected problem processing image for ds-wizard.data_stewardship_wizard https://purl.org/ds-wizard/logo/full.png: redirection forbidden: https://purl.org/ds-wizard/logo/full.png -> http://suchanek.cloud/img/dsw_full_logo.png
3. http://localhost:5000/services/d4science-spatial-data-repository
URL:
[...]

PR #2154, opened 2021-06-10
[#1843] Remove corrupted logos

danielkryska:
## Context
Currently, we are allowing to save logos in a corrupted state. Especially images with extensions like i. e. `SVG` or expired URL to the binary source. In PR #2078 such occurrences will be not allowed and timeout. The next step will be logging all such errors (#2143) and handle them accordingly. For current state will be enough to remove crashed logos from Services.

### DIFF
--- a/lib/tasks/remove_corrupted_logos.rake
+++ b/lib/tasks/remove_corrupted_logos.rake
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+require "mini_magick"
+
+desc "Remove corrupted logos"
+task remove_corrupted_logos: :environment do
+  service = Service.where(slug: "gbif-spain-spatial-portal").first
+  service.logo.detach
+  service.save
+end


lib/tasks/remove_corrupted_logos.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "mini_magick"
+
+desc "Remove corrupted logos"
+task remove_corrupted_logos: :environment do
+  service = Service.where(name: "GBIF Spain Spatial Portal").first
+  service.logo.detach
+  service.save
+end
```
jswk, line 10: I reckon the name (or rather slug?) should be an argument for the task, as there is a number of services with incorrect logos.
You could pass it as rake arguments or an envvar. I actually recently learned that using rake arguments is not the most preferred way of doing that in rails (despite my using this mechanism recently...).

db/schema.rb
```diff
@@ -231,6 +231,29 @@
...
+    t.bigint "service_id", null: false
+    t.bigint "user_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.integer "issue_id"
+    t.integer "issue_status", default: 2, null: false
+    t.index ["service_id"], name: "index_orders_on_service_id"
+    t.index ["user_id"], name: "index_orders_on_user_id"
+  end
+
```
jswk, line 256: These are some peculiar changes, as you don't introduce any migration in this PR...

jswk (changes requested):

db/schema.rb
danielkryska: Removed

lib/tasks/remove_corrupted_logos.rake
danielkryska: This rake doesn't need to be generic. The system will handle in the proper way the next occurrences of `SVG` to `PNG` conversion, but there are still crashed images in the app. Like in the `GBIF Spain Spatial Portal` service.

danielkryska: And they need to be handled only once

jswk: Ok, add this rationale to the commit message then, as it's not something that can be inferred from code.

Maybe also renaming the task to something more specific would be good as well, for example `remove_gbif_spain_spatial_portal_corrupted_logo`.

jswk (changes requested):
Please consider this https://github.com/cyfronet-fid/marketplace/pull/2154#discussion_r649202588, and the PR would be fine by me.

lib/tasks/remove_corrupted_logos.rake
danielkryska: I'd like to keep the rake name as it is and remove it in the next sprint with `add_extension_to_images`. The commit message has been changed.

danielkryska: I've created a task to remove images rakes #2155

jswk (commented):
Was about to approve, but one more thing. :D

Add information about running this task in `docs/upgrade/3.15.0.md`, so that Wojt knows to run it.

danielkryska:
@wziajka has run the task from the command line in production to save additional work of adding and removing tasks from the repository.

The only influenced user was: https://marketplace.eosc-portal.eu/services/gbif-spain-spatial-portal

Issue #2144, opened 2021-06-08
Add UID field to project owner and message author


jswk:
The project owner in Ordering API should have UID field.
The same applies to message author.

PR #2162, opened 2021-06-14
[#2144] Add UID to author in Ordering API

jswk:
Extend the Ordering API by adding `uid` property to author: both for
project and messages.

Allow to specify the extra field for provider and mediator messages as
well. Add it as an optional field in POST request to /messages endpoint.

Add missing GET /api/v1/oms/{oms_id}/messages/{m_id} ref.

## How to test

SOMBO token: `LBPcykf7dpzi_GmfXLzs`.

Create a project and a project item. Add messages to both.

Using Ordering API:

Verify that when you GET a project then the field `owner` now has the field `uid`.
Verify that when you GET any of your messages, then the field `author` has the field `uid`.

Verify that if you POST a message to project or project_item and specify field `uid` in `author` then its value is persisted and returned in this message GET responses.

Closes: #2144.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ## [3.15.0-milestone]
 
 ### Added
+- Message author's and project owner's UID field in the Ordering API (@jswk)
 
 ### Changed
 - Improve consistency between Marketplace provider and Provider Component provider form and data model - improvement of User experience (@danielkryska)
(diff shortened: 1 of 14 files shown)


wujuu (approved):

Marcelinna:
should we allow uid to be completed by POST? Below it has been completed as "string", it does not match the value it is when sending the message via UI

image
image

jswk:
> should we allow uid to be completed by POST? Below it has been completed as "string", it does not match the value it is when sending the message via UI

The UID for user messages is taken from our User records. For the externally inserted messages (i.e. provider/mediator role) the value has to be set by an OMS, and it is a value that should point at message's author, which will be different than the user that created the message through MP.
So it is an expected behavior, that they don't match.

Marcelinna (approved):

PR #2165, opened 2021-06-17
Add e2e tests for filters on services page

Marcelinna:
Add e2e tests for filters on services page 

Add test for scientific domains, providers, dedicated for, platforms, order type, rating, resource availability filters

tests select the first element of each filter and check the filter has been turned on and the url is correct

### DIFF
--- a/test/system/cypress/integration/regression_tests/servce-page/filters/dedicated_for.spec.ts
+++ b/test/system/cypress/integration/regression_tests/servce-page/filters/dedicated_for.spec.ts
@@ -1,0 +1,17 @@
+describe("Filter decidated for", () => {
+  beforeEach(() => {
+    cy.visit("/services");
+  });
+  
+  it("should select filter", () => {
+    cy.get("[data-e2e='filter-tag']")
+      .should("not.exist")
+    cy.get("#collapse_target_users [data-e2e='filter-checkbox']")
+      .eq(0)
+      .click();
+    cy.location("href")
+      .should("include", "/services?target_users");
+    cy.get("[data-e2e='filter-tag']")
+      .should("be.visible")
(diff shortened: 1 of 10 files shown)


danielkryska (approved):

test/system/cypress/integration/regression_tests/servce-page/filters/dedicated_for.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+  
+  it("should select filter", () => {
+    cy.get("#collapse_target_users [data-e2e='filter-checkbox']")
+      .eq(0)
+      .click();
+    cy.location("href")
+      .should("include", "/services?target_users");
+    cy.get("[data-e2e='filter-tag']")
+      .should("be.visible");
+  });
```
jswk, line 14: I would consider adding an assertion first like this:
```
    cy.get("[data-e2e='filter-tag']")
      .should("not.be.visible");
```
So that you are actually sure that the element was made visible by the actions in the test.

This is about all the tests in the PR, actually.

jswk (commented):
Also, why aren't the tests grouped in a single file? They seem quite similar to me, and they could easily share the `beforeEach` section.

jswk:
> Also, why aren't the tests grouped in a single file? They seem quite similar to me, and they could easily share the `beforeEach` section.

And what is an answer to this?

Marcelinna:
> > Also, why aren't the tests grouped in a single file? They seem quite similar to me, and they could easily share the `beforeEach` section.
> 
> And what is an answer to this?

it is clearer to me, but this is just a subjective opinion

jswk (approved):

PR #2169, opened 2021-06-22
[#2029] remove elements from choices in the forms

### LINKED ISSUE #2029: it is not possible to remove elements from choice list
Tested on 3.11.0-beta.1
**Describe the bug**
it is not possible to remove elements from choice list if something has already been added.

**To Reproduce**
1. Go to '/profile'
2. Click "Edit" button next to Additional information 
3. Add elements to Categories of interests
image
4. Click on button "Save changes" (elements are added)
5. Click "Edit" button second time
6. Delete added Categories of interests
image
7. Click button "Save changes" 
8. See error - elements are still visible

**Expected behavior**
Elements are deleted from list

**Screenshots**
After deleting and saving, items e

image

**Additional context**
The same situation for Scientific domain choice list in /projects

image

goreck888:
Remove unnecessary `include_hidden: false` attribute in the forms.
It causes submitted form doesn't have information, that field is empty.

Fixes #2029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Fixed
 - Don't leak sensitive OMS fields in Offering API (@jswk)
 - Forms for questions about providers/resources are working properly (@goreck888)
+- Ability to clear a choice field in the forms (@goreck888)
 
 ## [3.14.0] 2021-06-11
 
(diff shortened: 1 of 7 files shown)


Marcelinna:
it isnt work for projects

Marcelinna (approved):

jswk (changes requested):
I would remove the `include_hidden: false` from other fields too (i.e. `service.language_availability` and `project.countries_of_partnership`) as it's also causing trouble there.

I'm on the fence regarding `profile.categories_updates` and `profile.scientific_domains_updates`, but I don't think these are necessary either.

goreck888:
> I would remove the `include_hidden: false` from other fields too (i.e. `service.language_availability` and `project.countries_of_partnership`) as it's also causing trouble there.
> 
> I'm on the fence regarding `profile.categories_updates` and `profile.scientific_domains_updates`, but I don't think these are necessary either.

Done.

jswk (approved):

Issue #2129, opened 2021-05-31
OMS trigger_url basic auth


jswk:
Introduce an optional method to use basic authorization on trigger calls.
For now, add basic auth scheme, but make it extensible, so that other schemes can be used.

I think an extra entity with class hierarchy may be a way to store this configuration information.

Make sure that the trigger job applies the security scheme.

PR #2175, opened 2021-06-24
[#2129] Add OMS trigger basic authorization

jswk:
Add OMS::Authorization::Basic to represent a basic authorization scheme.
It uses STI already to facilitate using other authorization schemas in
the future.

The specified authorization is also returned in the Ordering API.

###### Refactor and extend OMS trigger for authorization

Introduce a separate table to store trigger definitions.
Allow to specify a custom HTTP method to be fired.

Extend the Ordering API OMS output to include the new entity details.

Also, remove unnecessary `validates_associated` from OMS, so that
related service or provider entries are not re-validated on OMS
validation.
Validating them is dangerous, especially since providers are now saved
despite having errors.

Don't force existence of associated service or providers on OMS, as it
is valid to add them only after OMS creation.

# How to test

The main output from this PR is that triggers can now specify custom HTTP method and use basic authorization.
It would be easiest to test locally, where you can set-up a local mock server to log your requests.

I suggest using mockserver: it can be run with `docker run --rm -p 1080:1080 mockserver/mockserver`.
Then, navigate to http://localhost:1080/mockserver/dashboard, where you'll see the mockserver dashboard and all requests it receives.
Try it out by running `curl -X POST http://localhost:1080/test/something` from the cmd. The request should appear in the "Received Requests" panel, it will return 404 but it doesn't matter for the trigger execution.

To prepare and run a local instance of MP, execute:
```
rails db:drop db:create db:migrate
rails dev:prime
rails ordering_api:triggers_test_setup
bin/server
```
The server should now be available under http://localhost:5000/.

There are three OMSes in the setup: SOMBO, OMS2 and OMS3.
OMS2 has the trigger method changed to `:get` and OMS3 has basic auth `magic:mushroom` (base64 encoded `bWFnaWM6bXVzaHJvb20=`).

For service "EGI Cloud Compute" there are two extra offers created: `offer_oms_2` and `offer_oms_3`. As their primary OMSes they have OMS2 and OMS3, respectively.

Requests sent to OMSes should be of such kinds:
- SOMBO, `POST /oms1` without Authorization header,
- OMS2,` GET /oms2` without Authorization header,
- OMS3, `POST /oms3` with Authorization header equal to `Basic bWFnaWM6bXVzaHJvb20=` (you can decode it in shell like this: `echo "bWFnaWM6bXVzaHJvb20=" | base64 --decode -`).

Then as a user order the said offers and verify that triggers are sent to correct endpoints with correct authorization.

Closes: #2129.


wujuu (approved):

Marcelinna:
1. select oms2 offer
2. go to /summary
3. create new project 
4. I see first request 
image
5. click button send access request
6. I have four additional request
image

Marcelinna:
1. select oms3 offer
2. go to /summary
3. create new project
4. I see first request
image
5. click button send access request
6. I have four additional request
image

`POST  /oms3

{
"method":"POST"
"path":"/oms3"
"headers":{
"Accept":[...
]
"User-Agent":[...
]
"Authorization":[
0:
"Basic bWFnaWM6bXVzaHJvb20="
]
"Accept-Encoding":[...
]
"Host":[...
]
"Content-Length":[...
]
"Content-Type":[...
]
}
"keepAlive":true
"secure":false
}

POST  /oms1

{
"method":"POST"
"path":"/oms1"
"headers":{
"Accept":[...
]
"User-Agent":[...
]
"Accept-Encoding":[...
]
"Host":[...
]
"Content-Length":[...
]
"Content-Type":[...
]
}
"keepAlive":true
"secure":false
}

POST  /oms3

{
"method":"POST"
"path":"/oms3"
"headers":{
"Accept":[...
]
"User-Agent":[...
]
"Authorization":[
0:
"Basic bWFnaWM6bXVzaHJvb20="
]
"Accept-Encoding":[...
]
"Host":[...
]
"Content-Length":[...
]
"Content-Type":[...
]
}
"keepAlive":true
"secure":false
}

POST  /oms1

{
"method":"POST"
"path":"/oms1"
"headers":{
"Accept":[...
]
"User-Agent":[...
]
"Accept-Encoding":[...
]
"Host":[...
]
"Content-Length":[...
]
"Content-Type":[...
]
}
"keepAlive":true
"secure":false
}

POST  /oms1

{
"method":"POST"
"path":"/oms1"
"headers":{
"Accept":[...
]
"User-Agent":[...
]
"Accept-Encoding":[...
]
"Host":[...
]
"Content-Length":[...
]
"Content-Type":[...
]
}
"keepAlive":true
"secure":false
}`

Marcelinna:
When I use existing project and add offer 3 
image

Marcelinna:
please write that duplicate requests are not a problem and that when creating a new project, the request to oms1 is okay?

jswk:
https://github.com/cyfronet-fid/marketplace/pull/2175#issuecomment-870344968: this is correct
https://github.com/cyfronet-fid/marketplace/pull/2175#issuecomment-870349215: this is correct
https://github.com/cyfronet-fid/marketplace/pull/2175#issuecomment-870360523: this is correct

All such behaviours are ok, such project_items have two OMSes associated, so both are triggered (albeit with GET for OMS2 and with basic credentials for OMS3).
Also, shortly after a project_item is created it transitions to another state, so you get e.g. such calls: OMS1, OMS2, OMS1, OMS2 (two OMSes, two events, so 4 calls in total).

Marcelinna (approved):

Issue #2176, opened 2021-06-25
When revoking a token a placeholder value "revoked" is used


jswk:
**Describe the bug**
When revoking a token a placeholder value "revoked" is used.
This leads to two main consequences:
- only a single user can have a token revoked (column `authentication_token` has unique constraint)
- anyone can impersonate said user by setting `X-User-Token: revoked`

**To Reproduce**

1. Log in as user1.
2. Go to /api_docs.
3. Revoke token.
4. Sign out.
5. Go to /api_docs/swagger/index.html.
6. Enter token `revoked`. See that you don't get 401 but 200.
7. Sign in as user2.
8. Go to /api_docs.
9. Revoke token. You get a 500, and cannot revoke token anymore.

**Expected behavior**
1. Log in as user1.
2. Go to /api_docs.
3. Revoke token.
4. Sign out.
5. Go to /api_docs/swagger/index.html.
6. Enter token `revoked`. See that you get 401.
7. Sign in as user2.
8. Go to /api_docs.
9. Revoke token.

**Screenshots**
![image](https://user-images.githubusercontent.com/583197/123428760-71895a80-d5c6-11eb-9d55-ed47745c2093.png)

PR #2177, opened 2021-06-25
[#2176] Only allow to regenerate API token

jswk:
Setting authorization_token to `revoked` made authorizing as such user
possible in the API. Having a unique constraint on that column also
prevented several users to have the token revoked simulataneously.

Remove possibility to revoke token altogether, leaving only the ability
to regenerate it.

Nevertheless, handle the case of blank? token, as it will appear for
users which created their accounts before simple_token_authorization was
added to the project.

Fixes: #2176.

### DIFF
--- a/app/views/api_docs/show.html.haml
+++ b/app/views/api_docs/show.html.haml
@@ -27,7 +27,7 @@
             = link_to "Generate token", api_docs_path, method: :post, class: "btn btn-primary float-right"
         - else
           %p
-            = link_to "Revoke token", api_docs_path, method: :delete, class: "btn btn-outline-danger float-right"
+            = link_to "Regenerate token", api_docs_path, method: :post, class: "btn btn-outline-danger float-right"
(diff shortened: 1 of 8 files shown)


Marcelinna (approved):

app/views/api_docs/show.html.haml
```diff
@@ -8,26 +8,14 @@
...
-              %br
-              Click
-              %b Generate token
-              to get one.
-          - else
-            %p.mb-0
-              %i.fas.fa-check-circle
-              Your token is active:
-            %p.active
-              = current_user.authentication_token
```
wujuu, line 23: As there are users which indeed don't have a token yet (those that haven't been updated since the simple_token_auth gem was introduced), I think we should keep this part of the view as it looks much better than a simple blank where an authentication token should be

wujuu (changes requested):

Marcelinna (approved):

wujuu (approved):

PR #2181, opened 2021-06-29
[#2171] Hide token on API docs page

### LINKED ISSUE #2171: Hide token on API docs page
On API docs page, don't show the authentication token by default, but obfuscate it (i.e. show it as `***********`, try to match the length of the token).
Add a button/link "Show token" alongside it, which will make the token visible.

Make the API docs page public. If an unauthenticated user navigates to the page, then she should still see the obfuscated token, but the button should read "Login to access token", and it should trigger login upon click (and return back to the API docs page).

goreck888:
Allow unlogged users to get the api_docs page
After log in api docs shows only '*' signs
Added button to reveal the token
Closes #2171

### DIFF
--- a/app/controllers/api_docs_controller.rb
+++ b/app/controllers/api_docs_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApiDocsController < ApplicationController
-  before_action :authenticate_user!
+  before_action :authenticate_user!, only: :create
 
   def show
     @subsection = extract_subsection
--- a/spec/features/api_docs_spec.rb
+++ b/spec/features/api_docs_spec.rb
@@ -32,7 +32,21 @@ RSpec.feature "Api docs page" do
       expect(page).to have_text("This is the API of the EOSC Marketplace.")
     end
 
-    scenario "I can regenerate my token" do
+    scenario "I cannot see my token before clicking 'Show token' button and after clicking 'Hide token'", js: true do
+      visit api_docs_path
+
+      expect(page).to_not have_text(user.authentication_token)
+
+      find("#toggler").click
+
+      expect(page).to have_text(user.authentication_token)
+
+      find("#toggler").click
+
+      expect(page).to_not have_text(user.authentication_token)
+    end
+
+    scenario "I can regenerate my token", js: true do
       prev_token = user.authentication_token
 
       visit api_docs_path
@@ -41,6 +55,10 @@ RSpec.feature "Api docs page" do
       user.reload
 
       expect(user.authentication_token).to_not eq(prev_token)
+
+      save_page
+
+      find("#toggler").click
 
       expect(page).to have_text(user.authentication_token)
       expect(page).to have_link("Regenerate token")
@@ -51,7 +69,7 @@ RSpec.feature "Api docs page" do
         user.update_column(:authentication_token, nil)
       end
 
-      scenario "I can generate token" do
+      scenario "I can generate token", js: true do
         visit api_docs_path
         expect(page).to have_text("You don't have an authentication token yet")
         click_link("Generate token")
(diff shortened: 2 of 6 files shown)


app/controllers/api_docs_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ApiDocsController < ApplicationController
-  before_action :authenticate_user!
+  before_action :authenticate_user!, only: [:create, :destroy]
```
jswk, line 4: I've recently stripped the `:destroy` method, so you can omit it here.

jswk (changes requested):
I'd rather KISS. The premise of #2171 is to hide the token, so for example during screen sharing it isn't accidentally leaked. There is no need to have an extra controller for that and load the token dynamically.
I would rather render both stars and the token into the DOM, but the token should be initially hidden. Then when the user clicks 'Show token'/'Hide token' it be toggled.

You wouldn't have to create all the request-handling machinery in the `unhide_controller.js` and extra routes. The JS controller would be much simpler, as it would only toggle visibility of the elements.

wujuu (approved):

spec/features/api_docs_spec.rb
```diff
@@ -32,7 +32,17 @@
       expect(page).to have_text("This is the API of the EOSC Marketplace.")
     end
 
-    scenario "I can regenerate my token" do
+    scenario "I cannot see my token before clicking 'Show token' button", js: true do
+      visit api_docs_path
+
+      expect(page).to have_text("*" * user.authentication_token.size)
```
jswk, line 38: Verify that the token isn't visible too:
```ruby
expect(page).not_to have_text(user.authentication_token)
```

app/javascript/app/controllers/unhide_controller.js
```diff
@@ -0,0 +1,29 @@
+import { Controller } from 'stimulus'
+import Rails from '@rails/ujs'
```
jswk, line 2: I believe you no longer need this.

spec/features/api_docs_spec.rb
```diff
@@ -32,7 +32,17 @@
...
+
+      find("#toggler").click
+
+      expect(page).to have_text(user.authentication_token)
```
jswk, line 42: Click the toggler again and verify the token is obfuscated again.

spec/features/api_docs_spec.rb
```diff
@@ -83,10 +101,12 @@
   end
 
   context "as anonymous user" do
-    scenario "I don't see my api_docs page" do
-      visit root_path
+    scenario "I can achieve the api_docs page" do
```
jswk, line 104: => `I see api_docs page with login prompt`, `achieve` is incorrect in this context

jswk (changes requested):
Much better now, I would only do some cleanup.

jswk (commented):
Aaaaand, changelog!

Marcelinna:
Please change button show/hide token
image
https://projects.invisionapp.com/share/JXOZKC8WMDU#/screens/454770927_MyProfile_-_API_-_Token_Generated_Hidden

image
https://projects.invisionapp.com/share/JXOZKC8WMDU#/screens/441030132_MyProfile_-_API_-_Token_Generated

jswk (approved):

Marcelinna:
Please look at the video
Steps to reproduce:
1. Click button show token
2. Click on button - Authentication - token is hiden
3. Click on button Intruduction - **token is shown for a moment**

is it possible to remove the situation when the token appears for a moment?


https://www.awesomescreenshot.com/video/4378912?key=225e82806a7a60b7b3857f09967375a4

michal-szostak (approved):

Marcelinna (approved):

PR #2195, opened 2021-07-12
[#2186] Provider can navigate from PPP to PC

### LINKED ISSUE #2186: As a provider I can trigger my profile edition from the MP Provider Presentation Page, so i can easily improve my profile once i spot the problem in the PPP
The implementation should follow the rules implemented in the resource presentation page.

this functionality applies to Provider Presentation Page

kmarszalek:
Provider can navigate from the Provider Presentation Page
to the provider dashboard of the Provider Component

To test you need to be provider data_admin.
Add yourself in backoffice as data_admin then go to providers presentation page.

Closes #2186

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -162,6 +162,10 @@ class Provider < ApplicationRecord
     )
   end
 
+  def administered_by?(user)
+    data_administrators.where(email: user.email).count.positive?
+  end
+
   private
     def remove_empty_array_fields
       array_fields = [
--- a/app/views/providers/_header.html.haml
+++ b/app/views/providers/_header.html.haml
@@ -16,6 +16,8 @@
       = link_to _("Browse resources"),
                     services_path(providers: provider.id),
                     class: "btn btn-primary d-block mb-3", "data-e2e": "btn-browse-resource"
+      - if user_signed_in? && policy(provider).data_administrator?
+        = render "providers/header/drop_down_menu", provider: provider
 .row.service-links
   .col-12.col-lg-9.row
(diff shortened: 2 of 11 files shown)


Marcelinna (approved):

app/views/providers/_header.html.haml
```diff
@@ -16,6 +16,24 @@
...
+            %li
+              %a.dropdown-item.dropdown-details{ href:
+                Mp::Application.config.providers_dashboard_url + "/dashboard/#{provider.pid}/stats",
+                target: :_blank, "data-probe": "" }
+                = _("Provider dashboard")
+            %li
+              %a.dropdown-item.dropdown-details{ href:
+                Mp::Application.config.providers_dashboard_url + "/provider/update/#{provider.pid}",
+                target: :_blank, "data-probe": "" }
+                = _("Edit provider details")
```
wujuu, line 36: I would add this to a partial

app/models/provider.rb
```diff
@@ -162,6 +162,10 @@ def set_default_logo
     )
   end
 
+  def administered_by?(user)
+    data_administrators.where(email: user.email).count.positive?
+  end
+
```
wujuu, line 168: Since you are adding this method here, I would just delegate the `administrate_by?` method in `Service` to here. Like this:

In `app/models/service.rb` lines 198 to 200
```
def administered_by?(user)
    resource_organisation.administered_by?(user)
  end
```

wujuu (changes requested):
Looks OK, just a few minor changes

wujuu (approved):

PR #2212, opened 2021-07-22
[#1999] Provider search links to the provider page

### LINKED ISSUE #1999: result in main searchbar should only show a matched provider in provider section
Tested on v3.10.0-beta.1
**Describe the bug**
after entering the word in main searchbar, the result in provider section should be the name of the provider without any services assigned

**To Reproduce**
1. Go to '/services'
2. Enter in main searchbar "EGI"

**Expected behavior**
in section Provider should only be visible name of matched provider,  here "EGI Foundation". When the user click on the provider name shoulbe be redirected to provider page. 

**Screenshots**
image

goreck888:
Change link in autocomplete for providers
to the provider_path
Fixes #1999

### DIFF
--- a/spec/features/search_spec.rb
+++ b/spec/features/search_spec.rb
@@ -40,17 +40,43 @@ RSpec.feature "Service searching in top bar", js: true do
     expect(page).to_not have_css(".categories", text: "Looking for: DDDD Something")
   end
 
-  scenario "redirect when selecting service_id by autocomplete controller", js: true, search: true do
+  scenario "redirect to service path when selecting service_id by autocomplete controller", js: true, search: true do
     service = create(:service)
     fill_in "q", with: service.name
+    expect(page).to have_css("#-option-0")
     find(:css, "li[id='-option-0']").click
+    expect(page).to have_content(service.name)
     expect(current_path).to eq(service_path(service))
   end
 
-  scenario "redirect when selecting service_id by autocomplete controller", js: true, search: true do
+  scenario "doesn't show unpublished records", js: true, search: true do
+    draft_service = create(:service, name: "Awesome 1", status: :draft)
+    published_service = create(:service, name: "Awesome 2")
+    fill_in "q", with: draft_service.name.truncate(5)
+
+    expect(page).to_not have_content(draft_service.name)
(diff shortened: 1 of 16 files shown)


jswk (changes requested):
There is a regression regarding service authorization.
It is now possible to find deleted or draft services in all the search bars, not only in backoffice.
image
Clicking it leads to:
image

Marcelinna:
Master
image

PR
image

goreck888:
> There is a regression regarding service authorization.
> It is now possible to find deleted or draft services in all the search bars, not only in backoffice.
> image
> Clicking it leads to:
> image

Fixed.

goreck888:
> Master
> image
> 
> PR it is problem with services
> image

We don't have too much influence on what searchkick returns. I try to change the limit of results to 10, it should be better.

spec/features/search_spec.rb
```diff
@@ -40,19 +40,45 @@
...
+  end
+
+  scenario "redirect to provider path when selecting provider_id by autocomplete controller", js: true, search: true do
+    provider = create(:provider)
+    fill_in "q", with: provider.name
+    expect(page).to have_css("#-option-0")
+    find(:css, "li[id='-option-0']").click
+    expect(page).to have_content(provider.name)
+    expect(current_path).to eq(provider_path(provider))
+  end
```
jswk, line 68: Something is off with this test failing. I cannot pinpoint a concrete place, but when I run the test without headless option for webdriver, it's impossible to find the created provider. Either the provider isn't synced to ES or the search query in autocomplete.rb is done incorrectly?
What's strange is that providers appear in development, or on the testing docker instance, but not in this test.

jswk (changes requested):
These two test issues are to be addressed, otherwise only the changelog is missing.

spec/features/search_spec.rb
goreck888: Resolved. The problem was the created provider wasn't reindexed. Now this test pass.

jswk (changes requested):
Aaand, changelog!

jswk (approved):

Marcelinna (approved):

Issue #2226, opened 2021-07-27
Improper search_data json format in /recommendations request


wujuu:
**Describe the bug**
Currently the search_data details require that all sub-fields (like `providers`) are arrays, however they are integers sometimes, which causes validation error on the recommender system part.

**Possible problem**:
`ids.instance_of?(Array) ? ids.map(&:to_i) : ids.first.to_i` map applied to all search_data fields in `app/controllers/concerns/service/recommendable.rb`

michal-szostak:
Blocked until @wujuu returns

PR #2227, opened 2021-07-28
[#2226] Fix invalid search_data format

wujuu:
Should fix this kind of recommender validation error:

```
2 is not of type 'array'

On instance['search_data']['providers']:
    2
```


Marcelinna:
I try reproduce problems which by known https://docs.cyfronet.pl/display/FID/Dzialanie+Rekomendera+-+bugi. 
1. Recomendation still disappear when I change sort
image

Marcelinna:
if there is only one service left from the filter results, then the recommendations should be visible? (I take into account the case with two recommended services)
I think it shouldn't be as in the case of 0?

image

Marcelinna:
Recomendation disapear when I select Access physical & eInfrastructures category (unlogin user)
image

Marcelinna:
The recommendation will disappear after changing the number of displayed items from 10 to 20
image

Marcelinna:
Disappearance of recomnedation after changing the page on the pagination bar - looks fixed

Marcelinna:
The recomnedation disappear after entering the text in search
image

Marcelinna:
The recommendation disappear when you turn on the filters
image

Marcelinna:
There was no error in sentry, except
[28/Jul/2021 13:19:19] code 400, message Bad request syntax ('\x03\x00\x00\x13\x0eà\x00\x00\x00\x00\x00\x01\x00\x08\x00\x03\x00\x00\x00')

kmarszalek (approved):

Marcelinna:
Steps to reproduce 
1. Go to https://marketplace-2.docker-fid.grid.cyf-kr.edu.pl/
2. Click on Engineering & Technology
3. Recommendations are visible
image
4. Go back to the previous page<---
5. Click on Click on Engineering & Technology
6. See problem - Recommendations aren't visible
image

app/controllers/concerns/service/recommendable.rb
```diff
@@ -25,13 +25,13 @@ module Service::Recommendable
       if ids.instance_of?(Array)
         ids.map(&:to_i) + ids.map { |id| ScientificDomain.find(id).descendant_ids }.flatten
       else
-        ids.first.to_i
+        Array(ids.first.to_i) | Array(ScientificDomain.find(ids).descendant_ids.flatten)
```
jswk, line 28: Why use `|` (https://ruby-doc.org/core-2.3.1/Array.html#method-i-7C) here? Wouldn't `+` (https://ruby-doc.org/core-2.3.1/Array.html#method-i-2B) suffice?

jswk: It would also be in line with how it's handled in the other branch of the if.

jswk, line 28: Another thing: why don't you pass `ids.first` to `ScientificDomain#find`, the current construct may not work.
BTW, I'd add a test to verify the logic of this transformer. Even a simple unit test directly calling the lambda, as the code path may not be easily reachable otherwise.

It would be also clearer what do we expect in the second branch if we had a test for this. Since it's not an array, but still it has a `first` method, a hash maybe? Definitely not an integer, as an integer doesn't have `first` method. Maybe string, but in this case `first` returns the first character, and it's probably not what we want?

Michal-Kolomanski: I have pushed a commit that addresses your comment. You are right that inline `28` `| operator` is not necessary but inline  `26` `| operator` should be used to avoid repetitions.

Michal-Kolomanski: Duplicates were not a concern in categories, but they were in scientific domains.

Marcelinna (approved):

Marcelinna:
changelog?

app/controllers/concerns/service/recommendable.rb
```diff
@@ -23,15 +23,15 @@ module Service::Recommendable
     geographical_availabilities: ->(name) { Country.convert_to_regions_add_country(name) },
     scientific_domains: ->(ids) do
       if ids.instance_of?(Array)
-        ids.map(&:to_i) + ids.map { |id| ScientificDomain.find(id).descendant_ids }.flatten
+        ids.map(&:to_i) | ids.map { |id| ScientificDomain.find(id).descendant_ids }.flatten
       else
-        ids.first.to_i
+        Array(ScientificDomain.find(ids.to_i).id) + Array(ScientificDomain.find(ids.to_i).descendant_ids.flatten)
       end
```
jswk, line 29: What about the part about the unit test though https://github.com/cyfronet-fid/marketplace/pull/2227#discussion_r800573568?
Few simple examples would make it clearer what is expected by this lambda.

Michal-Kolomanski: I'm working on the unit test right now. I wanted to get the code out as soon as possible so Marcelina could test it.

jswk: Great, I've got a re-review so I assumed the current state was the intended outcome. Thanks for the update.

Michal-Kolomanski (approved):

PR #2228, opened 2021-07-28
[#2103] Fix creating offers in ordering configuration

### LINKED ISSUE #2103: no possibility to add an offer by ordering configuration for the service with no offert status
3.12.0-beta.2
Tested 
**Describe the bug**
if the website does not have any offer, even the default one, it is not possible to add an offer by ordering configuration, the problem does not exist for backoffice

**To Reproduce**
1. Select resource for which you have admin privileges and delete all offers via API 
2. Go to resource page /services/{service.name}
3. Click button "Manage resource" and select "Set parameters and offers"
4. See error - user see information _You are not authorized to see this page_, but user has admin privileges and service is Published

**Expected behavior**
User should have possibility add offer

**Screenshots**
image

**Additional context**
View in Backoffice, user has possibility add new offert
image

goreck888:
Fix a possibility to create an offer in the ordering_configuration
panel when a resource has no offers
Fixes #2103

### DIFF
--- a/app/views/services/ordering_configurations/_about.html.haml
+++ b/app/views/services/ordering_configurations/_about.html.haml
@@ -3,8 +3,7 @@
     %main.col-12.col-xl-9.pr-5.mb-4{ "data-shepherd-tour-target": "service-about" }
 
       = markdown(service.description)
-      - if policy(service).order? && !offers&.empty?
-        = render "layouts/common_parts/services/offers", offers: offers, service: service
+      = render "layouts/common_parts/services/offers", offers: offers, service: service
 
     %sidebar.col-12.col-xl-3{ "data-shepherd-tour-target": "service-classification" }
       - service_sidebar_fields.each do |group|
--- a/spec/features/ordering_configuration_spec.rb
+++ b/spec/features/ordering_configuration_spec.rb
@@ -83,6 +83,34 @@ RSpec.feature "Services in ordering_configuration panel" do
       expect(page).to_not have_link("Delete Offer")
     end
 
+    scenario "I can add an offer if none exists", js: true do
+      service_without_offers = create(:service, resource_organisation: provider, offers: [])
+
+      visit service_ordering_configuration_path(service_without_offers)
+
+      expect(page).to have_content("Add new offer")
+
+      click_on "Add new offer"
+
+      expect {
+        fill_in "Name", with: "new offer 1"
+        fill_in "Description", with: "test offer"
+        find("li", text: "Input").click
+        find("#attributes-list-button").first("svg").click
+
+        within("div.card-text") do
+          fill_in "Name", with: "new input parameter"
+          fill_in "Hint", with: "test"
+          fill_in "Unit", with: "days"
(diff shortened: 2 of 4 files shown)


app/policies/ordering_configuration/offer_policy.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class OrderingConfiguration::OfferPolicy < ApplicationPolicy
+class OrderingConfiguration::OfferPolicy < OrderingConfigurationPolicy
```
jswk, line 3: Wouldn't you also want to override the `show?` method to return `offer_editor?`. `offer_editor?` captures a little broader category of users than just `record.administered_by?(user)` from `OrderingConfigurationPolicy`.
Also, for this class, `record.administered_by?` won't be a method, as it is only defined on Service, not on offer. This leads me to think changing the parent here isn't necessary at all, and will just create confusion.

app/views/services/ordering_configurations/_about.html.haml
```diff
@@ -3,8 +3,7 @@
     %main.col-12.col-xl-9.pr-5.mb-4{ "data-shepherd-tour-target": "service-about" }
 
       = markdown(service.description)
-      - if policy(service).order? && !offers&.empty?
```
jswk, line 6: Why was the `policy(service).order?` applied here? Is it safe to remove it?

spec/features/ordering_configuration_spec.rb
```diff
@@ -83,6 +83,34 @@
       expect(page).to_not have_link("Delete Offer")
     end
 
+    scenario "I can add an offer if no one exists", js: true do
```
jswk, line 86: I'd phrase it "if none exists", to be correct

jswk (commented):
I'm also thinking, there seems to be a discrepancy between `OrderingConfigurationPolicy` and `OrderingConfiguration::OfferPolicy`. In particular, the former is more restrictive than the latter. Do you know if it is a desired behaviour?

Also, changelog. ;)

app/policies/ordering_configuration/offer_policy.rb
goreck888: Ok

spec/features/ordering_configuration_spec.rb
goreck888: Ok.

app/views/services/ordering_configurations/_about.html.haml
goreck888: I think it's safe, moreover, it's necessary :)

jswk: Okay, what's the reason for this?

goreck888: As I suppose we assumed that default offer is indelible. Now we can remove all offers via offering API.

jswk (approved):

Marcelinna (approved):

PR #2246, opened 2021-08-09
[#2005] Store query params in comparison view

### LINKED ISSUE #2005: problem with comparison services functionality
**Describe the bug**
the service comparison doesn't always work correctly, the bar with services added to the comparison shows the wrong number of services, sometimes the services are on the bar although they are not selected. After clear all services from comparison /comparisons main filter is changed from BEST MATCH to by-name:A-Z it is desirable?

back from /favorites to /services also change filter from BEST MATCH to by-name:A-Z - please check it out

**To Reproduce*
For steps:
https://www.awesomescreenshot.com/video/3432706?key=161be2fd8cfe115a9625e3fcf16944b2

**Expected behavior**
The service comparison work correctly

**Screenshots**
GEOSS Web Portal is on comparison bar, although it is not selected
image

**Additional context**

For favourites and comparison, there are 'bugs' for multiple views and theirs states.
We don't have a mechanism to synchronise their states.
Hard to reproduce consistently.

goreck888:
When user navigates from resources list
to the comparison view, query params are saved now
Fixes #2005

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Added
 - First and last name to Ordering API project owner (@michal-szostak, @jswk)
+- Storing resource query parameters by visiting the comparison page (@goreck888)
 
 ### Changed
 
--- a/app/controllers/comparisons_controller.rb
+++ b/app/controllers/comparisons_controller.rb
@@ -2,15 +2,16 @@
 
 class ComparisonsController < ApplicationController
   include Service::Comparison
+  before_action :query_params, only: [:show, :destroy]
 
   def show
     @services = Service.where(slug: session[:comparison])
     if @services.blank?
       if params[:fromc]
         category = Category.find_by(slug: params[:fromc])
-        redirect_to category_services_path(category)
+        redirect_to category_services_path(category, params: @query_params)
       else
-        redirect_to services_path
+        redirect_to services_path(params: @query_params)
       end
     end
   end
@@ -18,12 +19,16 @@ class ComparisonsController < ApplicationController
   def destroy
     session[:comparison] = []
     respond_to do |format|
-      format.html { redirect_to services_path }
+      format.html { redirect_to services_path(@query_params) }
       format.js { render_json }
     end
   end
 
   private
+    def query_params
+      @query_params = session[:query] || {}
+    end
+
     def render_json
       render json: { data: @services&.map(&:slug), html: bottom_bar }
     end
--- a/app/controllers/comparisons_controller.rb
+++ b/app/controllers/comparisons_controller.rb
@@ -2,7 +2,7 @@
 
 class ComparisonsController < ApplicationController
   include Service::Comparison
-  before_action :query_params, only: [:show, :destroy]
+  before_action :load_query_params_from_session
 
   def show
     @services = Service.where(slug: session[:comparison])
@@ -25,7 +25,7 @@ class ComparisonsController < ApplicationController
   end
 
   private
-    def query_params
+    def load_query_params_from_session
       @query_params = session[:query] || {}
     end
 
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -120,6 +120,8 @@ module Service::Searchable
             params["#{filter.field_name}-all"] unless params["#{filter.field_name}-all"].blank?
       end
       session[:query][:q] = params[:q] unless params[:q].blank?
+      session[:query][:sort] = params[:sort] unless params[:sort].blank?
+      session[:query][:per_page] = params[:per_page] unless params[:per_page].blank?
     end
 
     def filter_classes
--- a/app/controllers/concerns/service/searchable.rb
+++ b/app/controllers/concerns/service/searchable.rb
@@ -122,6 +122,7 @@ module Service::Searchable
       session[:query][:q] = params[:q] unless params[:q].blank?
       session[:query][:sort] = params[:sort] unless params[:sort].blank?
       session[:query][:per_page] = params[:per_page] unless params[:per_page].blank?
+      session[:query][:page] = params[:page] unless  params[:page].blank?
     end
 
     def filter_classes
--- a/app/views/services/_index.html.haml
+++ b/app/views/services/_index.html.haml
@@ -1,4 +1,6 @@
-.container{ "data-controller": "favourite" }
+- effective_show_comparison = local_assigns.fetch(:show_comparison, true)
+- effective_show_recommendations = local_assigns.fetch(:show_recommendations, true)
+.container
   .row
     .col-lg-3.mb-5
       = render "services/nav/categories", current: category,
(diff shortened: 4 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -36,6 +37,10 @@ def welcome_popup
       end
     end
 
+    def load_query_params
+      @query_params = session[:query]
+    end
```
jswk, line 42: Why is this necessary? You extract the params in comparisions_controller manually anyway.

I would define such before_action directly in comparisions_controller, as other controllers don't need this information (the `@query_params` variable).

app/controllers/comparisons_controller.rb
```diff
@@ -5,20 +5,22 @@ class ComparisonsController < ApplicationController
 
   def show
     @services = Service.where(slug: session[:comparison])
+    @query_params = session[:query] || {}
```
jswk, line 8: This would go.
You could keep the default-setting in before_action though.

app/controllers/comparisons_controller.rb
```diff
@@ -5,20 +5,22 @@ class ComparisonsController < ApplicationController
...
       else
-        redirect_to services_path
+        redirect_to services_path(params: @query_params)
       end
     end
   end
 
   def destroy
     session[:comparison] = []
+    @query_params = session[:query_params]
```
jswk, line 21: This seems not to work since you use `:query` session key in other places, and not `:query_params`.

jswk (changes requested):
There is also a problem with multi-valued filters, yet it seems to be a broader problem.

1. Go to services.
2. Select two sci-domains for filtering (path: `/services?scientific_domains[]=1&scientific_domains[]=26`).
3. Change ordering, for example to "Z-A" (path: `/services?scientific_domains[]=26&sort=-name`). The filter is now only for a single sci-domain.
4. If you try to re-add it, then the ordering is cleared (path: `/services?scientific_domains[]=1&scientific_domains[]=26`).

This is also present in master, so it should rather be handled separately.
Nevertheless, if you enter a correct path by hand into the address bar then the correct results are shown.

This seems to be related to what is happening here:
1. Go to services.
2. Select a sci-domain and custom sort (path: `/services?scientific_domains[]=1&sort=-name`)
3. Select two services for comparison, click compare.
4. Click "Add next resource", you are redirected to `/services?scientific_domains[]=1`, without sort.
5. Alternatively, if in the comparison view (step 3) you click "Clear all", then filter is cleared totally (e.g. you go to `/services`).

I would expect the sort to be saved too, though I may be mistaken and it's correct.
Multiply valued filter should be retained too, although this may be a deeper problem with searchable, may be left for later.
Also, in the code you add the query handling to the `destroy` method, yet it doesn't work at all. I would either make it work, or assume you don't handle it at all (when using `destroy` user anyway goes back to search, she may enter the filters again).

@Marcelinna what's your view on this?

Marcelinna:
testing by adding and removing resource from comparison, on screen see that Alien and Invasive Species Virtual Research Environment is added to comparison, despite checkbox is not selected.
image

jswk (changes requested):
The comment about page is blocking, other two cosmetic.

jswk:
BTW, how about we don't store and restore the query params at all here, but open the comparison in a new tab, as other webpages sometimes do? See https://x-kom.pl for an example of this behaviour.
In that case state is stored in a separate tab, just like this. And a user can navigate between tabs as she wishes.

app/controllers/comparisons_controller.rb
```diff
@@ -2,28 +2,33 @@
...
     session[:comparison] = []
     respond_to do |format|
-      format.html { redirect_to services_path }
+      format.html { redirect_to services_path(@query_params) }
       format.js { render_json }
     end
   end
 
   private
+    def query_params
```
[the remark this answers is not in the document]
goreck888: Ok

app/controllers/concerns/service/searchable.rb
```diff
@@ -120,6 +120,8 @@ def store_query_params
             params["#{filter.field_name}-all"] unless params["#{filter.field_name}-all"].blank?
       end
       session[:query][:q] = params[:q] unless params[:q].blank?
+      session[:query][:sort] = params[:sort] unless params[:sort].blank?
+      session[:query][:per_page] = params[:per_page] unless params[:per_page].blank?
```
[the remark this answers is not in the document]
goreck888: Done

app/views/services/_index.html.haml
```diff
@@ -1,4 +1,4 @@
-.container{ "data-controller": "favourite" }
+.container
```
jswk, line 1: Later on you have the call `if (not defined? show_comparison) || show_comparison`, I would assign it here such as:
```haml
- effective_show_comparison = local_assigns.fetch(:show_comparison, true)
```
And later on use just the `effective_show_comparison`.

app/views/services/_index.html.haml
```diff
@@ -7,7 +7,7 @@
                                           services_count: services_count
       = render "services/filters", filters: filters
 
-    .col-lg-9
+    .col-lg-9{ "data-controller": "comparison favourite" }
```
jswk, line 10: I believe the `comparison` controller shouldn't be loaded if `effective_show_comparison` is false.
The line would read:
```
     .col-lg-9{ "data-controller": "favourite" + (effective_show_comparison ? " comparison" : "") }
```

app/views/services/_index.html.haml
```diff
@@ -40,4 +40,7 @@
 
         #popup-modal.modal.show{ "data-target": "favourite.popup", tabindex: "-1",
                                   role: "dialog", aria: { hidden: "true" } }
-
+      - if (not defined? show_comparison) || show_comparison
```
jswk, line 43: And you could use the `effective_show_comparison` here too.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix incorrect classification value on PC sync - better User Experience (@danielkryska)
 - Alphabetical sorting by name for the Resources (@kmarszalek, @goreck888)
 - Fix new resource preview error - show default eosc icon, imporved admin experience (@danielkryska)
+- Storing resource query parameters by visiting the comparison page (@goreck888)
```
jswk, line 30: I'd say it's rather an added feature.

app/controllers/comparisons_controller.rb
```diff
@@ -2,28 +2,33 @@
...
       else
-        redirect_to services_path
+        redirect_to services_path(params: @query_params)
       end
     end
   end
 
   def destroy
-    session[:comparison] = []
+    session[:comparison] = @services = []
```
jswk, line 20: Setting `@services` to an empty array seems unnecessary here.
There is no view associated with this method anyway.

jswk (approved):

Marcelinna (approved):

CHANGELOG.md
```diff
@@ -27,7 +28,10 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix incorrect classification value on PC sync - better User Experience (@danielkryska)
 - Alphabetical sorting by name for the Resources (@kmarszalek, @goreck888)
 - Fix new resource preview error - show default eosc icon, imporved admin experience (@danielkryska)
+<<<<<<< HEAD
 - CSS styles (@jarekzet)
+=======
+>>>>>>> Move comparison data-controller to service box partial
```
jswk, line 34: Please resolve this conflict before pushing

jswk (changes requested):

jswk (approved):

PR #2247, opened 2021-08-10
[#1961] Fix sorting for resources

### LINKED ISSUE #1961: Resources should be sort alphabetically
**Describe the bug**
Resources are sorting lexicographically

**To Reproduce**
1. Go to '/services'
2. Sort resource by A-Z or Z-A

**Expected behavior**
Resources should be sort alphabetically

**Screenshots**
Sort by Z-A
obraz

**Additional context**
relevant doc: http://www.technocratsid.com/sort-strings-alphabetically-rather-than-lexicographically-in-elasticsearch/

kmarszalek:
- Fixed sorting by name
- Added test that checks sorting with lower/upper cases

Elasticsearch follows ASCII sort order, the bytes that represent
capital letters have lower ASCII values than that of lowercase letters.
By changing name to name.downcase in search_data we can sort resources
in alphabetical order regardless of their ASCII codes.

Closes #1961

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Disable reloading search autocomplete list after click on its element (@goreck888)
 - Keep search input after redirect from autocomplete (@goreck888)
 - Fix incorrect classification value on PC sync - better User Experience (@danielkryska)
+- Alphabetical sorting by name for the Resources (@kmarszalek, @goreck888)
 
 ### Security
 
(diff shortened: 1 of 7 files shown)


jswk (approved):

Marcelinna (approved):

jswk (changes requested):
The change in search_data propagates to the autocomplete unfortunately. (Only the "Resources" section, "Providers" and "Offers" stay correct.)
image

jswk (changes requested):
There's sort_options in backoffice/services_controller.rb too, it should have the sorting field altered too.

jswk (approved):

PR #2256, opened 2021-08-12
[#2236] Fix autocomplete unnecessary reloading

### LINKED ISSUE #2236: Autocomplete results section strange behaviour
**Describe the bug**
Autocomplete starts showing different results after clicking on one of the results (and before teh system redirects to the chosen result)

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'the starting  page.
2. Insert something in the search bar. Wait for the results.
3. Remember the result list, click on one of the results.
4. See error - notice that the result list has changed while the system is redirecting to the chosen result
5. In the end, user is redirected to the page of choice

**Expected behaviour**
After selecting autocomplete option search bar should be filled with selected option but new autocomplete window should **not** be displayed. After reload query in search input should be preserved

**Screenshots**
After inserting '100' in the search bar:
image
After clicking 100Trust IT provider
image

**Additional context**

Manifests itself under Chromium, not in Firefox.

goreck888:
Fix reloading autocomplete list
after clicking on an item
After page reload search input
is filled with previously clicked item label

**NOTICE**
Fixed redirect_to `/null` isn't merged into master yet.

Fixes #2236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Proper displaying questions in the resource opinion form (@goreck888)
 - Service owners cannot edit categories in the backoffice view when service upstream is other than MP (@wujuu)
 - Disable reloading search autocomplete list after click on its element (@goreck888)
+- Keep search input after redirect from autocomplete (@goreck888)
 
 ### Security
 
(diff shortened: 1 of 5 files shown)


kmarszalek (changes requested):
works fine. code looks ok. do you plan to add some tests?

jswk (commented):
I would suggest splitting the changes into two commits: one for reload and another for keeping input after redirect. SRP, but with code chunks.

The commits could read:
```
[#2236] Disable reloading search autocomplete list after click on its element

Moving focus to the input triggered results reloading, which wasn't expected by the users.
Remove the call doing that in `commit` method, so that neither clicking by mouse, clicking Enter or Tab moves focus to input - as desired.
```

```
[#2236] Keep search input after redirect from autocomplete

Pass the input query string from `params["q"]` on to provider or service, so that it is displayed consistently in the target pages.
```

I'd find such description easier to follow later on when inspecting historical changes and looking for their rationale.

jswk (approved):

Issue #2085, opened 2021-05-18
New provider is automatically added to different classifications (created by PC)


Marcelinna:
Tested v3.12.0-beta.2
**Describe the bug**
When user create provider and doesnt fill in the fields **Classification (Scientific Categorisation)**, new created provider in MP is added to 5 Scientific Domain automatically.

**To Reproduce**
1. Create new provider in PC (https://beta.providers.eosc-portal.eu), don't fill in the fields **Classification (Scientific Categorisation)**, they are not required
2. See created provider in MP

**Expected behavior**
provider shouldn't be added to different classifications automatically if fields arent filled

**Screenshots**
![image](https://user-images.githubusercontent.com/59872096/118608262-cc56b780-b7b9-11eb-8ec7-181ffb9cd5ec.png)

**Additional context**

PR #2258, opened 2021-08-12
[#2085] Fix incorrect classification value on PC sync

danielkryska:


Marcelinna (approved):

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Secure the public service endpoints when it's deleted or it's draft (@danielkryska)
 - Fit a map not aligned to a parent element - the all countries in the service site map should be visible (@danielkryska)
 - Fix the provider backoffice form - better User Experience (@danielkryska)
+- Fix incorrect classification value on PC sync - better User Experience (@danielkryska)
```
jswk, line 43: Please add the changelog in the correct section (i.e. in `3.19.0-milestone`)

app/models/concerns/importable.rb
```diff
@@ -10,7 +10,9 @@ def map_categories(categories)
   end
 
   def map_scientific_domains(domains)
-    ScientificDomain.where(eid: domains)
+    domains.present? && domains.length > 0 ?
```
jswk, line 13: It's unnecessary to check for the length. The `present?` call already takes it into account, see:
```
$ rails c
Loading development environment (Rails 6.0.3.7)
[1] pry(main)> nil.present?
=> false
[2] pry(main)> [].present?
=> false
[3] pry(main)> [1].present?
=> true
```

app/services/service/pc_create_or_update.rb
```diff
@@ -82,7 +82,7 @@ def call
     Rails.logger.warn "#{e} Message arrived, but service is not updated. Message #{@eosc_registry_service}"
     if mapped_service.present? && mapped_service&.sources&.first.present?
       source = mapped_service&.sources&.first
-      source.update(errored: check_service.errors.messages)
+      source.update(errored: check_service&.errors&.messages)
```
jswk, line 85: `check_service` is a boolean as it's set like this `check_service = Service.new(service_hash).invalid?`, so it won't have more state.
I would just set the errored to `check_service`, i.e.:
```
       source.update(errored: check_service)
```

jswk (changes requested):

jswk (approved):

Issue #2117, opened 2021-05-26
RWD problems Part 2


Marcelinna:
**Describe the bug**
Tikect is opened and will be completed

**Screenshots**
1. 
- [x] QA
- [x] DEV

feedback panel: /services/{service.name}
![image](https://user-images.githubusercontent.com/59872096/119638958-73210080-be17-11eb-9941-8512eec332f7.png)
feedback panel: /profile
![image](https://user-images.githubusercontent.com/59872096/119639159-a06dae80-be17-11eb-9473-0a862ce4414a.png)

2.
- [x] QA
- [x] DEV

home page
![image](https://user-images.githubusercontent.com/59872096/125592712-35cf1601-b92d-43f6-885b-6f4964071d3c.png)

3.
- [x] QA
- [x] DEV
/communities
/providers

![image](https://user-images.githubusercontent.com/59872096/125593063-2d8b0b75-aa4c-45f2-8da1-df1e0bf38825.png)
![image](https://user-images.githubusercontent.com/59872096/125593120-887e3e67-6bc4-4ca9-b7b5-538f01b793a1.png)
![image](https://user-images.githubusercontent.com/59872096/125593216-8c6312fe-b05d-49a0-908e-df565e6abe21.png)

4.
- [x] QA
- [x] DEV
/providers/{provider.name}
![image](https://user-images.githubusercontent.com/59872096/125593332-0cc10d2e-0649-4030-bbee-cff623032ebf.png)

5. 
/admin/tour_feedbacks
- [x] QA
- [x] DEV

![image](https://user-images.githubusercontent.com/59872096/130212349-60b78037-fa52-472a-bc4e-6cc1629283b7.png)

6.
/admin/leads
- [x] QA
- [x] DEV
![image](https://user-images.githubusercontent.com/59872096/130212766-b0871d01-8543-4ef4-ab5c-2d9dde39cc23.png)

7.
/admin/features
- [x] QA
- [x] DEV

![image](https://user-images.githubusercontent.com/59872096/130212865-b73fc98b-59a0-4e8e-9e0e-cec1185c264f.png)

8.
due to difficulty in accessing the item, it will not be fixed, the matter is postponed
- [ ] QA
- [ ] DEV
/admin/ab_tests
![image](https://user-images.githubusercontent.com/59872096/130213028-e31a52d3-9991-43cc-957a-f283dac629a3.png)

PR #2291, opened 2021-08-30
[#2292, #2117] rwd problems 2

jarekzet:
RWD Admin panel mods according to #2117 

Fix for #2292 

commit:
[#2117 #2292] Fixed drop-down menu and RWD CSS styles


Marcelinna:
for size 978 - The menu for the mobile version has disappeared, now there is no menu at all
PR
image
Beta
image

Marcelinna (approved):

PR #2300, opened 2021-09-08
[#2299] Fix default offer's order_type bad mapping

### LINKED ISSUE #2299: The default service offer is a different type than the service
**Describe the bug**
Tested on beta Release 3.20.0-beta.1

**To Reproduce**
1. Resource has order type set to **Full Open Access** 
2. Delete offer via API - Resource is still **Full Open Access** 
3. Update resource to Order required type in PC - Resource is **Various Order type**, but has one offer
image
Resource is visible for filter Full Open Access and Order required
image
image

**Resource is order type: Order Required and one offer Full Open Access**

**Expected behavior**
Resource is Order required order type and offer is order reuqured type too

goreck888:
Issue cause by unnecessary double saving an offer in `Offer::Create` service.
(first save and then copy order_type from not updated resource)
Removed order_type's update and disabled `order_type` field with the service.order_type value set.
Fixes #2299

### DIFF
--- a/app/views/backoffice/services/offers/_form.html.haml
+++ b/app/views/backoffice/services/offers/_form.html.haml
@@ -10,7 +10,8 @@
       selected: (offer.order_type || service.order_type),
       input_html: { "data-target": "ordering.orderType",
                     "data-action": "ordering#updateVisibility",
-                    class: "form-control-lg col-lg-6" }
+                    class: "form-control-lg col-lg-6" },
+                    readonly: service.offers.published.size == 0 && service.order_type.present?
     = f.input :internal,
       wrapper_html: { "data-target": "ordering.internalWrapper" },
       input_html: { "data-target": "ordering.internal",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Reindexing and update resource `order type` by offers update or remove (@goreck888)
+- Default offer's `order type` correct mapping from a resource (@goreck888)
 
 ### Security
 
--- a/app/services/offer/create.rb
+++ b/app/services/offer/create.rb
@@ -7,9 +7,6 @@ class Offer::Create
 
   def call
     @offer.save
-    if @offer.service.offers.published.size == 1
-      @offer.update(order_type: @offer.service.order_type)
-    end
     @offer.service.reindex
     @offer
   end
--- a/app/views/backoffice/services/offers/edit.html.haml
+++ b/app/views/backoffice/services/offers/edit.html.haml
@@ -4,7 +4,7 @@
   .mb-4.pb-4.border-header
     %h1= _("Edit offer")
 
-  - if @service.offers.size == 1
+  - if @service.offers.published.size == 1
     = render "one_offer_edit_form",
       offer: @offer,
       service: @service,
--- a/app/views/services/ordering_configuration/offers/edit.html.haml
+++ b/app/views/services/ordering_configuration/offers/edit.html.haml
@@ -5,7 +5,7 @@
     %h1= _("Edit offer")
 
   - from = params[:from] || from
-  - if @service.offers.size == 1
+  - if @service.offers.published.size == 1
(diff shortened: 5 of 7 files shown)


Marcelinna:
Test cases 1

1. Create resource - Open Access - order type: Open Access
2. Update in PC - Order Required -  order type: Order Required
3. Update in PC - Other - order type: Other order type 
4. Add offer Full Open Acces - Various order types (too offers) Filters: Other, Full Open Access
5. Delete Other offer - order type: Other order type, Filters: Other
6. Update in PC - Open Access - order type: Open Access


Test case 2
1. Create resource - Order Required - order type: Order required
2. Add offer Open Access - Various order types (too offers) Filters: Order required, Open Access
3. Add to the project Open Access offer
4. Delete offer Open Access - order type: Order required, Filters: Order required
5. Update in PC - Full Open Access - order type: Full Open Access, Filters: Full Open Access

Marcelinna:
Problem in PR 
1. Create resource - Order required
2. Add new offer - Open Access
3. Order Open Access offer
4. Delete Open Access offer
5. Resource has one offer order required
image
6. Click button "Edit offer"
7. Change order type to Open Access
8. See problem Resource has one offer but it is Various order types, service and offer has different type
image

Marcelinna (approved):

jswk (changes requested):
I'd say that the lines 10-12 in app/services/offer/create.rb should be removed to fix this issue.
when Offer::Create is called the order_type should already be correctly set.
However, the order change is also okay and should work for now, until the offer logic isn't revamped.
Nevertheless, **a test for this behaviour must be added**, so that it isn't lost later on in development.

The extra changes (i.e. the added published) seem to be ok, but I would extract them to another commit, as they don't affect the main issue here.

Marcelinna:
The PR looks ok. 

It is possible to edit the offer when the resource has only one. If the resource has two offers, you just need to open both in edit mode in two separate tabs. One to delete and the other to edit. After editing such an offer, we have the problem indicated above in the comment. Resource has a different type and its only offer is different. This is an unlikely case, probably not for this PR, because this problem also applies to the possibility of removing the default offer (if it is fixed, it can be together)

app/services/service/update.rb
```diff
@@ -7,6 +7,11 @@ def initialize(service, params)
   end
 
   def call
+    if @service.errored? && @service.valid?
+      @service.update(@params.merge(status: :unverified))
+    else
+      @service.update(@params)
+    end
```
jswk, line 14: Since you removed the lines from Offer::Create, reordering this isn't needed anymore.

app/views/backoffice/services/offers/_form.html.haml
```diff
@@ -10,7 +10,8 @@
       selected: (offer.order_type || service.order_type),
       input_html: { "data-target": "ordering.orderType",
                     "data-action": "ordering#updateVisibility",
-                    class: "form-control-lg col-lg-6" }
+                    class: "form-control-lg col-lg-6" },
+                    readonly: service.offers.published.size == 0 && service.order_type.present?
```
jswk, line 14: Why such a condition?, and why apply it only to this field and not to `:internal` or some other fields? I don't understand the logic behind it, could you elaborate?

goreck888: Default offer should have the same order_type as a service. So the first condition is met if we create the first offer.  
The second condition applies to situation, when order_type in service is not set (because it's not mandatory field in the service form). 
Any other field isn't service-dependent, so that's the reason we leave it editable :)

jswk (changes requested):
Thanks for the clarification, this logic seems correct indeed.

Only thing left is that you reverted the changes from `update.rb` in the second commit, but introducing it in the first one. Just remove this reorder from the first commit, then reverting it in the second one won't be necessary.

Otherwise it's good to go.

jswk (approved):

jswk (commented):
Please add a changelog entry too in a the first, fixing, commit.

jswk (approved):

kmarszalek (approved):

PR #2303, opened 2021-09-13
Add e2e test for fileters and search

Marcelinna:


jswk (commented):
The test itself looks good.

I would only ask to correct the typo in the commit message, i.e. `fileters` (should be `filters`).

jswk (approved):

PR #2304, opened 2021-09-13
[#2297] Add button to clear query in the BO

### LINKED ISSUE #2297: no button to clear the text in the searchbar (Backoffice )
**Describe the bug**
Tested beta 3.20.0-beta.0
there is no X button in the backoffice to clear typed text

MP 
image

BO
image

goreck888:
Add button to query clearing
in the backoffice services
Fixes #2297


Marcelinna (approved):

app/views/backoffice/services/_search.html.haml
```diff
@@ -12,6 +12,10 @@
                                      placeholder: _("Find resource..."),
                                      "data-target": "autocomplete.input",
                                      "data-toggle": "dropdown"
+    - if params[:q].present? && params.permitted?
+      %a.search-clear{ href: url_for(params.except(:q, :sort, :service_id)), title: _("Clear search"),
```
jswk, line 16: Wouldn't we want to clear paging params too?
On PR, if I go to https://marketplace-5.docker-fid.grid.cyf-kr.edu.pl/backoffice/services?q=hoo&per_page=20&page=2&sort=-sort_name and click the X, then paging params remain. I would add `:per_page` and `:page` too.
On the other hand, `:service_id` shouldn't be cleared - if it's present then request is redirected to resource anyway, not displaying the clearing control anyway.

I would make this change in non-backoffice search input too, actually.

jswk (changes requested):
I would improve working of the clearing X as described in comment before merging this.

app/views/backoffice/services/_search.html.haml
Marcelinna: We can add, although this is not the scope of the pinned bug. We would have to change the behavior in the backoffice and in the main catalogue.

jswk: > We can add, although this is not the scope of the pinned bug. We would have to change the behavior in the backoffice and in the main catalogue.

Ok, so I wouldn't propagate it to the general clearing X, but still, it doesn't make sense for the code added in this PR to have `:service_id`. At least it should be removed.

goreck888: I agree with the `:page` parameter, but why `:per_page`? I think it's independent parameter and if it's set, it shouldn't be removed :)

jswk: I bundled `:per_page` up with `:page` and didn't give it enough thought. You're right that `:per_page` shouldn't be removed.

jswk (commented):
From https://github.com/cyfronet-fid/marketplace/pull/2304#discussion_r709085449:
> I agree with the :page parameter, but why :per_page? I think it's independent parameter and if it's set, it shouldn't be removed :)

Have something changed your mind in between changes? You removed the `:service_id` from cleared parameters which is good, but the `:page` param hasn't been added.

**PG:** Done.

jswk (approved):

kmarszalek (approved):

kmarszalek:
Add button to clear query in the Backoffice title vs Add button to query clear in the Backoffice changelog entry

jswk (changes requested):
Since 3.21.0-beta.0 was already released, please move the changelog entry to 3.22.0-milestone section.

CHANGELOG.md
```diff
@@ -6,9 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Please view this file on the master branch, on stable branches it's out of date.
 
-## [3.21.0-milestone]
+## [3.22.0-milestone]
```
jswk, line 9: Please don't remove the section for 3.21.0, even though it's empty.
Add the changelog entry in a new section.

jswk (changes requested):

PR #2307, opened 2021-09-20
[#2092] Fix anchor tag persistence on denied permission

### LINKED ISSUE #2092: anchor (#) does not disappear from the url after leaving the page
Tested 3.12.0-beta.2
**Describe the bug**
after revoking the rights from the page with the anchor, it does not disappear from the url where the user was redirected

**To Reproduce**
1. Select resource for which you have admin privileges and see "Manage the resource" button
2. Click on "Manage the resource" and select "Set parameters and offers" - url is https://beta.marketplace.eosc-portal.eu/services/{service.name}/ordering_configuration#offers
3. Delete admin privieges in PC or BO
4. Refresh page https://beta.marketplace.eosc-portal.eu/services/{service.name}/ordering_configuration#offers - user is redirect to https://beta.marketplace.eosc-portal.eu/#offers - anchor doesnt removed from url 

**Expected behavior**
Anchor is removed from url

**Screenshots**

**Additional context**

wujuu:
Bug happens due to the behaviour of 302 HTTP Redirect, which persists the anchor tag (#offers in our case) in the url after the redirection (https://stackoverflow.com/questions/5283395/url-hash-is-persisting-between-redirects). Due to losing permission to see the page, the user is being redirected by `pundit` (to the previous page if possible ("referer" header value), or to the root_url if not) - and the anchor persists, causing the bug.

My solution includes replacing the `redirect_back` with `redirect_to root_page(anchor: "")`. This is acceptable, because in MP we have a policy that the users cannot click the links or buttons that they aren't authorized to see (@marcelinna - please check this), thus making the `redirect_back` redundant - the only way to access unauthorized resources is to explicitly input the url, in which case the "referer" header is empty, in which case we are redirected to the root_page anyway (`fallback_url` in `redirect_back`).

Closes #2092

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Include resources without offers in statistics in the executive panel (@goreck888)
+- Fix anchor tag url persistence in redirect after trying to access a forbidden resource (@wujuu)
 
 ### Security
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -21,8 +21,8 @@ class ApplicationController < ActionController::Base
   end
 
   rescue_from Pundit::NotAuthorizedError do |exception|
-    redirect_back fallback_location: root_path,
-                  alert: not_authorized_message(exception)
+    redirect_to root_path(anchor: ""),
+                alert: not_authorized_message(exception)
   end
 
   def tour_disabled
--- a/spec/features/access_denied_spec.rb
+++ b/spec/features/access_denied_spec.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Access denied" do
+  include OmniauthHelper
+
+  context "After trying to enter an unauthorized resource" do
+    let!(:user) { create(:user) }
+
+    before { checkin_sign_in_as(user) }
+
+    scenario "I am being redirected to root_path" do
+      visit backoffice_path
+      expect(page).to have_current_path(root_path(anchor: ""))
+    end
+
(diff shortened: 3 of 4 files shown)


jswk (changes requested):
This is a subtle behaviour that you correct and it would make sense to test it explicitly to avoid regression.

Also, changelog entry is missing.

As you made some research to solving this issue, as described in the PR, I would include it in the commit message too, so that it's more readily available when reading the repo history.

jswk (approved):
Looks great!

Marcelinna:
@wujuu 
This is acceptable, because in MP we have a policy that the users cannot click the links or buttons that they aren't authorized to see (@Marcelinna - please check this)


Steps to reproduce
1. Go to resource which you have admin privileges and see "Manage the resource" button
image
2. Click on Manage Resources and open in new tab Set parametres and offers
3. In new tab open Edit parametres and Add new offer
image
4. Delete Admin privilage
5. As a user I can click on every button which I don't have admin privileges, becaus I lost privilage during the session.


I can click on page https://marketplace-11.docker-fid.grid.cyf-kr.edu.pl/services/resource-template-3-20-0-beta-2/ordering_configuration#offers on button Edit offer, create new offer, although I no longer have the rights.
image
When I click on Edit offer or create new offer, the form for editing, creating a new offer will open without any problems

image
image


Only pressing the "Update offer" or "Create new offer" or "Cancel and back to previous page " button results in displaying information about the lack of permissions



so I am not sure if we meet the requirement "This is acceptable, because in MP we have a policy that the users cannot click the links or buttons that they aren't authorized to see "

michal-szostak (approved):

kmarszalek:
This solution is good enough for an anchor in MP and can be merged.
MP Policy should be responsible for auth requests.

PR #2309, opened 2021-09-20
[#2139] Fix project message caption

### LINKED ISSUE #1359: Enhance the difference between User and Service Provider in the conversation tab view
Now it is hard to distinguish message from Service provider and User while having a discussion in the conversation Tab. 

Have a clear difference between messages 

Ideas to implement the change : 

* to make different colours 
* to use the real user name as it is done in JIRA 
* make the word " User"  and " Service Provider" bold 

**Additional context**

### LINKED ISSUE #2139: wrong information after sending a message to Project Support
Tested 3.13.0-beta.1

**Describe the bug**
after sending message to "Project Support", information is displayed "Your message to the resource provider"

**To Reproduce**
1. Go to '/projects'
2. Create project
3. Send message to Project Support

**Expected behavior**
information should be "Your message to the project support"

**Screenshots**
image

**Additional context**

wujuu:
Closes #2139. Potentially closes #1359 (to check by @michal-szostak)

- New conversations view [@jarekzet, @wujuu]
- Improved conversation labels (which now include: date, role and optionally email and name (not relevant in the context of user messages)) [@wujuu]

How to test:
1. Create a project
2. Navigate to `/projects/<project_id>/conversation`
2. Fill in and send the message form in the frontend to see the **user message**
3. Execute following lines in `rails c`:
```
> Message.create(message: "message from provider", scope: "public", author_role: "provider", author_name: "Jan Kowalski", author_email: "qwe@google.com", messageable: Project.last)
> Message.create(message: "message from customer support", scope: "public", author_role: "mediator", author_name: "Pan Kowalski", author_email: "qwe@google.com", messageable: Project.last)
> Message.create(message: "TOP SECRET message from provider", scope: "user_direct", author_role: "provider", author_name: "Pani Kowalska", author_email: "qwe@google.com", messageable: Project.last)
> Message.create(message: "message from provider", scope: "public", author_role: "provider", author_name: "Jan Kowalski", author_email: nil, messageable: Project.last)
> Message.create(message: "message from provider", scope: "public", author_role: "provider", author_name: nil, author_email: "qwe@google.com", messageable: Project.last)
> Message.create(message: "message from provider", scope: "public", author_role: "provider", author_name: nil, author_email: nil,  messageable: Project.last)
```
... to seed the conversation with:
- message from the **provider**, who is identified by **name and email**
- message from the **mediator** (customer support), who is identified by **name and email**
- message from the **provider**, who should have "VISIBLE ONLY TO YOU" label (provider identified by name and email
- message from the **provider**, who is identified by **email only**
- message from the **provider**, who is identified by **name only**
- message from the **provider**, who is **anymous** (didn't provide any additional information) 
4. _NOTE how the labels change!_
5. (optional) Mix and match `message`, `author_name`, `author_email` with different values if needed (to check how the UI behaves for longer names for example)
6. (optional) Fill in `author_role` with `provider` or `mediator` values to check for unspecified scenarios if needed (to check how the label behaves on different message lengths for example)
7. Project Item messages should look the same. To seed it in the same way, add a project item to your project and execute the above commands with the exception that the `messageable` field should be `ProjectItem.last`. Then navigate to `/projects/<project_id>/services/<project_item_id>/conversation` to see the seeded conversations.

Conversations should look like in the preview: https://invis.io/JXOZKC8WMDU#/460826937_Projects_-_Contact_Experts_-_Chat


config/locales/en.yml
```diff
@@ -47,8 +47,12 @@ en:
...
   conversations:
     message:
-      question: "Your message to the resource provider"
-      answer: "Message from the resource provider"
+      project_item:
+          question: "Your message to the resource provider"
+          answer: "Message from the resource provider"
+      project:
+          question: "Your message to the project support"
+          answer: "Message from the project support"
```
jswk, line 55: The scope of the ticket was only to change the title for the user messages. If you want to handle provider/ops-team messages then you should differentiate based on `message.author_role`.
For example, if a message from provider is shown in the context of a project, then the text "Message from the project support" will be incorrect, as the message will be from a resource provider. The inverse will happen if you are in the context of a project_item.

So, either stick to the ticket or extend the ticket to handle other author_roles correctly.

jswk (changes requested):

app/helpers/project_items_helper.rb
```diff
@@ -2,10 +2,30 @@
 
 module ProjectItemsHelper
   def label_message(message)
-    if message.question?
-      t("conversations.message.question")
-    else
-      t("conversations.message.answer")
+    date = message.created_at.to_formatted_s(:short)
```
jswk, line 5: Format the date with a more complete format. I would use `:db`, as it creates a nice "2021-11-15 11:25:06", following the prevs more closely.

app/helpers/project_items_helper.rb
```diff
@@ -2,10 +2,30 @@
...
+    "#{date}#{author_identity(message)}, #{author_title(message)}"
```
jswk, line 6: The ordering should actually depend on whether it's a provider/mediator message or a user message. In the latter case the order should be reversed.
The prev:
image
And in this branch:
image

app/views/messages/_message.html.haml
```diff
@@ -1,12 +1,6 @@
...
-    %p.frame{ class: ("edited" if message.edited?) }= message.message
-    .d-block.message-info= label_message(message)
-
-
-
-
-
+%li.step-progress-item.ml-3{ class: (message.role_user? ? "align-self-end" : "align-self-start") }
+  %p.frame{ class: ("edited" unless message.role_user?) }= message.message
+  -# TODO: Change the class name from "edited" to something else ("message_to_user" perhaps?)
```
jswk, line 3: Change the class name, not leaving behind a TODO. In the styles it only defines a different background anyway.
When changing it's name follow up to rename `$edited-comment-background-color` and `$edited-comment-border-color` as they won't describe edited styles anymore.

app/helpers/project_items_helper.rb
wujuu: I thought this was a mistake on the prev, will fix

app/helpers/project_items_helper.rb
wujuu: How about ""November 15, 2021 13:24", isn't this more readable and user-friendly?

jswk: I would stick to the `:db` version for now, it's more standard. If there arise issues with `:db` we could migrate to `:long` or something more custom (however, I would leave the `:db` rendering in `alt` in that case).

jswk (approved):
Perfect, thanks!
Just remember to squash before merging.

Marcelinna:
image
we should indicate what time we use

Marcelinna:
image
on previews I see pink underline with word "NEW". It has not been implemented in this PR?

Marcelinna:
image
image

the styles should be the same for both views

wujuu:
> image **QA comment:** on previews I see pink underline with word "NEW". It has not been implemented in this PR? **DEV comment:**

You're right, but this is a feature that will be implemented in another PR: #2406

Marcelinna:
image
email is write in brackets, when email isn't entered, bracet is empty. Maybe we should get rid of the bracket?

wujuu:
> image email is write in brackets, when email isn't entered, bracet is empty. Maybe we should get rid of the bracket?

This is dues to ability to enter an empty string ("") instead of nil in the `email` field in the Message model. This should not be possible. Will add the ticket to add proper validation to this field (nil or email regex)

wujuu:
> image we should indicate what time we use
> 
> QA comment: It will be continue in other PR

This is a scope for another issue: https://github.com/cyfronet-fid/marketplace/issues/2410

Marcelinna (approved):

PR #2326, opened 2021-10-04
Bundles, WIP but not draft

kmarszalek:


Marcelinna:
EGI Notebook - Notebooks for researchers is ordered via MARKETPLACE
B2DROP - For Researchers is an open-access order  type

bundle is via Marketplace is that how it was designed?
image

Marcelinna:
beta
image
preview
image

Marcelinna:
After add bundle I see "You have no resources in this project yet. "  ??
image

Marcelinna:
I don't know if any task was released, but every service has an offer at the beta instance

Production
image
Beta (offer description is a copy of the resource description)
image

PR #2339, opened 2021-10-12
Bundles part 2

jswk:
Since sadly a PR cannot be reopened after merge, I'm creating this one for ongoing corrections to be incorporated into 3.22.x before forward merges are done, as in current state we cannot do a merge to a branch with bumped ruby/rails.
- [x] Jira
- [x] tests
- [x] deleting all not bundled offers https://github.com/cyfronet-fid/marketplace/pull/2326#issuecomment-940746019
- [x] add happy-path bundle ordering feature test

The mentioned PR with history: #2326.

The bundle config validation has been extracted to #2342.

### DIFF
--- a/app/models/jira/client.rb
+++ b/app/models/jira/client.rb
@@ -206,9 +206,7 @@ private
       "category" => project_item.service.categories.first&.name,
       "service" => project_item.service.name,
       "offer" => project_item.name,
-      "attributes" => encode_properties(project_item.properties),
-      # TODO handle bundles
-      "bundled_services" => encode_properties(project_item.properties)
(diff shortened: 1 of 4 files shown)


Marcelinna:
When I order no bundle offer in configuration step during making the mistake I see an alert message "Please correct errors presented below"
image

When I order bundle offer and dont fiil field for second offer, in summary step after click on the button "Add to the project" I see alert "Service request configuration is invalid"
image

Expected result: we don't need to introduce a new alert message, we've simplified it recently. Change alert message to "Please correct errors presented below".

PR #2379, opened 2021-11-02
Bundle configuration form validation

### LINKED ISSUE #2342: Bundle config validation
Continuation of #2339 and #2326.

The scope is to make validation in bundles work correctly.

EDIT:

- [x] fix validation msg also.
https://github.com/cyfronet-fid/marketplace/pull/2339#issuecomment-943157747
- [x] fix order bundle offer without parameters
https://github.com/cyfronet-fid/marketplace/pull/2339#issuecomment-943383270
- [ ] fix grapfic problem
https://github.com/cyfronet-fid/marketplace/pull/2339#issuecomment-943435050

goreck888:
Add a validation of bundled parameters
in the configuration step of the service ordering form

Fixes #2342

### DIFF
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -59,6 +59,7 @@ class Offer < ApplicationRecord
            through: :target_offer_links,
            source: :target
 
+
   validate :set_iid, on: :create
   validates :service, presence: true
   validates :iid, presence: true, numericality: true
@@ -96,6 +97,10 @@ class Offer < ApplicationRecord
 
   def bundle?
     bundled_offers_count.positive?
+  end
+
+  def bundle_parameters?
+    bundle? && (parameters.present? || bundled_offers.map(&:parameters).any?(&:present?))
   end
 
   private
--- a/app/models/project_item/customization.rb
+++ b/app/models/project_item/customization.rb
@@ -5,7 +5,7 @@ module ProjectItem::Customization
 
   included do
     validate do
-      if self.offer.bundle?
+      if offer&.bundle?
         bundled_property_values.each do |offer, parameters|
           parameters
             .select { |pv| pv.invalid? }
@@ -29,11 +29,16 @@ module ProjectItem::Customization
   end
 
   def bundled_property_values
-    self.bundled_parameters.present? ?
-      self.bundled_parameters
-          .transform_keys { |offer| offer.instance_of?(Offer) ? offer : Offer.find(offer) } : offer_values
-                                                                                          .attributes_map
-                                                                                          .reject { |o, _| o == offer }
+    if defined?(self.bundled_parameters)
+      if self&.bundled_parameters.present?
+        self.bundled_parameters
+          .transform_keys { |offer| offer.instance_of?(Offer) ? offer : Offer.find(offer) }
+      else
+        offer_values.attributes_map.reject { |o, _| o == offer }
+      end
+    else
+      []
+    end
   end
 
   def bundled_property_values=(bundled_property_values)
--- a/app/models/project_item/customization.rb
+++ b/app/models/project_item/customization.rb
@@ -5,13 +5,13 @@ module ProjectItem::Customization
 
   included do
     validate do
-      # if offer.bundle?
-      #   bundled_property_values.each do |offer, parameters|
-      #     parameters
-      #       .select { |pv| pv.invalid? }
-      #       .each { |pv| errors.add(:property_values, :invalid, value: pv) }
-      #   end
-      # end
+      if self.offer.bundle?
+        bundled_property_values.each do |offer, parameters|
+          parameters
+            .select { |pv| pv.invalid? }
+            .each { |pv| errors.add(:bundled_property_values, :invalid, value: pv) }
+        end
+      end
       property_values
         .select { |pv| pv.invalid? }
         .each { |pv| errors.add(:property_values, :invalid, value: pv) }
@@ -29,7 +29,11 @@ module ProjectItem::Customization
   end
 
   def bundled_property_values
-    offer_values.attributes_map.reject { |o, _| o == offer }
+    self.bundled_parameters.present? ?
+      self.bundled_parameters
+          .transform_keys { |offer| offer.instance_of?(Offer) ? offer : Offer.find(offer) } : offer_values
+                                                                                          .attributes_map
+                                                                                          .reject { |o, _| o == offer }
   end
 
   def bundled_property_values=(bundled_property_values)
@@ -37,7 +41,7 @@ module ProjectItem::Customization
       offer = id_to_bundled_offer[offer_id]
       offer_values.update(offer.id => property_values) if offer
     end
-    self.properties
+    self.bundled_parameters = offer_values.to_json
   end
 
   private
--- a/app/models/project_item/project_validation.rb
+++ b/app/models/project_item/project_validation.rb
@@ -5,7 +5,7 @@ module ProjectItem::ProjectValidation
 
   included do
     validates :project, presence: true
-    validate :one_per_project, unless: :properties?
+    validate :one_per_project, unless: :properties_or_bundled_offer?
   end
 
   def one_per_project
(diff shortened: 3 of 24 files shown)


app/models/offer.rb
```diff
@@ -98,6 +99,14 @@ def bundle?
     bundled_offers_count.positive?
   end
 
+  def bundle_has_parameters?
+    bundle? && (parameters.present? || bundled_offers.map(&:parameters).present?)
```
jswk, line 103: I understand, that in `bundled_offers.map(&:parameters).present?` you wanted to catch a case, when any of the bundled offers has any params. However, this part will always return true, given any bundled offer, see below for why.

The `bundled_offers.map(&:parameters)` will return an array of parameters, e.g. `[{...}, {...}]`. If parameters are absent, then it will look like `[nil]`. If you call `[nil].present?`, you still get true.

What you wanted was rather `bundled_offers.map(&:parameters).any?(&:present?)`. It returns true for `[nil, {...}]` and false for `[nil, [], {}]`, as expected.

app/models/offer.rb
```diff
@@ -98,6 +99,14 @@ def bundle?
...
+  end
+
+  def parameters?
+    parameters.present?
+  end
```
jswk, line 108: Do you use this method anywhere? It seems to me that it's unused and could be removed.

app/models/project_item/customization.rb
```diff
@@ -5,13 +5,13 @@ module ProjectItem::Customization
...
   included do
     validate do
-      # if offer.bundle?
-      #   bundled_property_values.each do |offer, parameters|
-      #     parameters
-      #       .select { |pv| pv.invalid? }
-      #       .each { |pv| errors.add(:property_values, :invalid, value: pv) }
-      #   end
-      # end
+      if offer.present? && offer.bundle?
```
jswk, line 8: Wouldn't `if offer&.bundle?` be better?

app/models/project_item/customization.rb
```diff
@@ -29,15 +29,23 @@ def property_values=(property_values)
   end
 
   def bundled_property_values
-    offer_values.attributes_map.reject { |o, _| o == offer }
+    if defined?(self.bundled_parameters)
+      self&.bundled_parameters.present? ?
+        self.bundled_parameters
+          .transform_keys { |offer| offer.instance_of?(Offer) ? offer : Offer.find(offer) } : offer_values
+                                                                                          .attributes_map
+                                                                                          .reject { |o, _| o == offer }
```
jswk, line 37: I would be more explicit here and transform it to an expanded `if`, i.e.
```ruby
if self.bundled_parameters.present?
  self.bundled_parameters.transform_keys { |offer| offer.instance_of?(Offer) ? offer : Offer.find(offer) }
else
  offer_values.attributes_map.reject { |o, _| o == offer }
end
```

Current formatting is really difficult to follow.

app/models/project_item/project_validation.rb
```diff
@@ -16,4 +16,8 @@ def one_per_project
 
     errors.add(:project, :repeated_in_project) unless !project_items_services.present?
   end
+
+  def properties_or_bundled_offer?
+    properties? || offer.present? && (offer.bundle? || offer.source_offer_links.count.positive?)
```
jswk, line 21: Add parens to make logic more explicit.
```ruby
    properties? || (offer.present? && (offer.bundle? || offer.source_offer_links.count.positive?))
```
At least that is what I think you had on your mind.

Otherwise you rely on priorities between `||` and `&&`, which are non-trivial to follow at times.

spec/factories/customizable_project_items.rb
```diff
@@ -0,0 +1,19 @@
...
+    request_voucher { false }
+
+    offer
+    project
+
+    factory :customizable_project_item_with_voucher do
+      voucher_id { "some_voucher_id" }
+
+      offer { build(:voucherable_offer) }
+    end
```
jswk, line 17: This seems to be unused, good for pruning.

jswk (changes requested):
Also, the test: `ProjectItem::Create#bundle with error creates nothing` fails in the CI.
It's most likely because I depended on project being invalid when there are multiple offers from the same service. It's no longer enforced for bundled offers, so no error is generated.
You should make the project in that test invalid again, so that the test remains functional.

Marcelinna:
Steps to reproduce
1. Order bundle offer where no parameters
image
2. Click button "Next final details"
3. See error - page is blank

Marcelinna:
I can order bundle
image

I can't order bundle
image
image

Marcelinna:
changelog?

app/models/offer.rb
goreck888: Ok

spec/factories/customizable_project_items.rb
goreck888: Ok

app/models/project_item/project_validation.rb
goreck888: Done

Marcelinna:
the current ordering logic. 
1. User can add to the project only once an offer that has no parameters
2. User can add an offer with parameters to the project many times
3. A bundle offer can be added many times to a project, whether it has parameters or not

4. An offer that belongs to the bundle can be added to the project many times, whether it has parameters or not (
    B2DROP -For Premium)
5. When in project user has a Bundle offer (EGI Notebook+B2DROP) adding second offer EGI Notebook without parameters isn't possible.
6. When in project user has a Bundle offer (EGI Notebook+B2DROP) adding second offer B2DROP without parameters isn't possible.

app/views/projects/services/_project_item.html.haml
```diff
@@ -14,7 +14,8 @@
       -# TODO: refactor dynamic translation
       = t("project_items.status.#{project_item.status_type}")
     - elsif !project_item.orderable?
-      = link_to _("Visit website"), webpage(project_item)
+      %span
+        = link_to _("Visit website"), webpage(project_item)
```
jswk, line 18: Wouldn't you need to enclose the other branches of this if in `<span>` too?

jswk (changes requested):
Thanks for resolving most problems. There's two things to consider though introduced in the meantime. Have a look at these.

app/models/offer.rb
```diff
@@ -98,6 +99,10 @@ def bundle?
     bundled_offers_count.positive?
   end
 
+  def bundle_has_parameters?
```
kmarszalek, line 102: bundle_parameters?

kmarszalek (changes requested):
almost good

app/services/project_item/create.rb
```diff
@@ -19,15 +19,17 @@ def call
 
       if @project_item&.offer.bundle?
         bundled_project_items = @project_item.offer.bundled_offers.map do |offer|
+          bundled_parameters = @bundle_params.present? && @bundle_params&.has_key?(offer.id) ?
```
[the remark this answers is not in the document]
goreck888: Unfortunately it doesn't work. If `@bundle_params` is an empty Array it calls method `has_key?` and returns `NoMethodError`. Safe navigator `&.` protects only from calling a method on `nil`. So the `@bundled_params.present?` is necessary to return `false` in case of empty array and don't call `has_key?` method.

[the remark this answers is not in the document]
jswk: Right, then you can still get errors here, if the array has any element.
I'd suggest then to check if `@bundle_params` responds to `:has_key?`. Then you are safer than currently.

The condition would be then: `@bundle_params&.respond_to?(:has_key?) && @bundle_params.has_key?(offer.id)`.
You would only need the safe `&.` in the first part.

app/views/projects/services/_project_item.html.haml
```diff
@@ -14,7 +14,8 @@
       -# TODO: refactor dynamic translation
       = t("project_items.status.#{project_item.status_type}")
     - elsif !project_item.orderable?
-      = link_to _("Visit website"), webpage(project_item)
+      %span
```
kmarszalek, line 17: still here?

goreck888: No :)

jswk (approved):

kmarszalek (commented):
wait for Marecelina

kmarszalek (approved):
wait for Marecelina

Marcelinna (approved):

PR #2383, opened 2021-11-05
Test e2e for provider in backoffice

Marcelinna:
Add test for adding, editing and deleting provider without resources

Handle exception Failed to open TCP connection from Faraday

### DIFF
--- a/app/helpers/url_helper.rb
+++ b/app/helpers/url_helper.rb
@@ -25,7 +25,7 @@ module UrlHelper
 
     response = Faraday.get(url)
     response.status == 200
-  rescue URI::InvalidURIError, NoMethodError
+  rescue URI::InvalidURIError, NoMethodError, Faraday::Error
     false
   end
 end
--- a/app/views/backoffice/providers/form/_location.html.haml
+++ b/app/views/backoffice/providers/form/_location.html.haml
@@ -3,7 +3,7 @@
     [:street_name_and_number, :postal_code, :city, :region, :country])),
               data: { toggle: "collapse", target: "#location" },
               aria: { expanded: true, controls: "location" } }
-    .card-header.text-left{ id: "marketing-header" }
+    .card-header.text-left{ id: "location-header" }
       .row
         .col-10
           = _("Location")
--- a/test/system/cypress/factories/provider.factory.ts
+++ b/test/system/cypress/factories/provider.factory.ts
@@ -1,0 +1,43 @@
+import {Utilities} from "../support/utilities";
+
+export class IProviders {
+    basicName: string;
+    basicAbbreviation: string;
+    basicWebpage_url: string;
+    marketingDescription: string;
+    marketingMultimedia: string;
+    classificationScientificDomains: string[];
+    classificationTag: string;
+    locationStreet:string;
+    locationPostCode:string;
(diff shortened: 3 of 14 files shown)


app/helpers/url_helper.rb
```diff
@@ -25,7 +25,7 @@ def self.url_valid?(url)
 
     response = Faraday.get(url)
     response.status == 200
-  rescue URI::InvalidURIError, NoMethodError
+  rescue URI::InvalidURIError, NoMethodError, Faraday::ConnectionFailed
```
jswk, line 28: I would actually go a step further and rescue from `Faraday::Error`. Hava look at https://github.com/lostisland/faraday/blob/main/lib/faraday/error.rb, there is for example SSLError, ParsingError, which should also be handled by returning false in that case.
Good catch though!

jswk (commented):
I would introduce a single improvement, otherwise it looks good.

app/views/backoffice/providers/form/_location.html.haml
```diff
@@ -3,7 +3,7 @@
     [:street_name_and_number, :postal_code, :city, :region, :country])),
               data: { toggle: "collapse", target: "#location" },
               aria: { expanded: true, controls: "location" } }
-    .card-header.text-left{ id: "marketing-header" }
```
danielkryska, line 6: Have you checked if the id isn't used anywhere?

test/system/cypress/factories/provider.factory.ts
```diff
@@ -0,0 +1,43 @@
...
+    contactEmail:string;
+    publicContactsEmail:string;
+    adminFirstName?:string;
+    adminLastName?:string;
+    adminEmail?:string;
+}
+
+export const ProvidersFactory = {
+    create: (args: {[field: string]: string} = {}): IProviders => ({
+        basicName: Utilities.getUUID4(),
```
danielkryska, line 26: Good practice would be to remove Utilities related to randomly generated data and use the [faker JS lib](https://github.com/marak/Faker.js/).

test/system/cypress/support/providers.ts
```diff
@@ -0,0 +1,165 @@
+/**
+ * Define new commands types for typescript (for autocompletion)
+ */
+ import { IProviders } from "../factories/provider.factory";
+
+ declare global {
+   namespace Cypress {
+     interface Chainable {
+      
+       fillFormCreateProvider(provider: IProviders, logo:any): Cypress.Chainable<void>;
```
danielkryska, line 10: 1. provider: Partial<IProvider>
2. Name method as fillForm and use it for creating and updating

danielkryska (approved):

test/system/cypress/support/providers.ts
Marcelinna: I can't change every method for fillForm(), TS does not match them properly, for every test want to use the method for the provider, despite I added correctly another object.

app/views/backoffice/providers/form/_location.html.haml
Marcelinna: yes, it can be changed to correct name

jswk (approved):

kmarszalek (approved):

PR #2384, opened 2021-11-08
[#2372] Handle provider deletion in JMS

### LINKED ISSUE #2371: Consult Profile Service author endpoint with PCSS
Data format and what type of data should be decided together with PCSS and later implemented in Profile Service.

### LINKED ISSUE #2372: Handle provider deletion in JMS
Using new provider's soft delete feature (#2354) handle provider deleted message in JMS.

Additionally changes in provider status (other than delete) should set provider's status to default status (the same as with resources)

goreck888:
Add possibility to delete a provider via JMS
Closes #2372

### DIFF
--- a/app/jobs/provider/delete_job.rb
+++ b/app/jobs/provider/delete_job.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class Provider::DeleteJob < ApplicationJob
+  queue_as :pc_subscriber
+
+  def perform(provider_id)
+    Provider::Delete.new(provider_id).call
+  end
+end
--- a/app/services/importers/provider.rb
+++ b/app/services/importers/provider.rb
@@ -96,7 +96,8 @@ class Importers::Provider
       societal_grand_challenges: map_societal_grand_challenges(societal_grand_challenges),
       national_roadmaps: national_roadmaps,
       data_administrators: data_administrators,
-      synchronized_at: @synchronized_at
+      synchronized_at: @synchronized_at,
+      status: :published
     }
   end
 end
--- a/app/services/jms/manage_message.rb
+++ b/app/services/jms/manage_message.rb
@@ -38,7 +38,9 @@ class Jms::ManageMessage
       end
     when "provider"
       modified_at = modified_at(resource, "providerBundle")
-      if resource["providerBundle"]["active"]
+      if action == "delete"
+        Provider::DeleteJob.perform_later(resource["providerBundle"]["provider"]["id"])
+      elsif resource["providerBundle"]["active"]
         Provider::PcCreateOrUpdateJob.perform_later(resource["providerBundle"]["provider"], modified_at)
       end
     else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Added
 - Add the tags to a provider page (@danielkryska)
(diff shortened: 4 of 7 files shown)


Marcelinna:
image

app/jobs/provider/delete_job.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Provider::DeleteJob < ApplicationJob
+  queue_as :pc_subscriber
+
+  def perform(provider_id)
+    provider = Provider.find_by(pid: provider_id)
+    Provider::Destroy.new(provider).call
```
kmarszalek, line 8: since Provider::Destroy doesn't call destroy, maybe it should be renamed to Provider::Delete and then do like Service. 
```
Provider::Delete.new(service_id).call
```

kmarszalek (changes requested):
also i would include changelog :)

kmarszalek:
is ES aware of deleting the provider via JMS?

app/jobs/provider/delete_job.rb
jswk: It's probably called in `app/services/provider/destroy.rb`, but verify it's actually the case.

app/services/jms/manage_message.rb
```diff
@@ -38,8 +38,12 @@ def call
...
       modified_at = modified_at(resource, "providerBundle")
-      if resource["providerBundle"]["active"]
-        Provider::PcCreateOrUpdateJob.perform_later(resource["providerBundle"]["provider"], modified_at)
+      if action == "delete"
+        Provider::DeleteJob.perform_later(resource["providerBundle"]["provider"]["id"])
+      else
+        if resource["providerBundle"]["active"]
+          Provider::PcCreateOrUpdateJob.perform_later(resource["providerBundle"]["provider"], modified_at)
+        end
       end
```
jswk, line 47: Unnecessarily nested if, instead do:
```ruby
elsif resource["providerBundle"]["active"]
  Provider::PcCreateOrUpdateJob.perform_later(resource["providerBundle"]["provider"], modified_at)
end
```

jswk:
> is ES aware of deleting the provider via JMS?

It's likely done in `app/services/provider/destroy.rb`, but verify it's actually the case.

Marcelinna:
changelog?

app/services/jms/manage_message.rb
goreck888: Done

app/jobs/provider/delete_job.rb
goreck888: Done

Marcelinna:
- [ ] QA
- [ ] DEV

Steps to reproduce
1. Go to PC create a new provider with resource - resource is visible in MP
2. Delete Provider in PC
3. See error - Provider isn't deleted, resource isn't go to deleted status

app/services/importers/provider.rb
```diff
@@ -96,7 +96,8 @@ def call
       societal_grand_challenges: map_societal_grand_challenges(societal_grand_challenges),
       national_roadmaps: national_roadmaps,
       data_administrators: data_administrators,
-      synchronized_at: @synchronized_at
+      synchronized_at: @synchronized_at,
+      status: "published"
```
jswk, line 100: Better use enum in that case, i.e. `status: :published`.

Marcelinna (approved):

jswk (approved):

kmarszalek (approved):

PR #2406, opened 2021-11-15
[#2394] New message notification label

### LINKED ISSUE #2394: New message notification in project / resource view
Messages should be marked as `viewed` and if there is a message which hasn't been viewed in the project / resource the notification text should be shown.

The scope of this issue:
1. New messages are by default `viewed = false`
2. When user navigates to the conversation view all messages in the view are marked `viewed = true`
3. If there are messages `viewed == false` show notification text in the project view according to the previews

**Preview**

See places where information about new messages is displayed

https://invis.io/JXOZKC8WMDU#/460826934_Projects_-_Resources_List

https://invis.io/JXOZKC8WMDU#/460827061_Projects_-_Details

wujuu:
Closes #2394

Additionally:
- Implement dynamic new message separator
- Add Messageable concern
- Add MessageBoxComponent to conversation views
- Some refactoring work

**For Devs**

I decided to go with labeling Projects and Project_items (Messageables) with a timestamp (`conversation_last_seen`) rather than adding a `viewed` field to the Message model (as @jswk suggested). 

This approach is much less error-prone, due to the conflicts that could occur when interacting with messages through the Ordering API.

Additionally, the action of "viewing" the messages is more of a User's trait (and through DB relations, Messageable's) rather than Message's, especially since Messages could be technically "viewed" by OMS support teams (through the Ordering API). If we decided to go with the initial approach, then the Messages seen be the OMS team would have to be NOT labled as "viewed" for the front-end to work, which could be potentially confusing.

**For QA**

Previews:
- https://projects.invisionapp.com/share/JXOZKC8WMDU#/screens/460827061
- https://projects.invisionapp.com/share/JXOZKC8WMDU#/screens/460826934
- https://projects.invisionapp.com/share/JXOZKC8WMDU#/screens/460826937

How to test:
1. Setup new messages for project and project_item as in https://github.com/cyfronet-fid/marketplace/pull/2309
2. Add new project or project_items if needed
3. New notifications should appear (as in previews):
3.1 Next to project name if there are any new messages 
3.2 Next to "Contact with resource provider" or "Contact with project support" nav items
3.3 Next to a project item name ("You have a new message" + icon)
3.4 In the conversation view (a line with a "New" label, above the first new message)
4. There should be a "NEW" message separator (along with a red line, as in the preview) in both conversation views, separating old messages for the new ones. This should NOT appear on new user messages.
5. There should NOT be any new message notifications on messages that are sent by the user or have `scope` set to `internal`
6. After you've seen the new messages, all of the labels should be invisible

NOTE: Styling to be done in another issue. For now the new message icon is "[new_message_icon]" string. And the new message separator is experimentally implemented by me in CSS - so test for FUNCTIONALITY above all!


jswk (changes requested):
There seems to be a problem with unspecified ordering of entries in some calls in spec, have a look at failing cases. You should be able to specify that you want to compare regardless of ordering (see `match_array` or `contain_exactly`).

Marcelinna:
image
we should have icon here

jswk (commented):
As far as I'm concerned it looks fine now.

However, it would be safer to have the icon styling in place when merging this. Probably @jarekzet could assist in styling this element?

wujuu:
Ok, will add the icon in a minute

michal-szostak (approved):

jswk (approved):

Marcelinna (approved):

PR #2408, opened 2021-11-16
[#2393] Update the Wizzard Access error msg

### LINKED ISSUE #2393: More descriptive error when adding open access to the project second time
According to the discussion in the #2066 . 

The logic itself (user can not add open access resource without technical parameters more than once to the same project) is correct. But the error message should be more descriptive - **"Already pinned in the project <xyz>"**

image

danielkryska:
Closes #2393

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -110,4 +110,4 @@ en:
         project_item/wizard/summary_step:
           attributes:
             project:
-              repeated_in_project: "can only have one of this resource. Please choose another project"
+              repeated_in_project: "already pinned with this offer"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix RWD CSS Styles (@jarekzet)
 - Move authentication to the last step of the Access Wizard - User Experience improvement (@danielkryska)
 - Improved conversation view and styles (@jarekzet, @wujuu)
+- More descriptive error when adding open access to the project second time - User Experience improvement (@danielkryska)
 
 ### Deprecated
 
--- a/spec/features/order_spec.rb
+++ b/spec/features/order_spec.rb
@@ -148,7 +148,7 @@ RSpec.feature "Service ordering" do
 
         expect do
           click_on "Add to a project", match: :first
-          expect(page).to have_text("Project can only have one of this resource. Please choose another project")
+          expect(page).to have_text("already pinned with this offer")
         end.to change { ProjectItem.count }.by(0)
       end
     end


danielkryska:
localhost_5000_services_alien-and-invasive-species-virtual-research-environment_summary

config/locales/en.yml
```diff
@@ -113,4 +113,4 @@ en:
         project_item/wizard/summary_step:
           attributes:
             project:
-              repeated_in_project: "can only have one of this resource. Please choose another project"
+              repeated_in_project: "already pinned to the service %{service_name}"
```
kmarszalek, line 116: ``
repeated_in_project:
format: "Already pinned to the project"
``
and add config.active_model.i18n_customize_full_message = true in config/application.rb and you can make your own custom msgs :) 

source: https://www.bigbinary.com/blog/rails-6-allows-to-override-the-activemodel-errors-full_message-format-at-the-model-level-and-at-the-attribute-level

kmarszalek (dismissed):
I would make a custom msg, it works now. not sure why it wasn't working previously.

kmarszalek:
rebase and changelog :)

config/locales/en.yml
jswk: Rather:
```
project:
  repeated_in_project: "Already pinned to the project"
  format: "%{message}"
```

jswk (dismissed):
> rebase and changelog :)

Rebasing should also remove the superfluous commit.

config/locales/en.yml
kmarszalek: also works 🥇

kmarszalek:
> > rebase and changelog :)
> 
> Rebasing should also remove the superfluous commit.

that's the reason! :)

config/locales/en.yml
danielkryska: Sometimes I'm wondering if you're kind of evangelist... Have you considered that:
- format on the property level doesn't work
- forcing message template on the `project` model level will overwrite all error messages?
- the ticket is focusing on the simple change of text and creating additional layers of abstract magic is overthinking
- after a quick talk with the @michal-szostak value of the error content MUST NOT BE EXACTLY SAME AS IN TICKET

danielkryska: SUMMARIZING -> DO NOT OVERTHINK !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

kmarszalek: caps lock and exclamation marks aren't proper forms of conversation.

> * format on the property level doesn't work

well then pray tell, how I was able to make it work?

> * forcing message template on the `project` model level will overwrite all error messages?

on project model, yes. ok but.. you have all two messages for the project...

> * the ticket is focusing on the simple change of text and creating additional layers of abstract magic is overthinking

These additional layers of abstract magic make it work as intended in the ticket.

> * after a quick talk with the @michal-szostak value of the error content MUST NOT BE EXACTLY SAME AS IN TICKET

and where can I find any mention of that, as a reviewer I am looking in the ticket and PR description. 
if you do not put any effort into documenting your conversation or conclusions, how do you expect anyone to take it into account?

I am not an evangelist, I tried to make custom msg work a month ago but some dependency was able to thwart my effort. now it works, and it's not even too much work.

Zrzut ekranu 2021-11-17 o 09 28 27
Zrzut ekranu 2021-11-17 o 09 28 38

danielkryska:
The final error message version example:
`Project already pinned with the offer: General access`

Marcelinna:
the case when we have default offer
image

Marcelinna:
Taking into account the problems with the name of the offers, let's change to the **Project already pinned with this offer**

Marcelinna (approved):

michal-szostak (approved):

Issue #2411, opened 2021-11-17
Create JMS publisher singleton


danielkryska:
- [x] connection shouldn't be always opened
- [x] connection open only on publishing a message and closed after the action
- [x] JMS subscriber refactoring

unit tests extracted to #2418 (maybe later in the milestone)

Issue #2390, opened 2021-11-09
Send message to JMS about message to provider


michal-szostak:
**This ticket is blocked until communication details are resolved with PC** (do a provisional implementation for now)

After user asks question to provider in projects view apart from sending email to the said provider message to JMS should be send so that PC can show the message in its interface

jswk:
See also: #2185. For now just send the message alongside provider eid and optionally service id.

Issue #2389, opened 2021-11-09
Send message to JMS about resources coexistence


michal-szostak:
Another blocker **Create generic JMS publisher singleton #2411**
(@danielkryska)

---

**This ticket is blocked until communication details are resolved with PC**

Intro:
MP allows users to organize resources in projects. Projects are a logical , user driven way of combining resources in research project. In this requirement we simply care about the list of resources added together in a project . This information can be presented to the Providers dashboard and be exploited for future recommendation and other reporting purposes.
(Note that Logical Combining is different from resource composition, which requires more in-depth levels of technical interoperability).

Request:
Current status is that an event is generated and sent to the messaging service between UC and PC every time a user adds a resource to a project. This event simply indicates that this specific service has been added to a user project (e.g., similarly to the case a user adds a resource in a favorite list).
We need to extend this event (or generate a second one) with the list of resource already present in the project, in which the resource is added. In this case we need to know that
Resource A was added to a project which already contains Resource B, C, etc.
Alternatively, redundant events for Resource B, C, can be generated (technical details can be discussed)

jswk:
See also: https://jira.egi.eu/browse/EOSCENR-210.

PR #2412, opened 2021-11-17
Create JMS publisher

danielkryska:
Create JMS::Publisher.

## Local verification

To create a job, run `Jms::PublishJob.perform_later({foo: "bar"})` from the console.
If you have sidekiq running (with set `MP_STOMP_PUBLISHER_ENABLED=true`) then such messages should be sent to beta JMS, to queue `eosc-mp-beta`.

Example subscriber script:
```ruby
#!/usr/bin/env ruby
APP_ROOT = File.expand_path("..", __dir__)

require_relative "../config/boot"
require_relative "../config/environment"

def conf_hash(login, pass, host, client_name, ssl)
  {
    hosts: [
      {
        login: login,
        passcode: pass,
        host: "#{host}",
        port: 61613,
        ssl: ssl
      }
    ],
    connect_headers: {
      "client-id": client_name,
      "heart-beat": "0,20000",
      "accept-version": "1.2",
      "host": "localhost"
    }
  }
end

stomp_config = Mp::Application.config_for(:stomp_publisher)

@client = Stomp::Client.new(conf_hash(
                              stomp_config["login"],
                              stomp_config["password"],
                              stomp_config["host"],
                              "custom-cmd-1",
                              stomp_config["ssl-enabled"]))

@client.subscribe("/topic/eosc-mp-beta.>") do |msg|
  puts msg
end

raise ConnectionError.new("Connection failed!!") unless @client.open?()
raise ConnectionError.new("Connection error: #{@client.connection_frame().body}") if @client.connection_frame().command == Stomp::CMD_ERROR

@client.join
```

Closes: #2412.


danielkryska:
New ENV variables
- `STOMP_PUBLISHER_ENABLED` turn on publishing with JMS
- `MP_STOMP_LOGGER_PATH` path to the JMS log file, default: `"#{Rails.root}/log/jms.log"`
- `MP_STOMP_PUBLISHER_TOPIC` topic of a new message to be published

jswk:
What messages should go through this publisher to the JMS in the long run? From #2389 it seems that we should publish information about service/offer coexistence within projects. I wonder how does publishing messages which user sends to support within a project fit in this picture. It isn't described in #2411 and seems odd to me, as the provider's OMS should receive them through the Ordering API. Why push them through to the PC, which won't have information about the project and the project items, and no ability to respond to them?

jswk:
Ok, I see now that there is also #2390, which is not connected to this PR/issue.
However, it seems that despite #2390 saying "in projects view", it may refer to this modal and the `Provider::Question` model:
image

danielkryska:
@jswk yesterday o the daily meeting I and @michal-szostak mentioned that PC developers don't give us any details yet. That's why we are sending a message 1:1 with those sent with an email.

danielkryska:
In the future, we'll update the publisher code and message itself. Also, I changed original ticket to #2390

danielkryska:
"persistent": true,
"suppress_content_length": true,
"content-type": "application/json"

Those parameters are related to the known sidekick issue of cutting messages to the length of 256 chars. To suppress the "feature" they must to be set like that.

lib/jms/publisher.rb
```diff
@@ -0,0 +1,63 @@
...
+      def initialize(msg)
+        super(msg)
+        Sentry.capture_exception(self)
+      end
+    end
+
+    def publish(msg)
+      unless ENV["STOMP_PUBLISHER_ENABLED"]
+        return
+      end
```
jswk, line 19: Test that it actually behaves like a boolean, i.e. for `STOMP_PUBLISHER_ENABLED=false` it deactivates. Because currently probably not (i.e. `"false" => true`).

config/stomp_connection.rb
```diff
@@ -0,0 +1,46 @@
...
+                presence: true,
+                default: ENV["MP_STOMP_HOST"] || Rails.application.credentials.stomp.dig(:host) || "''"
+    attr_reader :name,
+                string,
+                presence: true,
+                default: ENV["MP_STOMP_CLIENT_NAME"] || "'MPClient'"
+    attr_reader :ssl_enabled,
+                  boolean,
+                  presence: true,
+                  default: ENV["MP_STOMP_SSL"] || false
```
jswk, line 24: Move these to config.

jswk (dismissed):

jswk:
BTW, the connection to the beta JMS works correctly. See the "Local verification" section in the main PR for how I tested it.

README.md
```diff
@@ -244,6 +244,15 @@ We are currently using the following ENV variables:
...
   * ENV Variables connected to JIRA integration are described in [JIRA integration manual](./docs/jira_integration.md)
+  * `MP_STOMP_CLIENT_NAME` (Optional) - stomp client name (default `MPClient`)
+  * `MP_STOMP_LOGIN` (Optional) - stomp client login (default from `credentials.stomp.login`)
+  * `MP_STOMP_PASS` (Optional) - stomp client password (default from `credentials.stomp.password`)
+  * `MP_STOMP_HOST` (Optional) - stomp client host (default from `credentials.stomp.host`)
+  * `MP_STOMP_DESTINATION` (Optional) - stomp client destination (default from `credentials.stomp.destination`)
+  * `MP_STOMP_SSL` (Optional) - stomp connection SSL (default `false`)
+  * `MP_STOMP_PUBLISHER_ENABLED` (Optional) - turn on publishing with JMS (default `false`)
+  * `MP_STOMP_PUBLISHER_LOGGER_PATH` (Optional) - path to the JMS log file (default `"RAILS_ROOT/log/jms.publisher.log"`)
+  * `MP_STOMP_PUBLISHER_TOPIC` (Optional) - topic of a new message to be published (default `eosc-mp-beta`)
```
kmarszalek, line 255: add mention of this for @wziajka in appropriate section 'docs/upgrade'

app/jobs/jms/publish_job.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def finalize
+      close_publisher!
+      @logger.close
+    end
+
+    def close_publisher!
+      @publisher_instance&.close
+    rescue Error => e
+      @logger.warn("Cannot close publisher #{e}")
```
kmarszalek, line 60: sentry?

app/services/jms/publisher.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class Jms::Publisher
+  class ConnectionError < StandardError
+    def initialize(msg)
+      super(msg)
+    end
+  end
+
+  def initialize(topic, login, pass, host, client_name,
```
kmarszalek, line 10: pasta?

app/services/jms/publish.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Jms::Publish
+  def initialize(message, publisher, logger)
+    @message = message
+    @publisher = publisher
+    @logger = logger
+  end
+
+  def call
+    serialized_message = @message.to_json
```
danielkryska, line 13: Why do you validate the existence of `to_json` function in the job instead of here?

danielkryska:
```
unless message&.respond_to?(:to_json)
      raise ArgumentError.new("Message must have a :to_json method, passed: #{message}")
    end

File: app/jobs/jms/publish_job.rb [hunk 7]
```diff
@@ -0,0 +1,63 @@
...
+    finalize
+  end
+
+  private
+    def logger_path
+      ENV["MP_STOMP_PUBLISHER_LOGGER_PATH"] || "#{Rails.root}/log/jms.publisher.log"
+    end
+
+    def publisher_disabled?
+      ENV["MP_STOMP_PUBLISHER_ENABLED"] != "true"
```
danielkryska, line 37: If you want to disable Publisher job rename ENV VAR to MP_STOMP_PUBLISHER_JOB_ENABLED or move logic to Publisher as it is.

README.md
```diff
@@ -244,11 +244,20 @@ We are currently using the following ENV variables:
   * `EOSC_COMMONS_ENV` - EOSC commons environment type enum: 'production' | 'development'
   * `EOSC_COMMONS_BASE_URL` - EOSC commons base URL: s3 instance + bucket 
   * ENV Variables connected to JIRA integration are described in [JIRA integration manual](./docs/jira_integration.md)
+  * `MP_STOMP_CLIENT_NAME` (Optional) - stomp client name (default `MPClient`)
+  * `MP_STOMP_LOGIN` (Optional) - stomp client login (default from `credentials.stomp.login`)
```
danielkryska, line 248: Do we have any mention of the encrypted file with credentials? If not maybe here is the place for it.

app/jobs/jms/publish_job.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require "sentry-ruby"
+
+class Jms::PublishJob < ApplicationJob
```
danielkryska, line 5: I'm not sure if keeping logic in here except the call to service is consistent with other Jobs. See:

```
class Project::ProjectRegisterJob < ApplicationJob
  queue_as :orders

  rescue_from(Jira::Client::JIRAIssueCreateError) do |exception|
    # TODO: we need to define what to do when question registration in e.g.
    #       JIRA fails. Maybe we should report this problem to Sentry and
    #       do some manual intervantion?
    raise exception
  end

  rescue_from(StandardError) do |exception|
    # This is general error, which should not occur, but should be
    # caught just in case
    raise exception
  end

  def perform(project)
    Project::Register.new(project).call
  end
end
```

```
class Project::JiraUpdateJob < ApplicationJob
  queue_as :orders

  def perform(project)
    Project::JiraUpdate.new(project).call
  end
end
```

```
class ProjectItem::ReadyJob < ApplicationJob
  queue_as :orders

  def perform(project_item, message = nil)
    ProjectItem::Ready.new(project_item, message).call
  end
end
```

... and so on...

app/jobs/jms/publish_job.rb
```diff
@@ -0,0 +1,63 @@
...
+  queue_as :pc_publisher
+
+  rescue_from ArgumentError, Jms::Publisher::ConnectionError, Stomp::Error do |e|
+    @logger.warn("Exception occurred: #{e}")
+    Sentry.capture_exception(e)
+
+    finalize
+  end
+
+  def perform(message)
```
danielkryska, line 15: We're using consistently word call on instances. Why have you used `perform`? (Go to jobs code I've added in PublishJob section)

app/jobs/jms/publish_job.rb
danielkryska: `From YAML 1.2 the bool type is defined as follows: [ true, True, false, FALSE ]`

What if someone set a boolean value instead of string?

docs/upgrade/3.26.0.md
```diff
@@ -0,0 +1,21 @@
...
+- make database dump and all application files.
+- `bundle install --deployment --without development test`
+- `bundle exec rake assets:clean assets:precompile`
+- `rails db:migrate`
+
+# Additional info
+
+## Added JMS::Publisher
+
+The publisher has to be explicitly enabled by setting `MP_STOMP_PUBLISHER_ENABLED=true`, otherwise it will stay disabled.
```
danielkryska, line 16: Won't be easier to write: `The publisher is turned off by default. Can be enabled by adding 'MP_STOMP_PUBLISHER_ENABLED=true' to the .env file.`

app/services/jms/publish.rb
jswk: I wanted to validate the message sooner than later. And I would keep the validation in that place.

However, the Jms::Publish service can then accept an already serialized (after calling .to_json) message. I will modify it that way.

app/jobs/jms/publish_job.rb
jswk: > If you want to disable Publisher job rename ENV VAR to MP_STOMP_PUBLISHER_JOB_ENABLED or move logic to Publisher as it is.

I'd prefer to keep the enable/disable logic outside of the actual publisher implementation. I was considering creating a factory for returning a new publisher, but decided it's too complex for the use case, thus the placement.

About the name, I would argue for the current name, as the prefix `MP_STOMP_PUBLISHER` houses all publisher related values, and doesn't really adhere to the class structure. I would rather look at it functionally, i.e., that the env disables the publisher, but the check doesn't have to take place directly in the publisher class.

jswk: > What if someone set a boolean value instead of string?

I understand, that ENVs are always strings. See, e.g. https://man7.org/linux/man-pages/man7/environ.7.html.
YAML is only an extra level of abstraction, these values will translate to strings at the level of ENV.

danielkryska: Ok, it does have a sense now.

jswk (approved):
I'm self accepting since @danielkryska approved the corrections (and he cannot be added as a reviewier to his own PR).

kmarszalek (approved):

PR #2413, opened 2021-11-17
[#2066] Fix logic of adding offers to a project

### LINKED ISSUE #2066: wrong logic of adding offers to a project
tested on 3.11.0-beta.2
**Describe the bug**
the current logic does not allow adding two different offers of the same service to the same project if they do not have the properties (constant, input, select, multiselect, date, range, quantity price)

I order offer in existed services EGI Cloud Compute
1. Add **General purpose** (order required **with properties**) offer to **"project 1"** - it is ok
2. Add second time the same offer **General purpose** (order required **with properties**) offer to "**project 1"** - it is ok
3. Add other offert **Compute-intensive** (order required **with properties**) offer to **"project 1"** - it is ok
4. Add second time offert **Compute-intensive** (order required **with properties**) offer to **"project 1"** - it is ok

I order offer in existed services da|ra - DOI Registration Service for social science and economic data
1. Add **DOI Registration**  (order required **without properties**) offer to **"project 2"**- it is ok
2. Add second time **DOI Registration**  (order required **without properties**) offer to **"project 2"**- isn't ok, information is displayed _Please note that this resource can be added to the Project only once. Please choose another project_
3. Add other **Metadata harvesting** (open access **without properties**) offer to **project 2** - isn't ok _Please note that this resource can be added to the Project only once. Please choose another project_
4. Create **new offer with properties**, order required type and add to the **project 2** - it is ok

The same situation for default offert is the same, If default offer has properties can be added second time to the same project. if it doesn't have properties it can't be added second time. 
The question is whether properiets should decide about this, because it will be enough that it will be constant which makes almost no distinction between the offer with properties and without.

**Expected behavior**
We should decide how it should works. 

**EDIT**
The follow up development issue has been created #2393

@roksanaer @bwilk 

**Screenshots**

**Additional context**

kmarszalek:
open_access/fully_open_access/other offer can be added only once.
order_required - external can be added only once.
order_required - internal can be added multiple times.
bundled offer (parent and child) can be added multiple times.
child of the bundled offer can be added once if it is oa/foa/oo/or-e offer if it is ordered solo.

Closes #2066

### DIFF
--- a/spec/factories/services.rb
+++ b/spec/factories/services.rb
@@ -10,8 +10,14 @@ FactoryBot.define do
     factory :open_access_service do
       sequence(:order_type) { :open_access }
     end
-    factory :fully_open_access_servoce do
+    factory :fully_open_access_service do
       sequence(:order_type) { :fully_open_access }
+    end
+    factory :other_service do
+      sequence(:order_type) { :other }
+    end
+    factory :order_required_service do
+      sequence(:order_type) { :order_required }
     end
     sequence(:webpage_url) { "https://wabpage.url"  }
     sequence(:manual_url) { "https://manual.url"  }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Improved conversation view and styles (@jarekzet, @wujuu)
 - More descriptive error when adding open access to the project second time - User Experience improvement (@danielkryska)
 - Disabled possibility to send an empty resource's review (@goreck888)
+- Clearing logic of adding offers to a project (@kmarszalek)
 
 ### Deprecated
 
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -83,18 +83,6 @@ class Offer < ApplicationRecord
     super || service.service_type
   end
 
-  def open_access?
-    offer_type == "open_access"
-  end
-
-  def normal?
-    offer_type == "normal"
-  end
-
-  def catalog?
-    offer_type == "catalog"
-  end
-
   def bundle?
     bundled_offers_count.positive?
   end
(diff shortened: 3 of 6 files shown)


jswk:
> **Fix** logic of adding offers to a project

How about simplify? It's actually about clearing up the logic there.

Marcelinna:
I can add many time the same offer open access 
image

spec/factories/services.rb
```diff
@@ -10,9 +10,15 @@
     factory :open_access_service do
       sequence(:order_type) { :open_access }
     end
-    factory :fully_open_access_servoce do
+    factory :fully_open_access_service do
       sequence(:order_type) { :fully_open_access }
     end
+    factory :other do
```
jswk, line 16: => `other_service`

spec/factories/services.rb
```diff
@@ -10,9 +10,15 @@
...
+      sequence(:order_type) { :other }
+    end
+    factory :order_required do
```
jswk, line 19: => `order_required_service`

jswk (changes requested):

jswk (approved):

Marcelinna:
Steps to reproduce:
1. Go to /services
2. Choose B2FIND
image
3. Order the first offer
4. Order the second offer - "Project already pinned with this offer" error message is displayed

Expected result: second offer should be ordered properly

Marcelinna:
Tests: 

1. Add offer open access
2. Add offer full open access
3. Add offer other
4. Add order required internal
5. Add order required external
6. Try to add second time the same offer open acccess 
7. Try to add second time the same offer full open acccess 
8. Add second time the same offer order required internal
9. Try to add second time the same offer order required external
10. Try to add second time the same offer other
11. Try add offer open access and order required internal from the same service
12. Try add offer order required and open access from the same service
13. Try add two open access offer from the same services
14. Try add two order required internal offer from the same services

Marcelinna:
- [ ] QA
- [ ] DEV

Steps to reproduce

1. Add to project service open access
2. Try add the same service open access to the same project
3. See error - user is redirected to previous step
image

jswk (approved):

Marcelinna (approved):

Issue #2400, opened 2021-11-15
Create data.yml and rake task for e2e test


Marcelinna:
**Is your feature request related to a problem? Please describe.**
e2e.yml uses data.yml and dev:prime

**Describe the solution you'd like**
create new data set and task for e2e tests

EDIT:
in case of any doubts contact @kmarszalek

PR #2440, opened 2021-12-02
[#2400]  Add data_tests.yml and rake task for e2e tests

Marcelinna:


kmarszalek (approved):
it works just fine.
data_test could be smaller, or more test-oriented. but I understand that it will be changed while doing new tests.
it's a good start.

EDIT: 
maybe add some entry in readme about e2e tests and data_test file?

lib/tasks/dev.rake
```diff
@@ -23,6 +23,24 @@ namespace :dev do
...
+    create_vocabularies
+    create_categories(yaml_hash["categories"])
+    create_providers(yaml_hash["providers"])
+    create_scientific_domains(yaml_hash["domain"])
+    create_platforms(yaml_hash["platforms"])
+    create_target_users(yaml_hash["target_users"])
+    create_services(yaml_hash["services"])
+    create_relations(yaml_hash["relations"])
+
+    OrderingApi::AddSombo.new.call
```
jswk, line 39: I'd extract this part to a separate method and only change the file path, i.e.
```ruby
  desc "Sample data for local development environment"
  task prime: "db:setup" do
    create_all_from_path("db/data.yml")
    puts "Done!"
  end

  desc "Sample data for e2e tests"
  task prime_tests: "db:setup" do
    create_all_from_path("db/data_tests.yml")
    puts "Done!"
  end

  def create_all_from_path(path)
    yaml_hash = YAML.load_file(path)

    create_vocabularies
    create_categories(yaml_hash["categories"])
    create_providers(yaml_hash["providers"])
    create_scientific_domains(yaml_hash["domain"])
    create_platforms(yaml_hash["platforms"])
    create_target_users(yaml_hash["target_users"])
    create_services(yaml_hash["services"])
    create_relations(yaml_hash["relations"])

    OrderingApi::AddSombo.new.call
  end
```

Less repetition this way.

lib/tasks/dev.rake
```diff
@@ -23,6 +23,24 @@ namespace :dev do
     puts "Done!"
   end
 
+  desc "Sample data for e2e tests"
+  task prime_tests: "db:setup" do
+    yaml_hash = YAML.load_file("db/data_tests.yml")
```
jswk, line 28: You aren't actually using it for rspec tests, so I would consider naming the task and the file: `prime_e2e` and `db/data_e2e.yml`, respectively. It seems to me that it would convey the meaning better.

jswk (approved):
Apart from the suggested improvements it's fine.

test/system/README.md
```diff
@@ -62,4 +62,6 @@ To run e2e tests on URL other than `localhost` preset variable `base url` by:
 - Pull request (nrPR = Pull request number in github): http://pr-nrPR.docker-fid.grid.cyf-kr.edu.pl/
 
 ### Running on
-- Pull Requests Github Actions with title `Integration tests`
\ No newline at end of file
+- Pull Requests Github Actions with title `Integration tests`
+- Filling the database is done by parsing yaml: `db/data_tests.yml` for `Integration tests`
+- Run ./bin/rails dev:prime_tests to seed the database as in Pull Requests Github Actions
```
jswk, line 67: Why is this under `Running on`?

This would rather fit in `How to run the test suite?`, as a preparation for running the suite locally. Also, use the quotes for commands, i.e. `./bin/rails dev:prime_tests` (not ./bin/rails dev:prime_tests).

jswk (approved):
Perfect, thanks!

PR #2450, opened 2021-12-09
Run default rubocop, turn off pre-push rspec run

danielkryska:
Turing on rubocop in rubymine:
https://www.jetbrains.com/help/ruby/rubocop.html#disable_rubocop


app/components/conversations/message_box_component.rb
```diff
@@ -8,12 +8,12 @@ def initialize(message, message_post_path, message_input_label)
   end
 
   def render?
-    if @message.messageable_type == "Project"
-      project = @message.messageable
+    project = if @message.messageable_type == "Project"
```
kmarszalek, line 11: do you know which rule changed this code?

.github/workflows/styles.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - name: Sign overcommit
       run: |
-        bundle exec overcommit --sign
+        bundle exec overcommit --sign && bundle exec overcommit --sign pre-commit
```
jswk, line 14: I would verify this is necessary

.rubocop.yml
```diff
@@ -1,9 +1,10 @@
 inherit_gem:
-  rubocop-rails_config:
-    - config/rails.yml
+  prettier: rubocop.yml
```
jswk, line 2: If you do:
```yml
inherit_gem:
  prettier: rubocop.yml
  rubocop-rails_config:
    - config/rails.yml
```
then you don't have to add all the exclusions at the end of the file. Only the cyclomatic complexity exclude for `'app/services/project_item/create.rb'`.

jswk (commented):
Running `rbprettier --write .` changes a number of files, for starters:
image

I would also remove the scripts (`pre-commit-prettier-*`) from root, as we should check all the files in the pre-commit. Otherwise no files will be cheked in CI.

danielkryska:
Automatic check issue, that can't be reproduced manually (skipped temporarily):

```
scenario "sorting will set query param and preserve existing ones", js: true do
      create(:service, name: "DDDD Something 1", rating: 4.1)
      create(:service, name: "DDDD Something 2", rating: 4.0)
      create(:service, name: "DDDD Something 3", rating: 3.9)

      visit services_path(q: "DDDD Something", sort: "rating")

      expect(page).to have_text("by rate 1-5")
      expect(page).to have_select("sort", selected: "by rate 1-5")
      expect(page).to_not have_select("sort", selected: "by rate 5-1")
    end
```

app/policies/service_context_policy.rb
```diff
@@ -24,7 +24,6 @@ def self.permitted?(user, service, from_backoffice = false)
     true
   end
 
-  private
 
   def self.has_public_access(record)
     record.published? || record.unverified? || record.errored?
```
jswk, line 28: Since it seems the intention was for the following methods to be private (but they weren't...), please actually make them private, i.e.
```ruby
    private_class_method def self.has_public_access(record)
      record.published? ||
        record.unverified? ||
        record.errored?
    end

    private_class_method def self.has_additional_access(user, record)
      if user.blank?
        return false
      end

      user.service_portfolio_manager? ||
        record.owned_by?(user) ||
        record.administered_by?(user)
    end
```

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_16_143044) do
+ActiveRecord::Schema.define(version: 20_211_116_143_044) do
```
jswk, line 13: The `db/schema.rb` file should be ignored as it's autogenerated.

When removing the rails rubocop config, you should include some of the excludes from it anyway. See https://github.com/toshimaru/rubocop-rails_config/blob/main/config/rails.yml#L12 for source.
I would propagate the following excludes:
```yml
AllCops:
  Exclude:
    - '**/tmp/**/*'
    - '**/vendor/**/*'
    - '**/node_modules/**/*'
    - 'bin/**/*'
    - 'db/schema.rb'
```

jswk (changes requested):
Much easier to check thanks to splitting into commits!

app/helpers/image_helper.rb
```diff
@@ -89,7 +89,6 @@ def self.base_64_extension(base_64)
     extension
   end
 
-  private
 
   def self.get_file_extension(file_path)
     file_name = file_path.split("/")[-1]
```
jswk, line 91: Also a problem with class_private here.

app/models/project_item/wizard.rb
```diff
@@ -30,7 +30,6 @@ def step_names
   class InvalidStep < StandardError
   end
 
-  private
 
   class Base
     include ActiveModel::Model
```
jswk, line 34: Also a problem with class_private here, but I don't know if there's a way to make nested classes private this way.

app/controllers/concerns/service/searchable.rb
```diff
@@ -115,8 +115,8 @@ def filter_constr(filters, constr = {})
 
   def highlights(from_search)
     result = from_search.try(:with_highlights) if (params[:q]&.size || 0) > 2
-
-    result || [].to_h.transform_keys { |s| s.id } || {}
+    return {} if result.blank?
+    result.to_h.transform_keys { |s| s.id }
```
danielkryska, line 119: turn off and add a comment to the hash rule in .rubocop.yml file

app/controllers/concerns/service/searchable.rb
```diff
@@ -72,17 +72,16 @@ def filter_counters(scope, filters, current_filter)
     {}.tap do |hash|
       unless current_filter.index.blank?
         services = search_for_filters(scope, filters, current_filter)
-        services.aggregations[current_filter.index][current_filter.index]["buckets"].inject(hash) do |h, e|
+        services.aggregations[current_filter.index][current_filter.index]["buckets"].each_with_object(hash) do |e, h|
           h[e["key"]] = e["doc_count"]
-          h
```
danielkryska: to @danielkryska: is the returning value influence anything?

app/controllers/tour_feedbacks_controller.rb
```diff
@@ -18,11 +18,9 @@ def create
 
     @form = params.require(:content).permit(@feedback["questions"].map { |question| question["name"].to_sym })
     @errors =
-      Hash[
-        @feedback["questions"]
+      @feedback["questions"]
           .select { |question| !@form[question["name"]].present? }
-          .collect { |question| [question["name"], "This field is required"] }
-      ]
+          .collect { |question| [question["name"], "This field is required"] }.to_h
```
danielkryska: to @danielkryska: refactor/update hash change

app/controllers/users/auth_mock_controller.rb
```diff
@@ -8,7 +8,7 @@ def login
 
     return unless Mp::Application.config.auth_mock && Rails.env.development?
 
-    encrypted_password = ::BCrypt::Password.create("#{params[:password]}#{nil}", cost: 11).to_s
+    encrypted_password = ::BCrypt::Password.create("#{params[:password]}nil", cost: 11).to_s
```
danielkryska, line 11: to @danielkryska: update this line of code and check rue providing this change

app/helpers/scientific_domains_helper.rb
```diff
@@ -6,11 +6,10 @@ def grouped_scientific_domains
     ScientificDomain.arrange.each { |k, v| group_scientific_domains!(result, "/", k, v) }
 
     result
-      .inject({}) do |acc, record|
+      .each_with_object({}) do |record, acc|
```
danielkryska, line 9: check if it works correctly

app/models/attribute.rb
```diff
@@ -70,19 +70,18 @@ def to_json
...
             String(param)
           end
       when "string"
-        @value = String(param)
+        String(param)
       else
-        @value = param
-      end
+        param
+               end
```
danielkryska: check rule providing change and turn it off

app/models/attribute.rb
```diff
@@ -105,24 +104,24 @@ def config_schema
   def self.from_json(json)
     JSON::Validator.validate!(ATTRIBUTE_SCHEMA, json)
 
-    case json["type"]
+    attr = case json["type"]
```
danielkryska: check rule providing change and turn it off

app/models/attribute.rb
```diff
@@ -134,7 +133,6 @@ def self.from_json(json)
     attr
   end
 
-  protected
```
danielkryska, line 139: check rule providing change and turn it off, refactor

app/models/attribute/date.rb
```diff
@@ -26,7 +26,6 @@ def date_format
     I18n.t("date.formats.default")
   end
 
-  protected
```
danielkryska, line 36: check rule providing change and turn it off, refactor

app/models/attribute/input.rb
```diff
@@ -3,7 +3,6 @@
 require "json-schema"
 
 class Attribute::Input < Attribute
-  protected
```
danielkryska, line 6: check rule providing change and turn it off, refactor

app/models/attribute/multiselect.rb
```diff
@@ -35,15 +35,14 @@ def value_schema
 
   def value_from_param(param)
     param = param.reject(&:blank?)
-    case @value_type
+    @value = case @value_type
```
danielkryska: check rule providing change and turn it off, refactor

app/models/attribute/range.rb
```diff
@@ -19,17 +19,16 @@ def value_schema
...
+        begin
             Integer(param)
           rescue StandardError
             String(param)
           end
       else
-        @value = param
-      end
+        param
+               end
```
danielkryska: check rule providing change and turn it off, refactor

app/models/attribute/range.rb
```diff
@@ -54,7 +53,6 @@ def config_schema
     }
   end
 
-  protected
```
danielkryska, line 55: check rule providing change and turn it off, refactor

app/models/attribute/range_property.rb
```diff
@@ -18,11 +18,10 @@ def value_schema
     }
   end
 
-  def value_from_param(param)
+  def value_from_param(_param)
     raise "not implemented yet"
   end
 
-  protected
```
danielkryska, line 25: check rule providing change and turn it off, refactor

app/models/attribute/select.rb
```diff
@@ -26,21 +26,19 @@ def config_schema
   def value_from_param(param)
     param = param.reject(&:blank?)
     if param.length > 0
-      case @value_type
+      @value = case @value_type
       when "integer"
-        @value =
-          begin
+        begin
```
danielkryska: check rule providing change and turn it off, refactor

app/models/filter.rb
```diff
@@ -4,7 +4,7 @@ class Filter
   attr_accessor :counters
   attr_reader :title, :field_name, :type, :index
 
-  def initialize(params: {}, field_name:, type:, title:, index:)
+  def initialize(field_name:, type:, title:, index:, params: {})
```
danielkryska, line 7: check if code is working correctly

app/models/jira/checker.rb
```diff
@@ -197,28 +187,23 @@ def check_webhook!(host)
...
-          else
-            raise CheckerWarning.new(
-                    "Webhook \"#{wh.name}\" does not define proper \"Issue related events\" - required: " +
+        next unless wh.attrs["url"] == (host + api_webhooks_jira_path + "?issue_id=${issue.id}")
+        raise CheckerWarning, "Webhook \"#{wh.name}\" is not enabled" unless wh.enabled
+
+        if wh.filters["issue-related-events-section"].match?(/project = #{self.client.jira_project_key}/)
+          webhook = wh
+        else
+          raise CheckerWarning, "Webhook \"#{wh.name}\" does not define proper \"Issue related events\" - required: " +
```
danielkryska: check rule providing change and turn it off, refactor

config/initializers/split.rb
```diff
@@ -6,14 +6,14 @@
...
-    expire_seconds: 2592000)
+    expire_seconds: 2_592_000)
 
 Split.configure do |config|
   config.persistence = Split::Persistence::DualAdapter.with_config(
       logged_in: -> (context) { context.request.env["warden"]&.user&.id.present? },
       logged_in_adapter: redis_adapter,
       logged_out_adapter: cookie_adapter)
-  config.persistence_cookie_length = 2592000 # 30 days
+  config.persistence_cookie_length = 2_592_000 # 30 days
```
danielkryska, line 16: turn off rule changing numbers formatting

app/components/conversations/message_box_component.rb
danielkryska: Rules providing changes:|
- Lint/UselessAccessModifier: Useless private access modifier
- Lint/IneffectiveAccessModifier: private (on line 27) does not make singleton methods private. Use private_class_method or private inside a class << self block instead

app/policies/service_context_policy.rb
jswk: Okay, so you chose to ignore the rubocop suggestions.
This makes sense too, we will turn these cops on one a time in future and correct it.

app/helpers/image_helper.rb
jswk: Okay, so you chose to ignore the rubocop suggestions.
This makes sense too, we will turn these cops on one a time in future and correct it.

app/models/project_item/wizard.rb
jswk: Okay, so you chose to ignore the rubocop suggestions.
This makes sense too, we will turn these cops on one a time in future and correct it.

jswk (commented):
This looks good now. I would just move the `.rubocop` changes to the first PR commit and generate all the changes from scratch.
In current commits you have for example missed the `protected` keyword from `app/models/attribute/quantity_price.rb`, running the formatters again would just ignore these.

I would also split the changes into four commits within this PR:
1. configuration changes etc. (the same as the current first commit)
2. prettier changes (the same as the current second commit)
3. automatic rubocop fixes (incorporating necessary prettier runs)
4. manual rubocop fixes (incorporating necessary prettier runs)

Gemfile
```diff
@@ -5,61 +5,61 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby "3.0.2"
 
-gem "rails", "6.1.4.1"
+gem "nori"
 gem "pg", ">= 0.18", "< 2.0"
 gem "puma"
-gem "nori"
+gem "rails", "6.1.4.1"
```
jswk, line 11: And I would still disable this, i.e. turn off the `Bundler/OrderedGems` cop.
It will scramble Gemfile history otherwise.

app/models/attribute/quantity_price.rb
```diff
@@ -43,11 +43,11 @@ def value_type
   end
 
   def value_valid?
-    begin
+
```
kmarszalek: it will take time to remember this rule. luckily, it's automatic.

jswk (approved):

Marcelinna (approved):

kmarszalek (approved):

PR #2451, opened 2021-12-09
Add e2e tests for provider and resources in BO

Marcelinna:
Adjust data_e2e.yml
Add test for delete provider in BO
Change test for Preview mode in BO

### DIFF
--- a/test/system/cypress/integration/regression_tests/backoffice/providers/providers.spec.ts
+++ b/test/system/cypress/integration/regression_tests/backoffice/providers/providers.spec.ts
@@ -3,7 +3,7 @@ import { UserFactory } from "../../../../factories/user.factory";
 
 describe("Providers", () => {
   const user = UserFactory.create({ roles: ["service_portfolio_manager"] });
-  const [provider, provider2, provider3] = [...Array(4)].map(() =>
+  const [provider, provider2, provider3] = [...Array(3)].map(() =>
     ProvidersFactory.create()
   );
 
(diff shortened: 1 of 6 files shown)


test/system/cypress/integration/regression_tests/backoffice/providers/providers.spec.ts
```diff
@@ -176,6 +176,64 @@ describe("Providers", () => {
       });
   });
 
+  it("should deleted provider with resource with deleted status", () => {
```
jswk, line 179: Shouldn't it read `should delete provider...`? The current grammar is incorrect.
I would correct it in the tests below too.

test/system/cypress/integration/regression_tests/backoffice/resources/owned_resources.spec.ts
```diff
@@ -206,19 +206,23 @@ describe("Owned resources", () => {
...
+
+  it("shouldn't go to Preview mode with wrong data format", ()=> {;
+    cy.visit("/backoffice/services/new");
+    cy.fillFormCreateResource({...resource5, basicWebpage_url:"wrongFormat", contactsEmail:"wrongFormat"}, correctLogo);
     cy.get("[data-e2e='preview-btn']")
-        .click();
+      .click();
     cy.contains("div.invalid-feedback", "Email is not a valid email address")
         .should("be.visible");
     cy.contains("div.invalid-feedback", "Webpage url is not a valid URL")
```
jswk, line 228: In these two tests you verify the validation works for logo and other fields. However, you pass both incorrect logo and other fields to `fillFormCreateResource` in both cases. I would either pass only the incorrect field(s) for each case, or verify both the other fields and the logo in a single test.

jswk (changes requested):

test/system/cypress/integration/regression_tests/backoffice/resources/owned_resources.spec.ts
Marcelinna: I can't directly verify the other fields and the logo in a single test, because there is a bug in the application. 
I want to rewrite the previous test for Preview mode because it fails too often.

jswk: Okay, makes sense!

jswk (approved):

kmarszalek (approved):

PR #2456, opened 2021-12-14
Upgrade ruby to 3.0.3

jswk:
Three CVEs we're not directly affect by are resolved in this security release: https://www.ruby-lang.org/en/news/2021/11/24/ruby-3-0-3-released/. So to be on the safe side...


kmarszalek (approved):

jswk:
Rails 6.1.4.1 isn't compatible with ruby 3.1.0 yet, so stick to 3.0.x until support is introduced.

wziajka (approved):

jswk:
As per @wziajka suggestions, wait with the patch update until other sprint PRs are merged.

PR #2459, opened 2021-12-16
Change wizard description for other order type resources

Marcelinna:


jswk (commented):
You will need to correct the spec here too `spec/components/services/information_description_component_spec.rb`, as it verifes a correct text is generated for different order types.

jswk (approved):

kmarszalek (approved):

PR #2470, opened 2021-12-23
Rubocop rules extension 3

danielkryska:
For the review, some of the cop changes are split into separate commits.
Most of them (without prominent logic changes) will be squashed before merging.

-------------------------

Some Cops enabling commits are squashed together. However, some remain expanded to be easier to analyse in the future.


app/helpers/image_helper.rb
```diff
@@ -3,18 +3,10 @@
...
-    @@default_logo_path
-  end
-
-  def self.permitted_ext_message
-    @@permitted_ext_message
-  end
+  DEFAULT_LOGO_PATH = Webpacker.manifest.lookup("media/images/eosc-img.png")
 
   def self.to_temp_file(logo, ext)
     tmp_logo = Tempfile.new(["logo_temp", ext])
```
Marcelinna, line 12: changes in  app/helpers/image_helper.rb looks fine? 
On PR is a problem with the image, so it is hard to check it.

jswk: Yes, they're fine. There were several affected tests in `spec/lib/import/resources_spec.rb`, which traced the modified code paths and are now corrected.

app/controllers/favourites/services_controller.rb
```diff
@@ -22,7 +22,7 @@ def update
   private
 
   def many?
-    UserService.where(user: current_user).size > 1 || cookies[:favourites]&.size > 1
+    UserService.where(user: current_user).size > 1 || (cookies[:favourites]&.size || 0) > 1
```
kmarszalek, line 25: brackets

app/helpers/scientific_domains_helper.rb
```diff
@@ -18,11 +18,11 @@ def grouped_scientific_domains
   private
 
   def group_scientific_domains!(result, current_path, key, values)
-    if !values.empty?
```
kmarszalek, line 21: unless [']

kmarszalek (changes requested):
git. git pl
squash smth, after Marcelinas approval and add brackets.

Marcelinna (approved):

kmarszalek (approved):

PR #2474, opened 2021-12-30
[#2446] Fix updating resources with invalid providers

### LINKED ISSUE #2446: Error when updating resource caused by wrong provided by
@Marcelinna - please fill in the details

1. Go to PC
2. Select one of the resources
3. Fill field _Resource Providers_ by the provider for which there was a synchronization error in MP  (e.g the logo is not correct, the data format is not correct), edit the Webpage (*)  field
4. Click submit in PC - Resource is submitted correctly in PC
6. Go to MP
7. See problem Resource isn't udpate in MP

When resource admin in PC try to add a provider for which there was a synchronization error in MP to field Resource Providers, updating the whole resource in MP fails.

Expected result: the service should be updated correctly, the external provider should not affect the updating of the service

goreck888:
This PR fixes adding providers to the resource
even if chosen provider has improper data
Fixes #2446

### DIFF
--- a/app/controllers/backoffice/services_controller.rb
+++ b/app/controllers/backoffice/services_controller.rb
@@ -163,7 +163,7 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
     if Service::Update.new(@service, attributes).call
       update_logo_from_session!
       session.delete(preview_session_key)
-      redirect_to backoffice_service_path(@service), notice: "Service updated correctly"
+      redirect_to backoffice_service_path(@service), notice: "Resource updated correctly"
     else
       add_missing_nested_models
       render :edit, status: :bad_request
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Please view this file on the master branch, on stable branches it's out of date.
 
+## [3.29.0-milestone]
+
+### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+- Adding providers with improper data to the resource (@goreck888)
+
+### Security
+
 ## [3.28.0] 2021-12-22
 
 ### Changed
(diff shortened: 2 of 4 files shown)


spec/features/backoffice/services_spec.rb
```diff
@@ -206,6 +206,26 @@
...
+      select invalid_provider.name, from: "Providers"
+
+      click_on "Update Resource"
+
+      expect(page).to have_content("Resource updated correctly")
+
+      service.reload
+
+      expect(service.providers.map(&:id)).to eq([invalid_provider.id])
+    end
```
jswk, line 227: Actually, when updating through `record.update(params)` as is done in `backoffice/services_controller.rb`, the failing validation doesn't stop saving. So this test is unnecessary - it also doesn't fail when the `validate: false` is removed.
I would remove it.

jswk (changes requested):
Correct one test and strip the other as described. Otherwise it's good to go.

jswk:
Aah, and rebase onto the new master.

spec/features/backoffice/services_spec.rb
goreck888: Ok.

jswk (approved):

app/controllers/backoffice/services_controller.rb
```diff
@@ -163,7 +163,7 @@ def perform_update
     if Service::Update.new(@service, attributes).call
       update_logo_from_session!
       session.delete(preview_session_key)
-      redirect_to backoffice_service_path(@service), notice: "Service updated correctly"
+      redirect_to backoffice_service_path(@service), notice: "Resource updated correctly"
     else
       add_missing_nested_models
       render :edit, status: :bad_request
```
Marcelinna, line 169: You can change to "New service created successfully"

Marcelinna (approved):

app/controllers/backoffice/services_controller.rb
goreck888: Done

kmarszalek (approved):

PR #2475, opened 2022-01-03
[#2446] Add missing text improvements

goreck888:


jswk (commented):
In general ok, although I would go through messages in the entire project looking for "services".
For example: there's another instance of "service" in `app/controllers/backoffice/services_controller.rb`. Maybe it's all, but I would grep the project nonetheless.

jswk:
Have a look in particular at the files in `config/locales`. We have some instances of "service" there too.

Marcelinna:
changelog?

Marcelinna:
It nice to change message "Service ordered successfully" in app/controllers/services/summaries_controller.rb, depending on the type of offer. 

"Offer ordered successfully" - order required internal
"Offer pinned successfully" other

config/locales/simple_form.en.yml
```diff
@@ -78,11 +78,11 @@ en:
           them on home page
       report:
         text: |
-          We will send your report to our customer service.
+          We will send your report to our customer resource.
```
jswk, line 81: This should stay "customer service". It's not about services/resources in the meaning of Resource Profile.

jswk (changes requested):
Please also update the commit message, as it seems to have been copied from the current master's HEAD.

config/locales/simple_form.en.yml
goreck888: Sorry, I remember that text and i had to click a wrong button, thanks!

Marcelinna:
image
here, we should change to Resource?

Marcelinna:
in mail notification we have service too
image

jswk:
The commit message is still wrong.

Please also correct the inconsistencies in [emails](https://github.com/cyfronet-fid/marketplace/pull/2475#issuecomment-1004661167) and [profile](https://github.com/cyfronet-fid/marketplace/pull/2475#issuecomment-1004656043) identified by @Marcelinna.

jswk (changes requested):
Push to 462ee85ee7b8e373cdd8455dcd44e0390451db87 handled the profile text, yet not all text in emails is corrected. Go through all the email texts and subjects to catch them all.

jswk (changes requested):
Please correct failing tests.

Marcelinna (approved):

jswk (approved):

Issue #2465, opened 2021-12-21
Add docker-compose.yml


michal-szostak:
Move from asdf to docker-compose.yml for services (elastic, postgres, etc...)
Tools as nodejs should still be handled by asdf.

PR #2476, opened 2022-01-03
[#2465] Add docker-compose.yml to manage services

jswk:
Closes: #2465.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -24,9 +24,9 @@ default: &default
 development:
   <<: *default
   database: mp_development
-  host: <%= ENV["MP_DATABASE_HOST"] %>
-  username: <%= ENV["MP_DATABASE_USERNAME"] %>
-  password: <%= ENV["MP_DATABASE_PASSWORD"] %>
+  host: <%= ENV["MP_DATABASE_HOST"] || "127.0.0.1" %>
+  username: <%= ENV["MP_DATABASE_USERNAME"] || "mp" %>
+  password: <%= ENV["MP_DATABASE_PASSWORD"] || "mp" %>
 
 
   # The specified database role being used to connect to postgres.
@@ -62,9 +62,9 @@ development:
 test:
   <<: *default
   database: mp_test
-  host: <%= ENV["MP_DATABASE_HOST"] %>
-  username: <%= ENV["MP_DATABASE_USERNAME"] %>
-  password: <%= ENV["MP_DATABASE_PASSWORD"] %>
+  host: <%= ENV["MP_DATABASE_HOST"] || "127.0.0.1" %>
+  username: <%= ENV["MP_DATABASE_USERNAME"] || "mp" %>
+  password: <%= ENV["MP_DATABASE_PASSWORD"] || "mp" %>
 
 # As with config/secrets.yml, you never want to store sensitive information,
 # like your database password, in your source code. If your source code is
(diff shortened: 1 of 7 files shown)


kmarszalek (approved):

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+  db:
+    image: postgres:14.1
+    ports:
+      - "5432:5432"
+    networks:
+      - mp-local
+    environment:
+      POSTGRES_DB: ${DB_POSTGRES_DB:-mp}
+      POSTGRES_USER: ${DB_POSTGRES_USER:-mp}
+      POSTGRES_PASSWORD: ${DB_POSTGRES_PASSWORD:-mp}
```
michal-szostak, line 13: It would be nice to define persistent volume for development to prevent DB from being destroyed when calling `docker-compose down`.

```
# in db section
    volumes:
      - db-data:/var/lib/postgresql/data

# at the end of the file
volumes:
  db-data:      
```

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+  el:
+    image: elasticsearch:7.5.2
+    ports:
+      - "9200:9200"
+      - "9300:9300"
```
michal-szostak, line 18: The same as with binding postgres

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    networks:
+      - mp-local
+    environment:
+      discovery.type: single-node
+  redis:
+    image: redis:6.2.6
+    ports:
+      - "6379:6379"
```
michal-szostak, line 26: The same as with binding postgres

config/database.yml
```diff
@@ -62,9 +62,9 @@ development:
 test:
   <<: *default
   database: mp_test
-  host: <%= ENV["MP_DATABASE_HOST"] %>
-  username: <%= ENV["MP_DATABASE_USERNAME"] %>
-  password: <%= ENV["MP_DATABASE_PASSWORD"] %>
+  host: <%= ENV["MP_DATABASE_HOST"] || "localhost" %>
```
michal-szostak, line 65: I would use `127.0.0.1` here. There are known issues with using `localhost` when connecting to postgres with intention to use IPv4

config/database.yml
```diff
@@ -24,9 +24,9 @@ default: &default
 development:
   <<: *default
   database: mp_development
-  host: <%= ENV["MP_DATABASE_HOST"] %>
-  username: <%= ENV["MP_DATABASE_USERNAME"] %>
-  password: <%= ENV["MP_DATABASE_PASSWORD"] %>
+  host: <%= ENV["MP_DATABASE_HOST"] || "localhost" %>
```
michal-szostak, line 27: I would use 127.0.0.1 here. There are known issues with using localhost when connecting to postgres with intention to use IPv4

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+version: "3"
+
+services:
+  db:
+    image: postgres:14.1
+    ports:
+      - "5432:5432"
```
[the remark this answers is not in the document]
jswk: Why not bind to `0.0.0.0` for development?

config/database.yml
jswk: I haven't encountered any, but ok.

docker-compose.yml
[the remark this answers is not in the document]
michal-szostak: On linux it also opens FW for that port, I would say that this is a security issue

michal-szostak (approved):

PR #2480, opened 2022-01-03
[#2472] Fix clearing a resource form

### LINKED ISSUE #2472: The form is cleared after clicking the "Create resource" button, if user previously clicked "Preview" on an empty form
Tested:
3.28.0-beta.0

**Describe the bug**
The form is cleared after clicking the "Create resource" button, if you previously clicked "Preview" on an empty form

**To Reproduce**
1. Go to '/backoffice/services'
2. Click on the 'Preview' button
3. Fill required field in the form
4. Click on the 'Create resource' button
5. See error - resource isn't create, whole form is cleared
image

**Expected behavior**
Resource is created

goreck888:
Click on a `Preview` button when the form is empty
caused saving empty attributes in a session.
Then after filling the form and click on
"Create Resource" the priority was in session attributes
Fixes #2472

### DIFF
--- a/app/controllers/backoffice/services_controller.rb
+++ b/app/controllers/backoffice/services_controller.rb
@@ -106,6 +106,7 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
       render :preview
     else
       render error_view, status: :bad_request
+      session.delete(preview_session_key)
     end
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Adding providers with improper data to the resource (@goreck888)
+- Clear a resource form if clicked on Preview button on an empty form (@goreck888)
 
 ### Security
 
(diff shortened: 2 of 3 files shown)


app/controllers/backoffice/services_controller.rb
```diff
@@ -150,7 +151,7 @@ def perform_create
     if @service.persisted?
       update_logo_from_session!
       session.delete(preview_session_key)
-      redirect_to backoffice_service_path(@service), notice: "New service created successfully"
+      redirect_to backoffice_service_path(@service), notice: "New resource created successfully"
     else
       add_missing_nested_models
```
Marcelinna, line 156: You need make change in e2e test

jswk: Shouldn't this service=>resource thingy be addressed in #2475 anyway?

I would just revert the change in this line

jswk (commented):
Apart from my suggestion to revert the naming change this PR is fine.

Marcelinna (approved):

app/controllers/backoffice/services_controller.rb
goreck888: Ok I'll revert it, but I want to merge #2475 first.

jswk (approved):

kmarszalek (approved):

Issue #2468, opened 2021-12-22
Draft provider visual indication in Backoffice


michal-szostak:
Blocked by #2445

Design and add visual indicator for draft providers in backoffice.

PR #2491, opened 2022-01-11
[#2468] Visual representation of statuses for providers in BO

jarekzet:
added provider status label on BO providers list and provider details view according to #2468


Marcelinna:
I think you can change the color for the draft resource to gray too
image

kmarszalek (approved):
looks good. it works. - even if changelog skips some parts of the previous milestone. -

Marcelinna (approved):

Issue #2302, opened 2021-09-10
Resource with no order type set, is a different type than offer


Marcelinna:
Tested 3.20.0 beta
**Describe the bug**
When creating a new resource, the order type field is not completed, the default resource is Other order type, but resource access is Open Access.

**To Reproduce**
1. Go to 'backoffice'
2. Click on Owned Resources
3. Click on Create new resource
4. Fill form, field Order type leave empty
5. Click on Create Resource
6. See error - Resource is Other order type, but offer is Open Access

![image](https://user-images.githubusercontent.com/59872096/132842603-37ff129e-3024-440d-8534-b598caba24f0.png)
![image](https://user-images.githubusercontent.com/59872096/132842846-0ab2c23b-9aa1-4095-9518-360ca4b6e649.png)

Marcelinna:
done in https://github.com/cyfronet-fid/marketplace/pull/2507

PR #2500, opened 2022-01-13
[#2322] Fix various types for service with one offer

### LINKED ISSUE #2322: Resource with one offer is Various order type
**Describe the bug**
Tested on v3.22.0-beta.0"

**To Reproduce**
1. Go to the resource where you have administrator rights - Order type: Open Access
2. Delete all offers (via API)
3. Click the button Manage resource --> Set parameters and offers
image
4. Click the button "Add new offer"
5. Fill the form, field order type is editable, change to Order required, and save
6. See error -> resource is Various type although has one offer (resource is open access, offer order required)
image
7. Update resource in PC
8. See error - resource has two offers
image

Expected result:
offer is of the same type as resource

**similar problem**
1. Go to Resource with order type Open Access
2. Create second offer Order required 
3. Resource is Various Order type
image
4. Click "Edit" for every offer and open the link in the new tab
5. In one tab click "Delete offer"
6. In the second tab change the offer order type to Full Open Access and Update offer
7. See error - resource is various order type with one offer
image

goreck888:
Fixes posibility of setting different order types
for offer and resource if that resource has
only one offer
Fixes #2322

### DIFF
--- a/app/models/offer.rb
+++ b/app/models/offer.rb
@@ -57,6 +57,14 @@ class Offer < ApplicationRecord
   validate :proper_oms?, if: -> { primary_oms.present? }
   validates :oms_params, absence: true, if: -> { current_oms.blank? }
   validate :check_oms_params, if: -> { current_oms.present? }
+  validate :same_order_type_as_in_service,
+           if: -> {
+             service&.order_type&.present? &&
+               (
+                 (new_record? && service.offers.published.size.zero?) ||
+                   (persisted? && service.offers.published.size == 1)
+               )
+           }
 
   def current_oms
     primary_oms || OMS.find_by(default: true)
@@ -97,6 +105,12 @@ class Offer < ApplicationRecord
     missing_keys = Set.new(current_oms.custom_params.keys) - Set.new(oms_params.keys)
     unless (missing_keys & Set.new(current_oms.mandatory_defaults.keys)).empty?
       errors.add(:oms_params, "missing mandatory keys")
+    end
+  end
+
+  def same_order_type_as_in_service
+    unless order_type == service.order_type
+      errors.add(:order_type, "must be the same as in the resource: #{service.order_type}")
     end
   end
 
--- a/app/views/backoffice/services/offers/_one_offer_edit_form.html.haml
+++ b/app/views/backoffice/services/offers/_one_offer_edit_form.html.haml
@@ -4,6 +4,7 @@
   .col-lg-8.pl-0{ "data-controller": offer.order_required? ? "ordering" : "" }
     = f.error_notification
     = f.hidden_field :id
+    = f.hidden_field :order_type, value: service.order_type
     - if offer.order_required?
       = f.input :internal,
         wrapper_html: { "data-target": "ordering.internalWrapper" },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 
 ### Fixed
 - Redirection after Scientific Domain deletion in backoffice (@kmarszalek)
+- Consistence of Resource/Offer order types (@goreck888)
 
 ### Security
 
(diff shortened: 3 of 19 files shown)


Marcelinna:
1. Create new resource, fill only required fields? Why offer doesnt created?
image

Marcelinna:
1. Go to PC
2. Create resource with order type: open access
3. Update resource in PC, change order type to full open access
In MP we have resource  Various order types  instead Full open Access
image

app/models/offer.rb
```diff
@@ -100,6 +104,12 @@ def oms_params_match?
     end
   end
 
+  def same_order_type_as_in_service
+    unless order_type == service.order_type
+      errors.add(:order_type, "must be the same as in the resource: #{service.order_type}")
+    end
+  end
+
```
Marcelinna, line 112: see this bug too https://github.com/cyfronet-fid/marketplace/issues/2302

Marcelinna:
If we don't fill field order_type in backoffce, whether we set to field by default?

app/models/offer.rb
```diff
@@ -57,6 +57,14 @@ def should_index?
   validate :proper_oms?, if: -> { primary_oms.present? }
   validates :oms_params, absence: true, if: -> { current_oms.blank? }
   validate :check_oms_params, if: -> { current_oms.present? }
+  validate :same_order_type_as_in_service,
+           if: -> {
+             service&.order_type&.present? &&
+               (
+                 (service&.offers&.published&.size&.zero? && new_record?) ||
+                   (service&.offers&.published&.size == 1 && persisted?)
+               )
```
jswk, line 66: Most of the safe navigation operators are redundant here, this should suffice (existence of `service` is checked in the first line, `.present?` doesn't need the safe navigation):
```ruby
             service&.order_type.present? &&
               (
                 (new_record? && service.offers.published.size.zero?) ||
                   (persisted? && service.offers.published.size == 1)
               )
```
I would also put the record state check before offer counts. Seems easier to read.

app/models/service.rb
```diff
@@ -39,7 +39,7 @@ class Service < ApplicationRecord
 
   enum status: STATUSES
 
-  has_many :offers, dependent: :restrict_with_error
+  has_many :offers, dependent: :restrict_with_error, validate: false
```
jswk, line 42: I don't think it's needed. Setting `validate: false` here only disables validation for offers added to the relation.
The default of true should work ok.

app/views/backoffice/services/offers/_one_offer_edit_form.html.haml
```diff
@@ -4,6 +4,7 @@
   .col-lg-8.pl-0{ "data-controller": offer.order_required? ? "ordering" : "" }
     = f.error_notification
     = f.hidden_field :id
+    = f.hidden_field :order_type, value: service.order_type
```
jswk, line 7: Why it's necessary here? Wouldn't a single offer already have a correct order_type set when editing it?

jswk (commented):
It seems to work ok, I have only given a few minor correction ideas.
Also, please clarify the reason for the hidden field and disabling validation in relation.

Marcelinna:
1. Go to Backoffice
2. Create resource order type: Full open access
3. Add second offer: open access
4. Add offer in two separate tab
5. Delete open access offer
6. For the second offer change open-access type to order requirder
7. Click button update
8. See problem - after first update user see view for order required defauilt offer
image
9. Click update - offer is update to full open access - ok

app/models/service.rb
goreck888: Done.

app/views/backoffice/services/offers/_one_offer_edit_form.html.haml
goreck888: I think it's necessary. See @Marcelinna comment below, that's the reason of this addition.

app/models/offer.rb
goreck888: Done.

jswk (approved):

jswk (approved):

Marcelinna (approved):

Issue #2495, opened 2022-01-12
BadRequest at the /recommendations endpoint


Michal-Kolomanski:
It is a duplicate of #2226
Currently, on the production server, we get a lot of logs containing `BadRequest` at the `/recommendations` endpoint.

An example query body that triggered this error:
```
{
"page_id": "/service", 
"panel_id": "disabled", 
"search_data": {
"categories": [
12, 
40, 
62, 
65, 
79, 
87, 
95, 
100, 
110, 
112
], 
"providers": 2
}, 
"timestamp": "2022-01-13T12:02:43.057+0000", 
"unique_id": "some-id", 
"visit_id": None
}
```

Error message:
```
ValidationError
2 is not of type 'array'

Failed validating 'type' in schema['properties']['search_data']['properties']['providers']:
    {'example': [113, 140, 70],
     'items': {'title': 'Provider', 'type': 'integer'},
     'type': 'array'}

On instance['search_data']['providers']:
    2
```

Solution:
Even if there is a single value in the array - a list have to be returned

Issue #2487, opened 2022-01-10
No possibility to start the v2 experiment for recommendations


Marcelinna:
**Describe the bug**
Tested on v3.29.0-beta.0

**To Reproduce**
1. Go to 'https://beta.marketplace.eosc-portal.eu/admin'
2. Click on 'AB experiments
3. Select experiment v2 and click the button "Use this"
4. See error - go to Marketplace /services and notice that recommendations aren't visible
5. Go to Ab experiments again /admin/ab_tests and click the button "Reopen experiment"
6. Select v1 and click the button "Use this"
7. Go to Marketplace /services and notice that recommendations are visible

**Expected behavior**
the recommendation are visible when user use v2 experiment

Marcelinna:
close because A/B test doesn't be used anymore. Exp v1 has been selected.

PR #2501, opened 2022-01-14
[#2494] Fix InvalidRecommendationPanelIDError

### LINKED ISSUE #2494: InvalidRecommendationPanelIDError
Currently, on the production server, we get a lot of logs containing InvalidRecommendationPanelIDError.
The scope of this issue is to investigate the situation carefully and to fix the error.

Michal-Kolomanski:
Fixes #2494


Marcelinna:
If we delete A/B test, bug "No possibility to start the v2 experiment for recommendations #2487" will be invalid and it should be close?

Michal-Kolomanski:
> If we delete A/B test, bug "No possibility to start the v2 experiment for recommendations #2487" will be invalid and it should be close?

Yes. We are abandoning the v2 version entirely.

Michal-Kolomanski:
I removed certain files, so please verify that there is no regression overall in MP @Marcelinna. Please remember that there is [second bug](https://github.com/cyfronet-fid/marketplace/issues/2495) so recommendations will not be always visible. Recommendations should be visible when using "Categories" or more than one "Scientific domains". I have tested this PR locally and recommendations were visible for me.

jswk:
It looks good @Michal-Kolomanski.
Since we're scrapping the AB testing for now, then it would make sense to drop the line `gem "split", require: "split/dashboard"` from the Gemfile too. (You may see remaining places where it's used that way either, since execution will break on these.)

Michal-Kolomanski:
> It looks good @Michal-Kolomanski. Since we're scrapping the AB testing for now, then it would make sense to drop the line `gem "split", require: "split/dashboard"` from the Gemfile too. (You may see remaining places where it's used that way either, since execution will break on these.)

Thank you for noticing. My new commit addresses your request.

jswk (commented):
One more leftover I noticed: grep the project for the strings `ab_` and `AB `. There's yet several instances in the code, which don't lead to immediate failures, but could be pruned.
Sorry I didn't think of this earlier.

jswk (approved):
Looks good now. I don't see any more leftovers.

Marcelinna:
we don't need a changelog for remove the a/b experiment from backoffice?

Marcelinna (approved):

Michal-Kolomanski:
> we don't need a changelog for remove the a/b experiment from backoffice? something has changed for a user with admin privileges.

You are right. I have added those changes in the changelog.

Issue #1977, opened 2021-04-09
required fields in PC are not required in backoffice


Marcelinna:
**Describe the bug**
there is no consistency, required fields in PC are not required in Backoffice, when creating provider and resource. 
below field which should be required in MP too. 

Create Provider
**Basic: Abbrevation, Website, Legal entity**
**Marketing: Description, Logo**
**Location: Street Name and Number, Postal Code, City, Country**
**Contact: Main Contact -> First Name, Last Name, Email. Public Contact -> Email** _(in Marketplace Main Contact -> First Name, Last Name, Email they are required for each data set, but there can be 0 sets, in PC are required always, we change so that it works the same as in PC? @bwilk @roksanaer )_
**Admins: First name, Last name, Email** -_(in Marketplace they are required for each data set, but there can be 0 sets, in PC are required always, we change so that it works the same as in PC? @bwilk @roksanaer )_

Create Resource/Service
**Basic: Resource Organisation, Webpage**
**Marketing: Logo**
**Classification: Categories, Target user**
**Availability:Language availability**
**Contact: Main Contact -> First Name, Last Name, Email. Public Contact -> Email, Helpdesk email, Security contact email**-
_(First Name, Last Name, Email (in Marketplace they are required for each data set, but there can be 0 sets, in PC are required always, we change so that it works the same as in PC? @bwilk @roksanaer)_
**Maturity: Technology Readiness Level**
**Order: Order Type**

the classification in the PC is divided, in MP not
which makes it possible to add the service without subdomain and subcategory in the MP, it should be standardize ? @bwilk @roksanaer ?

MP (create resource) 
Classification
![image](https://user-images.githubusercontent.com/59872096/114346291-81ff6c80-9b63-11eb-9466-08b93740cdca.png)
![image](https://user-images.githubusercontent.com/59872096/117653158-5be5e000-b194-11eb-86c5-819cff124b4e.png)

PC
Classification
![image](https://user-images.githubusercontent.com/59872096/114346319-8deb2e80-9b63-11eb-8061-c023466c233e.png)
![image](https://user-images.githubusercontent.com/59872096/114346337-95aad300-9b63-11eb-9bb6-73811cdd4baa.png)

MP (create provider)
Classification
![image](https://user-images.githubusercontent.com/59872096/114346897-71032b00-9b64-11eb-99f8-af871c57da55.png)

PC
Classification
[...]

PR #2507, opened 2022-01-19
[#2505] Fix adding parameters to the default offer

### LINKED ISSUE #2505: [FIX] adding parameters to default offer when order type isn't filled
Tested:
v3.30.0-beta.0

**Describe the bug**
It isn't possible to add parameters to the offer, when for service order type isn't filled

**To Reproduce**
1. Go to 'backoffice/services/new'
2. Fill required fields, don't fill field order type
3. Click the button "Create resources"
4. Click the button "Edit parameters"
5. Try to add parameters and click the button "Update offer"
6. See error - it isn't possible to add parameters
image

**Expected behavior**
It is possible to add parameters

goreck888:
Problem caused by hidden field which is blank
when no order_type is chosen in a service.
Add default "other" order_type in this case
Fixes #2505 #2302

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -136,6 +136,7 @@ class Service < ApplicationRecord
   validates :access_policies_url, mp_url: true, if: :access_policies_url?
   validates :sla_url, mp_url: true, if: :sla_url?
   validates :webpage_url, mp_url: true, if: :webpage_url?
+  validates :order_type, presence: true
   validates :status_monitoring_url, mp_url: true, if: :status_monitoring_url?
   validates :maintenance_url, mp_url: true, if: :maintenance_url?
   validates :order_url, mp_url: true, if: :order_url?
--- a/app/services/service/create.rb
+++ b/app/services/service/create.rb
@@ -11,7 +11,7 @@ class Service::Create
       Offer.new(
         name: "Offer",
         description: "#{@service.name} Offer",
-        order_type: @service.order_type || "open_access",
+        order_type: @service.order_type,
         order_url: @service.order_url,
         internal: @service.order_url.blank?,
         status: "published",
--- a/app/services/service/update.rb
+++ b/app/services/service/update.rb
@@ -12,21 +12,18 @@ class Service::Update
     else
       @service.update(@params)
     end
+    order_type = @params[:order_type].presence || @service.order_type.presence
     if @service.offers.published.size == 1
       Offer::Update.new(
         @service.offers.first,
-        {
-          order_type: @params[:order_type] || @service.order_type,
-          order_url: @params[:order_url] || @service.order_url,
-          status: "published"
-        }
+        { order_type: order_type, order_url: @params[:order_url] || @service.order_url, status: "published" }
       ).call
     elsif @service.offers.published.empty?
       Offer::Create.new(
         Offer.new(
           name: "Offer",
           description: "#{@params[:name] || @service.name} Offer",
-          order_type: @params[:order_type] || @service.order_type,
+          order_type: order_type,
           order_url: @params[:order_url] || @service.order_url,
           status: "published",
           service: @service
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_16_143044) do
+ActiveRecord::Schema.define(version: 2022_01_20_112130) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -179,7 +179,7 @@ ActiveRecord::Schema.define(version: 2021_11_16_143044) do
     t.string "author_uid"
     t.index ["author_id"], name: "index_messages_on_author_id"
     t.index ["author_role"], name: "index_messages_on_author_role"
-    t.index ["messageable_type", "messageable_id"], name: "index_messages_on_messageable_type_and_messageable_id"
+    t.index ["messageable_type", "messageable_id"], name: "index_messages_on_messageable"
     t.index ["scope"], name: "index_messages_on_scope"
   end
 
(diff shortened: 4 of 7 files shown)


app/services/service/create.rb
```diff
@@ -11,7 +11,7 @@ def call
       Offer.new(
         name: "Offer",
         description: "#{@service.name} Offer",
-        order_type: @service.order_type || "open_access",
+        order_type: @service.order_type || "other",
         order_url: @service.order_url,
         internal: @service.order_url.blank?,
         status: "published",
```
Marcelinna, line 17: Why change order type from open access to other help?

1. I create a new service, I didn't set "order type"
2. I see on the label that service is  "Other order type" and offer is "other" type
image
3. I delete the offer - service doesnt have any offer
4. I create offer offer open access - on label I see Other order type for service, offer is open access
5. I add parameters - parameters are added

I have the same situation in bug descripton, but there adding parametres don't working.

jswk: The suggestion to make `order_type` a required field seems to make sense to me, although for some reason it wasn't marked required before. Do you know why such case was allowed? I cannot think of anything in particular, but maybe whitelabel-related stuff?

Nevertheless, I don't see any services in staging (apart from the ones created for testing) that have `order_type == nil`. Resource Profile marks this field as mandatory, so it shouldn't be a problem in the main MP instance.

-----

About falling back to `"other"` here, I would rather make the `ServiceHelper::order_type` default to `open_access` in that case. It wouldn't modify the DB population process.
Also, historically we defaulted to `"open_access"` and `"external"` there, not to `"other"` (e.g. https://github.com/cyfronet-fid/marketplace/blob/2602338d7492493b5a14cc3ebc9b53de6d657a53/app/helpers/service_helper.rb#L70 and https://github.com/cyfronet-fid/marketplace/blob/2602338d7492493b5a14cc3ebc9b53de6d657a53/app/helpers/service_helper.rb#L81).

Marcelinna: I don't know why in past order_required filed wasn't required but according to https://github.com/cyfronet-fid/marketplace/issues/1977 we should set in BO required field as are in PC. In PC order type is required, so I think we can make it required in this PR too.

goreck888: So let's return to the `open_access` default matching and show unmarked services as an `open_access` resource

Marcelinna (approved):

app/services/service/create.rb
```diff
@@ -11,7 +11,7 @@ def call
       Offer.new(
         name: "Offer",
         description: "#{@service.name} Offer",
-        order_type: @service.order_type || "open_access",
+        order_type: @service.order_type || "other",
```
jswk, line 14: As per https://github.com/cyfronet-fid/marketplace/pull/2507#discussion_r787928143 let's not change the current handling. Will you add a default on the column, the part `|| "open_access"` should be dropped.

app/services/service/update.rb
```diff
@@ -12,21 +12,18 @@ def call
...
-        }
+        { order_type: order_type, order_url: @params[:order_url] || @service.order_url, status: "published" }
       ).call
     elsif @service.offers.published.empty?
       Offer::Create.new(
         Offer.new(
           name: "Offer",
           description: "#{@params[:name] || @service.name} Offer",
-          order_type: @params[:order_type] || @service.order_type,
+          order_type: order_type,
```
jswk, line 26: I don't think these are necessary when field order_type will be required in the form and have a default in DB.

app/views/backoffice/services/offers/_one_offer_edit_form.html.haml
```diff
@@ -4,7 +4,7 @@
   .col-lg-8.pl-0{ "data-controller": offer.order_required? ? "ordering" : "" }
     = f.error_notification
     = f.hidden_field :id
-    = f.hidden_field :order_type, value: service.order_type
+    = f.hidden_field :order_type, value: service.order_type || "other"
```
jswk, line 7: Same here.

app/models/service.rb
```diff
@@ -136,6 +136,7 @@ class Service < ApplicationRecord
   validates :access_policies_url, mp_url: true, if: :access_policies_url?
   validates :sla_url, mp_url: true, if: :sla_url?
   validates :webpage_url, mp_url: true, if: :webpage_url?
+  validates :order_type, presence: true
```
jswk, line 139: I would also add a DB constraint: `null: false, default: "open_access"` to the column, when we're at it.
I would add the default so that even if we had such records in the DB we don't have errors.

You may also add a data migration to fill the missing order_type, and only then impose the constraint.

jswk (changes requested):

jswk:
The two failing tests in `spec/features/backoffice/services_spec.rb` need correction too, since this PR introduces a change in validation.

db/schema.rb
```diff
@@ -268,6 +268,29 @@
...
+    t.bigint "service_id", null: false
+    t.bigint "user_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.integer "issue_id"
+    t.integer "issue_status", default: 2, null: false
+    t.index ["service_id"], name: "index_orders_on_service_id"
+    t.index ["user_id"], name: "index_orders_on_user_id"
+  end
+
```
jswk: These tables were removed some time ago. Please remove them.
Don't you have some local migrations which are not checked-in to the repository that generate those records? It doesn't happen to others.

jswk (changes requested):
Why not combine both migrations? I don't see why we need two separate ones.
```ruby
class SetOrderTypeAsRequiredInService < ActiveRecord::Migration[6.1]
  def change
    exec_update "UPDATE services SET order_type = 'open_access' WHERE order_type = '' OR order_type IS NULL"
    change_column :services, :order_type, :string, null: false
  end
end
```

Actually, I'd also remove the default value, since it won't be necessary. I know I suggested it, but it only makes sense in case we wouldn't do a data migration. Now it's a no-op.

jswk (approved):

db/schema.rb
goreck888: I had loaded anonymized production db.

Issue #2510, opened 2022-01-24
Jira integration doesn't trim length of some fields


jswk:
Jira has character limit of 255 on some fields. We have length validation on some, but not all affected.
According to Jira docs the limit applies to URL and short text (in our instance I actually see `Text Field (single line)` type there) fields (https://support.atlassian.com/jira-cloud-administration/docs/custom-fields-types-in-company-managed-projects/).

There is also length validation for project name, but since something is appended to it before sending to jira then it's still possible to trigger this bug.

For now, trim all the fields sent to Jira. Introduce an allowlist of fields that we know can handle longer values and don't trim those.

In the long run, when the Jira integration is dropped this should be handled on the adapter level (since this is the Jira limitation, and not inherent to ordering).

The Jira field type that is unaffected is `Text Field (multi-line)`.

Jira epic types:

Summary | System field |   |  
-- | -- | -- | --
Description | System field |   |  
Assignee | System field |   |  
Epic Color | Colour of Epic |   |  
Epic Name | Name of Epic |   |  
CI-DisplayName | Text Field (single line) |   |  
CI-Name | Text Field (single line) |   |  
CI-Surname | Text Field (single line) |   |  
CI-Email | Text Field (single line) |   |  
CI-EOSC-UniqueID | Text Field (single line) |   |  
CI-LinkedinProfile | URL Field |   |  
CI-Institution | Text Field (single line) |   |  
CI-Country | Text Field (single line) |   |  
CI-Department | Text Field (single line) |   |  
CI-DepartmentalWebPage | URL Field |   |  
CI-ResearchGateProfile | URL Field |   |  
CI-SupervisorName | Text Field (single line) |   |  
CI-SupervisorProfile | URL Field |   |  
CP-CustomerTypology | Select List (single choice) |   |  
SO-ProjectName | Text Field (single line) |   |  
CP-UserGroupName | Text Field (single line) |   |  
CP-CollaborationCountry | Text Field (multi-line) |   |  
CP-CustomerCountry | Text Field (multi-line) |   |  
CP-ScientificDiscipline | Text Field (single line) |   |  
CP-ProjectInformation | Text Field (multi-line) |   |  
CP-ReasonForAccess | Text Field (multi-line) |   |  
EOSC-hub TEST | Select List (single choice) |   |  
CP-Nationality | Text Field (multi-line) |   |  
Involved | User Picker (multiple users)

Jira issue types:

Field | Type |   |  
-- | -- | -- | --
[...]

PR #2511, opened 2022-01-24
[#2510] Truncate some too long Jira fields

jswk:
It was possible to create a Jira issue prototype, which won't be
accepted by Jira (either one with a limit-long name, or with too many
scientific disciplines).

Truncate the fields that have a type, which doesn't allow more
characters.

Fixes: #2510.


Marcelinna:
the error was about the length of characters in all fields?
Project name *
Reason to request access to the EOSC resources *
Additional information
Email *
Organization *
Department
Origin country *
Webpage *
User group name *
Collaboration countries *
Company name *
Company website url *
Project name *
Project website url *

jswk:
> the error was about the length of characters in all fields?

I identified that it was possible to trigger the error for Collaboration countries and for SO-ProjectName.
For others not really since we have model validation in place. Nevertheless, I feel that truncating all fields doesn't hurt.

kmarszalek (approved):

Marcelinna:
It is ok that in JIRA I see this: ?
image
but in Project in MP I selected this:
image

jswk:
> It is ok that in JIRA I see this: ? [truncated value] but in Project in MP I selected this: [full value in MP]

I gather that yes. It's necessary to shorten the value to 255 chars. We could also limit the number of selected domains, but we have to account for a fact, that we have a project in production having excessive scientific domains length anyway. I don't think such broad categorization serves any purpose during ordering.

Marcelinna:
> > It is ok that in JIRA I see this: ? [truncated value] but in Project in MP I selected this: [full value in MP]
> 
> I gather that yes. It's necessary to shorten the value to 255 chars. We could also limit the number of selected domains, but we have to account for a fact, that we have a project in production having excessive scientific domains length anyway. I don't think such broad categorization serves any purpose during ordering.

I understand, for example, If I edit a project name, the displays of the scientific domain in JIRA will change, but I understand it's not a problem.

Marcelinna (approved):

PR #2512, opened 2022-01-25
[#1459] Fix validation of parameters

### LINKED ISSUE #1459: String is treated as Integer by JSON::Validator
**Describe the bug**
This is very strange `JSON::Validator` behavior, which causes a very strange error for attribute

**To Reproduce**

1. Define the following attribute for one of the service offer:
```json
{
"id":"id1",
"type":"select",
"label":"Number of CPU Cores",
"config":{"mode":"buttons", "values":["1", "2", "4", "8"]},
"value_type":"string",
"description":"Select number of cores you want"
}
```
2. Try to order the service with this offer

**Expected behavior**
"1" is also a string it should not be treated as an Integer.

**Additional context**

This error is connected with the following strange `JSON::Validator` behavior:

```ruby
JSON::Validator.validate!({:type=>"string"}, "1")
```

result:

```
JSON::Schema::ValidationError: The property '#/' of type integer did not match the following type: string
```

Credits of bug discovering go to @martaswiatkowska :smile:

goreck888:
Update choices.js to 10.0.0
Fix problem with casting numerical strings
Fix escaping html in parameter values

Fixes #1459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Please view this file on the master branch, on stable branches it's out of date.
 
+## [3.31.0-milestone]
+
+### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+- Improper casting numerical string to a number in offer's parameters (@goreck888)
+
+### Security
 
 ## [3.30.0-milestone]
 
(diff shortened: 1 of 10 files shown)


jswk (commented):
I'd add a test case for a string integer in `spec/models/parameter/select_spec.rb`.

Also, is doing the unescaping a security risk? Where do the unescaped values come from? What is the use case for this unescaping?

Marcelinna (approved):

jswk (approved):

PR #2528, opened 2022-02-02
[#2525] Show manage button on Resource and Provider views

### LINKED ISSUE #2525: Resource, Provider with MP upstream shouldn't have "Manage the resource", "Manage the Provider" button
Tested
v3.30.0-beta.1

**Describe the bug**
Resource, Provider with MP upstream shouldn't have "Manage the resource", "Manage the Provider" button which redirects to PC

image
image
image

kmarszalek:
Show button 'Manage the..' o Resource and Provider views only when
upstream is set to 'eosc_registry'

Fixes #2525

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Strip leading and trailing whitespaces from Resource and Provider name (@kmarszalek)
 
 ### Changed
+- Show 'Manage the provider/resource' button only when upstream set to 'eosc_registry' (@kmarszalek)
 
 ### Deprecated
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -53,4 +53,8 @@ module ApplicationHelper
   def placeholder(variant, text = "Placeholder - link to about", link = about_path)
     render "layouts/placeholder", text: text, link: link, variant: variant
   end
+
+  def show_manage_button?(record)
+    policy(record).data_administrator? && record.upstream&.eosc_registry?
+  end
 end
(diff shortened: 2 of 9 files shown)


app/helpers/backoffice/services_helper.rb
```diff
@@ -36,4 +36,8 @@ def preview_link_parameters(is_preview)
       {}
     end
   end
+
+  def show_manage_button?(service)
+    policy(service).data_administrator? && service.upstream&.eosc_registry?
+  end
```
goreck888, line 42: Maybe use only one `show_manage_button?` method? The code is identical.

app/helpers/provider_helper.rb
```diff
@@ -29,4 +29,8 @@ def field_tree(service, field)
     parents = service.send(field).map { |f| f.parent.blank? ? f : f.parent }
     parents.map { |parent| [parent.name, (parent.children & service.send(field)).map(&:name)] }.to_h
   end
+
+  def show_manage_button?(provider)
+    policy(provider).data_administrator? && provider.upstream&.eosc_registry?
```
goreck888, line 34: as above

goreck888 (approved):
Generally code looks ok, consider removeing duplicated method.

app/helpers/backoffice/services_helper.rb
kmarszalek: done

app/helpers/provider_helper.rb
kmarszalek: done

Marcelinna:
If we change upstream to MP we should lost access to ordering_configuration#offers page ??

image
image

Marcelinna (approved):

Issue #2338, opened 2021-10-12
Unexpected error occurred while searching in the main search


Marcelinna:
Tested Release 3.22.1

**Describe the bug**
Typing an additional keyword in the main search while being on a page other than 1, an unexpected error occurs

**To Reproduce**
1. Go to 'https://beta.marketplace.eosc-portal.eu/services'
2. Type in the main search keyword e.g "data"
![image](https://user-images.githubusercontent.com/59872096/136902820-a7e9ab32-9b30-463b-8cb7-533fb106d2f0.png)
2. On pagination change page to 2
![image](https://user-images.githubusercontent.com/59872096/136900669-9a913fb8-ee9e-43fe-acad-1308e2f6058a.png)
3. Type in the main search additional word e.g "data" and click enter (the word must stick to the previous one)
![image](https://user-images.githubusercontent.com/59872096/136902614-a305ed01-a3e5-438c-a978-038a0570d122.png)
![image](https://user-images.githubusercontent.com/59872096/136902929-5485a376-3cf3-4bc8-8810-c152f5707eb4.png)

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/59872096/136902988-c9850805-bf69-499f-b419-d9b9463d5080.png)

**Additional context**

PR #2538, opened 2022-02-14
[#2338] Fix researching resources from nth page

danielkryska:


Marcelinna:
I reproduce steps from the linked issue, and I see that problem still exist
image

Marcelinna:
changelog?

jswk (changes requested):
This PR unfortunately doesn't exactly fix #2338, the issue there is rather the behaviour in the backend.

In case the passed page is out of range, then there are several options (all in the backend, not front):
1. return 400 (instead of 500)
2. redirect to an url without the page param
3. show first page (without redirect or modifying the URL)
4. show last page (without redirect or modifying the URL)

Personally, I feel 2. is the way to go. The behaviour will be explicit this way and the user will end up on the first page, which seems logical to me.
What do you think @Marcelinna? (EDIT: We should follow 3. instead. It's how `:page` behaves for negative or non-integer values, so we should keep it consistent.)

(Also please conform to the commit message schema and add appropriate changelog while you're at it.)

Marcelinna:
we should follow with the solution from 3 (show first page)

danielkryska:
Option 2 and 3 are the same (, the missing param is equal to page 1 search). Fixes have been provided with CHANGELOG

jswk (changes requested):
We discussed with @Marcelinna and going for 3. is a more unified way. The option 2. is behaving inconsitently. I.e., if `page=9999` then you get redirected, if `page=-1` you don't.
If you strongly want to do the redirect then do it on other invalid inputs too, e.g. `page=asdf`, `page=.5`, `page=-1`, but probably not for `page=123.12` (as it will be parsed to 123).

For both cases, don't reset the filters when handling the exception. Now the filters are cleared in that case (to see invalid behaviour select a category and then trigger the exception).

jswk:
Also, commit message.

danielkryska:
I got it. Done.

jswk (changes requested):
Works great now, thanks!

Just add the same handling code to `app/controllers/backoffice/services_controller.rb` too, as it still crashes. Sorry I missed it earlier.

jswk (changes requested):
Great that you added the analogous code to the backoffice!

Nevertheless, please revert the changes from #2539 that made their way to this PR.

jswk (approved):

Marcelinna:
https://marketplace-2.docker-fid.grid.cyf-kr.edu.pl/services/c/aggregators-integrators?page=2000
https://marketplace-2.docker-fid.grid.cyf-kr.edu.pl/services?scientific_domains[]=1&page=3000
https://marketplace-2.docker-fid.grid.cyf-kr.edu.pl/services?page=2000
image

jswk:
> we handle page=300, but at page=3000 an error has already occurred. Why?

That's why:
```
[cd160567-46f5-458d-83f6-f4dbab0a2b73] Searchkick::InvalidQueryError ([400] {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"services_production_20220215101552394","node":"qV2clyRNROeFKWrP4tbt0A","reason":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}],"caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.","caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}},"status":400}):
```
In particular, Elastic complains:
> Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.

It's a bit different situation that the original problem though. We may limit the page param based on the number of entries in our DB, so this error is avoided until we hit these 10000 entries limit.
But I'm not sure if this is to be handled in this PR.

jswk:
Moreover, this won't manifest in normal browsing situation, as opposed to the #2338. So probably 500 is the right response for this (at least for now).

Marcelinna:
> > we handle page=300, but at page=3000 an error has already occurred. Why?
> 
> That's why:
> 
> ```
> [cd160567-46f5-458d-83f6-f4dbab0a2b73] Searchkick::InvalidQueryError ([400] {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"services_production_20220215101552394","node":"qV2clyRNROeFKWrP4tbt0A","reason":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}],"caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.","caused_by":{"type":"illegal_argument_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}},"status":400}):
> ```
> 
> In particular, Elastic complains:
> 
> > Result window is too large, from + size must be less than or equal to: [10000] but was [200000]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.
> 
> It's a bit different situation that the original problem though. We may limit the page param based on the number of entries in our DB, so this error is avoided until we hit these 10000 entries limit. But I'm not sure if this is to be handled in this PR.

I think this pr doesn't have to solve this

kmarszalek (approved):

Issue #1809, opened 2021-02-22
Remove code duplication in provider and service view with component


kmarszalek:
https://github.com/cyfronet-fid/marketplace/pull/1794
Needed changes after this PR with a review from @wujuu, more details in PR
Generally, it's ok, but there is a lot of duplicated code in the views (provider views <-> service views). It's understandable since these views are awfully similar, but we could generalize them if it wouldn't be too confusing (maybe ViewComponent can help)? But that's for another issue.

michal-szostak:
Is this issue still relevant?

PR #1794, opened 2021-02-05
[#1768] Providers view

kmarszalek:
* [x] view -> basic
* [x] breadcrumb
* [x] router
* [x] link from list
* [x] model extension
* [x] importers
* [x] slug -> breadcrumb
* [x] full view
* [x] link from resources list
* [x] providers section on home page
* [x] css styles for map-component

Closes #1768

PR #2540, opened 2022-02-14
 [#1809] Unify service/provider views by the ViewComponent 

goreck888:
Add new `view_components` for displaying unified
header, description, sidebar and details.
Unify method names
Commit with implementation of header_component,
header_links_component
Fixes https://github.com/cyfronet-fid/marketplace/issues/1809


app/helpers/service_helper.rb
```diff
@@ -114,14 +105,6 @@ def highlighted_for(field, model, highlights)
...
 
-  def service_logo(service, classes = "align-self-center mr-4 float-left img-responsive", resize = "100x67")
-    if service.logo.attached? && service.logo.variable?
-      image_tag service.logo.variant(resize: resize), class: classes
-    else
-      image_pack_tag("eosc-img.png", size: resize, class: classes)
-    end
-  end
-
   def data_for_map(geographical_availabilities)
```
jswk, line 108: Do we still need this helper? Wouldn't the helper `data_for_map_provider`, maybe renamed to `data_for_map`, be enough?
They are duplicated 1-to-1.
See also: https://github.com/cyfronet-fid/marketplace/pull/1794#discussion_r579148705.

jswk: Similar situation with the two helper methods below, i.e. `data_for_region` and `any_non_european?`.
Probably both should be moved to presentable_helper? The `any_non_european?` may be made private.

app/helpers/application_helper.rb
```diff
@@ -57,4 +57,12 @@ def placeholder(variant, text = "Placeholder - link to about", link = about_path
...
     policy(record).data_administrator? && record.upstream&.eosc_registry?
   end
+
+  def object_logo(object, classes = "align-self-center mr-4 float-left img-responsive", resize = "100x67")
+    if object.logo.attached? && object.logo.variable?
+      image_tag object.logo.variant(resize: resize), class: classes
+    else
+      image_pack_tag("eosc-img.png", size: resize, class: classes)
+    end
+  end
```
jswk, line 67: I feel like this one would better fit in `app/helpers/presentable_helper.rb`.
Also, isn't it actually a `presentable_logo`?

app/helpers/application_helper.rb
```diff
@@ -57,4 +57,12 @@ def placeholder(variant, text = "Placeholder - link to about", link = about_path
   def show_manage_button?(record)
     policy(record).data_administrator? && record.upstream&.eosc_registry?
   end
```
jswk, line 59: I feel like this one would better fit in `app/helpers/presentable/header_helper.rb`.
I know it's not your changes it was added in, but anyways, it's only used in the header_component.

Maybe even better, it could be incorporated into the `/app/components/presentable/header_component.rb` directly not to pollute the helpers namespace?

app/helpers/presentable_helper.rb
```diff
@@ -21,12 +21,12 @@ def data_for_region_provider(countries)
     countries
   end
 
-  def provider_any_present?(record, *fields)
+  def any_present?(record, *fields)
     fields.map { |f| record.send(f) }.any?(&:present?)
```
jswk, line 25: A short-circuiting implementation maybe:
```ruby
fields.any? { |f| record.send(f).present? }
```
It won't call `send` for every field before checking if any is present but return early.

jswk (changes requested):
First of all, good job on refactoring all these.

I've only several suggestions in comments, they are mostly some improvement on code deduplication and nothing major.

There's also these few unaddressed comments from #1794 which should be easy to incorporate:
- https://github.com/cyfronet-fid/marketplace/pull/1794#discussion_r579142621
- https://github.com/cyfronet-fid/marketplace/pull/1794#discussion_r579145217
- https://github.com/cyfronet-fid/marketplace/pull/1794#discussion_r579178639

app/helpers/service_helper.rb
goreck888: Done.

app/helpers/application_helper.rb
goreck888: Done.

app/helpers/application_helper.rb
goreck888: I moved it to the `HeaderHelper`, because this method is not used directly in the `HeaderComponent`, but in partials

app/helpers/presentable_helper.rb
goreck888: Done.

Marcelinna:
beta 
image
pr
image

jswk (approved):
From my side this PR looks good now.

As it's massive, I would suggest to only add the `[#1809]` tag to all commits' messages and __not squash__ prior to merging. This will make the changes more readable in the repo's history.

kmarszalek (approved):

Marcelinna:
please check the "Updates" display

beta
image
pr
image

Marcelinna:
pr
image
beta
image


We don't have "Add to favorite", "Add to comparison" button on the beta instance in BO
In the back office, these buttons do not work, I don't know if we should leave them?

Marcelinna (approved):

Issue #2381, opened 2021-11-04
Add to favourites does not work on all sub-pages of the service


Marcelinna:
Tested on v3.24.0-beta.0

**Describe the bug**
Add to favourites does not work on sub-pages: /services/{service.name}/details, /services/{service.name}/opinions

**To Reproduce**
1. Go to '/services'
2. Select one of the services (e.g 100 Percent IT Trusted Cloud)
3. Click the `Details` button 
4. Click `Add to favorites button `
5. See problem - service doesn't add to favourites 
6. Click The Review button 
4. Click `Add to favorites button `
5. See problem - service doesn't add to favourites 

**Expected behavior**
Service should be added to favourites

**Screenshots**
![image](https://user-images.githubusercontent.com/59872096/140271498-2afa406f-e1fb-4e49-bee6-e9fe47a1d00c.png)
![image](https://user-images.githubusercontent.com/59872096/140271520-6dc2040e-222f-410b-9f67-09f3d3b2c363.png)

PR #2542, opened 2022-02-15
[#2381] Fix resource subpages add to favourite button

danielkryska:


Marcelinna:
beta
image
pr
image

I don't see a popup when I add the first resource to favorite when I am on the service presentation page

jswk (changes requested):
The additions in details and opinions controllers are ok and making the checkbox work.

However, moving the JS controller root made the popup not to work. See how in the master instance you get a popup when you add your first favourite in a service view. The popup doesn't show up in the PR though and also you get an error in the console:
image

jswk (changes requested):
I'm still getting the exception when trying to favourite a service:
image
Please correct this.

jswk (changes requested):
Still doesn't work (see https://github.com/cyfronet-fid/marketplace/pull/2542#pullrequestreview-882798364).

danielkryska:
Now it should work fine

jswk (changes requested):
Good idea to move the modal closer to checkbox! The modal also won't be unnecessarily added to DOM when favourite checkbox isn't displayed, which is nice.

Just correct the changelog, since the fix will go to version 3.33.0 instead of 3.32.0, and the PR is good to go.

jswk:
Also, please move the changelog entry from the merged #2538 to the correct section while you're at it.

jswk (approved):

Marcelinna (approved):

PR #2548, opened 2022-02-17
[#2549] Fix resources statistics for executive

### LINKED ISSUE #2549: "Statistics - Total orders" are visible for Order required external resources which are not orderable in MP
Tested:
3.32.0-beta.0

**Describe the bug**
"Statistics - Total orders" are visible for Order required external resources which are not orderable in MP, they only pinned to the project

Actual result

Statistic for open access, fully open access, other 
image

Statistic for order required internal
image

Statistic for order required external
image

Expected result:
Please delete "Total orders" from order required external resources because there are pinned to the project

goreck888:
Add `orderable` scope for `offers` and `project_items`
for proper displaying and counting ordered (only internal `order_required`) resources.

Fixes #2549

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Please view this file on the master branch, on stable branches it's out of date.
 
-
 ## [3.33.0-milestone]
 
 ### Added
@@ -27,6 +26,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix resource subpages add to favourite button (@danielkryska)
 - Fix researching resources from nth page of resources (@danielkryska)
 - Don't depend on offer when determining whether a Project Item is a bundle (@jswk)
+- Fix displaying statistics of the resource for executive role (@goreck888)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


kmarszalek:
quickfix to the master?

Marcelinna:
Please add changelog

jswk:
Quickfix or a hotfix. If a hotfix why not merge to the release branch then?

Also, why is this change needed? The ordered count was reported based on order_type only before and there's no connected ticket saying it's incorrect.

BTW, maybe add an `alt` attribute to the label "Total orders", so that it's more clear what is that number. For example:
- "A number of created Project Items that use EOSC Portal as the order management platform" (if we change the current logic)
- "A number of created order required Project Items" (if we keep the current logic)

Marcelinna:
Personally, I would not change the name "Total orders".
We are introducing a change because it should apply only to the ordered resources, and we can say that something has been ordered in MP only if it is order required internal type

jswk:
> Personally, I would not change the name "Total orders".

I don't suggest to change the name though.

> We are introducing a change because it should apply only to the ordered resources, and we can say that something has been ordered in MP only if it is order required internal type.

Okay, this makes sense. But it was included neither in the commit message nor in the PR description.

Marcelinna (approved):

jswk (approved):

Issue #2331, opened 2021-10-05
Redundant comma and Languages in Resource details (resource create via BO)


Marcelinna:
Tested beta v3.22.1"
**Describe the bug**
When creating a new resource on the form, complete the Language availability field. There is an unnecessary comma in the Details

**To Reproduce**
Steps to reproduce:
1. Go to '/backoffice/services/new'
2. Fill in required fields Availability   - >  Language availability
![image](https://user-images.githubusercontent.com/59872096/136006321-d87273fa-589a-44f7-8757-b20858afd8a0.png)
3. Click button Create a resource
4. Go to Resource Details, see the problem - redundant ","
![image](https://user-images.githubusercontent.com/59872096/136006573-4a759781-df10-4e0a-974e-518a7aec110e.png)

Steps to reproduce:
1. Go to '/backoffice/services/new'
2. Don't fill in fields Availability   - >  Language availability
3. Click button Create a resource
4. Go to Resource Details, see the problem - field language availability isn't filled but "Languages" is visible on details
![image](https://user-images.githubusercontent.com/59872096/140272941-eb6bfad5-c32b-40e6-a131-4113b92261aa.png)

PR #2551, opened 2022-02-18
[#2331] Remove redundant comma in service details languages

danielkryska:


Marcelinna:
Please resolve second problem from isuue too
image

Marcelinna:
> Please resolve second problem from isuue too image

**Go to Resource Details, see the problem - field language availability isn't filled but "Languages" is visible on details**

Languages isn't required field, if you don't fill it, should be visible in details

jswk:
> Go to Resource Details, see the problem - field language availability isn't filled but "Languages" is visible on details

I believe the whole issue arose from the fact that when editing with the multiselect input, the first element will always be `""` (empty string).
Examples is DB:
```
irb(main):008:0> Service.all.filter {|s| s.languages[0] == "" }.map {|s| [s.name, s.languages] }
=> 
[["Science Software on Demand", ["", "English"]],
 ["test", [""]],
 ["OpenAIRE Research Graph", [""]],
 ["EnviDat", ["", "English"]],
 ["test", [""]],
 ["O3as: ozone assessment", ["", "English"]],
 ["EGI Cloud Compute", ["", "English"]]]
```

Then when you render service `"test"` for example, then the field is rendered since it contains an element.
The same goes for the comma, `["", "English"].join(", ")` will give `, English`.

I'd say the way to tackle this is to make the multiselect input not create this empty string at the beginning. It would normalize the DB and make the problems outlined in the #2331 go away (it would also possibly make sense to verify we don't have such entries in the production and execute a corrective action there once this PR is deployed there).

jswk:
Last but not least, verify no other array columns are affected by this bug. If they are correct their handling either.

danielkryska:
Languages are fixed

db/migrate/20220218104738_remove_empty_strings_from_service_languages.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class RemoveEmptyStringsFromServiceLanguages < ActiveRecord::Migration[6.1]
+  def change
+    Service
+      .all
+      .select { |service| service.language_availability.present? }
+      .each do |service|
+        service.language_availability = service.language_availability.reject(&:blank?)
+        service.save
+      end
```
jswk, line 11: Don't add references to code in migration, rather correct this using plain SQL.
This could lead to failures in future, when we'll want to refactor some code used here.

Alternatively, add this command to the upgrade doc to run directly in console.

jswk (changes requested):
Apart from the migration which needs to be corrected, the current fix works well.

db/migrate/20220218104738_remove_empty_strings_from_service_languages.rb
danielkryska: Done

danielkryska:
`Simple form` input param `include_hidden: false` doesn't work with an empty array.

danielkryska:
Looking for a solution I found excellent lib for our purposes:
https://github.com/holli/auto_strip_attributes

Marcelinna (approved):

app/models/service.rb
```diff
@@ -40,6 +40,9 @@ class Service < ApplicationRecord
 
   enum status: STATUSES
 
+  extend AutoStripAttributes
```
jswk, line 43: This line isn't needed since all ActiveRecords will extend it anyway.

jswk (changes requested):
Please remove the unnecesary extend and it's good to go.

jswk (approved):

Issue #2543, opened 2022-02-16
API for bundles


michal-szostak:
Start implemetation with assumption that any services can be bundled together, setup meeting to discuss restrictions / validators.

jswk:
Access control is based on the parent, as one can create links to foreign offers.
Block creation of multi-level bundles in validator.

Assumption is to use the `#create/update` method to link offers with the field exposed as `bundled_offers` (empty in case offer is not a bundle).

The `bundled_offers` is unordered, behaves like a set.

Offer's id schema is `<service.slug>/<offer.iid>`, following the convention from URL paths and other Offering API endpoints.

## Example flows:

### Create a new offer with bundled offers

Precondition: offers with `id1` and `id2` exist.

Then one just POSTs a new offer with `bundled_offers: [id1, id2]`.

### Add offer to bundle

Precondition: offers with `id1` and `id2` exist and are bundled. Also offer `id3` exists.

Then one just PATCHes the bundle offer with `bundled_offers: [id1, id2, id3]`.

PR #2561, opened 2022-02-22
[#2543] API for bundles

jswk:
## Add bundles to Offering API

Extend the Offering API maintaining backward compatibility.

The bundled_offers are identified in API by "<service.slug>/<offer.iid>"
to match their API identification.

## Add actions on bundle-relevant service and offer changes

In particular:
- unbundle offers that were made non-public and send notification to the
  remaining bundle offer then
- notify provider if her offer was bundled in another provider's bundle

Don't enforce rubocop's Lint/AmbiguousBlockAssociation in spec, as such
a syntax is a preferred way to express
`expect {...}.not_to change {...}` in rspec.

# How to test

Make sure to see whether [the assumptions from the issue](https://github.com/cyfronet-fid/marketplace/issues/2543#issuecomment-1043075327) hold both when using API and when triggering the emails from UI (as some mails are to be sent when an offer is unbundled due to, for example, being deleted/drafted).

Closes: #2543.


Marcelinna:
image
image

we don't have any error messages? which would indicate that the bundle value is invalid

jswk:
> we don't have any error messages? which would indicate that the bundle value is invalid

@Marcelinna: this is how the JSON validation works in Offering API. It could be improved, but it's more like a topic for another issue.

Marcelinna:
image
Link [offer's management](https://marketplace-13.docker-fid.grid.cyf-kr.edu.pl/backoffice/services/2) redirects100 Percent IT Trusted Cloud me to EGI High-Throughput Compute service. I created a bundle for 100 Percent IT Trusted Cloud.

jswk:
> Link [offer's management](https://marketplace-13.docker-fid.grid.cyf-kr.edu.pl/backoffice/services/2) redirects me to EGI High-Throughput Compute service. I created a bundle for 100 Percent IT Trusted Cloud.

A bug when generating the link, will correct.

> Where do you want to redirect the user? page with API?

I will redirect the user to the "API docs" page, instead of the BO. BO doesn't make sense since we don't have bundle handling in BO UI.
```
Go to the (Offering API)[api_docs_url] to verify your bundle offer.
```

Marcelinna:
1. Order bundle offer
image
2. Delete offer which is linked to bundle offer. 
3. See problem - an offer which was linked to bundle offer was removed from bundle view in /project
image

Please give me information that should be resolved in another ticket?

Marcelinna:
image

I change the linked offer to open access, offer is still public, I think the message isn't correct.

Marcelinna:
1. Create bundle offer 100 Percent IT Trusted Cloud > main offer with Software Repository>linked offer
image
2. Change Software Repository>linked offer to open access 
3. Go to 100 Percent IT Trusted Cloud click Access to resource - I cant order resources, service in BO has too offer visible but in MP I cant select any of them
image

Marcelinna:
1. Create bundle offer
2. Delete linked offer
3. Bundle owner doesn't achieve mail notification that linked offer is not public anymore

app/controllers/api/v1/resources/offers_controller.rb
```diff
@@ -60,24 +71,50 @@ def find_and_authorize
...
     schema_file = action_name == "create" ? "offer_write.json" : "offer_update.json"
-    JSON::Validator.validate!(Rails.root.join("swagger", "v1", "offer", schema_file).to_s, params[:offer].as_json)
+    JSON::Validator.validate!(Rails.root.join("swagger", "v1", "offer", schema_file).to_s, request.body.read)
   rescue JSON::Schema::ValidationError => e
     render json: { error: e.message }, status: 400
   end
 
+  def load_bundled_offers
+    bundled_offers = params[:bundled_offers]
+    unless bundled_offers.nil?
```
danielkryska, line 86: Wouldn't be better to use `if bundled_offers.present?`?

app/controllers/api/v1/resources/offers_controller.rb
```diff
@@ -60,24 +71,50 @@ def find_and_authorize
...
+      if bundled_offers.is_a?(Array) && bundled_offers.all? { |o| o.is_a?(String) }
+        @bundled_offers = bundled_offers.map { |val| Offer.find_by_slug_iid!(val) }
+      else
+        render json: { error: "Bundled offers must be an array of strings" }, status: 400
+      end
```
danielkryska, line 91: Wouldn't be better to write the code as such?

```
bundled_offers_valid = bundled_offers.is_a?(Array) && bundled_offers.all?(&:is_a?(String))
render json: { error: "Bundled offers must be an array of strings" }, status: 400 unless bundled_offers_valid
@bundled_offers = bundled_offers.map { |val| Offer.find_by_slug_iid!(val) }

File: app/services/service/update.rb [hunk 3]
```diff
@@ -28,7 +31,42 @@ def call
...
+    unbundle_and_notify! if public_before && !@service.public?
+  end
+
+  def notify_bundled_offers!
+    @service
+      .offers
+      .published
+      .filter(&:bundle?)
+      .each do |published_bundle|
+        published_bundle.bundled_offers.each do |bundled_offer|
```
danielkryska, line 51: 3rd repetition of code block?

danielkryska (approved):

danielkryska:
@jswk, @kmarszalek I'd be glad to participate, more often, in reviews. Since Rubocop set up (from around the end of last year) I've learned a lot about Rails syntax and I'm ready to be at least part of the code quality check. 

Short story long: Don't be afraid to add me into reviews in your PR's

.prettierignore
```diff
@@ -36,5 +36,7 @@
 # Ignore styles
 /app/javascript/stylesheets/*
 
-# Ignore autogenerated schema
+# Ignore autogenerated files
 /db/schema.rb
+/swagger/v1/offering_swagger.json
+/swagger/v1/ordering_swagger.json
```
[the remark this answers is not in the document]
jswk: Other files aren't autogenerated.

app/controllers/api/v1/resources/offers_controller.rb
jswk: It would change the semantics. In case `bundled_offers = []` we need to remove all bundled_offers, not ignore them.

app/controllers/api/v1/resources/offers_controller.rb
jswk: But you need to return the result of `render` from the method, so it would have to look rather like this:
```ruby
bundled_offers_valid = bundled_offers.is_a?(Array) && bundled_offers.all?(&:is_a?(String))
return render json: { error: "Bundled offers must be an array of strings" }, status: 400 unless bundled_offers_valid
@bundled_offers = bundled_offers.map { |val| Offer.find_by_slug_iid!(val) }
```

I don't like the convolution of this though. An `if` is more clear on having only two code paths too.

app/controllers/api/v1/resources/offers_controller.rb
```diff
@@ -32,20 +33,30 @@ def create
...
   private
 
+  def permitted_attributes(record)
+    super(record)
+  rescue ActionController::ParameterMissing => e
+    if @bundled_offers.is_a?(Array)
+      {}
+    else
+      raise e
+    end
```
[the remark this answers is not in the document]
jswk: I prefer the expanded `if`, without an unnecessary `unless` operator.

app/services/offer/create.rb
```diff
@@ -1,13 +1,20 @@
...
+    notify_added_bundled_offers! if @offer.save
     @offer.service.reindex
     @offer
   end
+
+  private
+
+  def notify_added_bundled_offers!
+    @offer.added_bundled_offers&.each { |added_bundled_offer| Offer::Mailer::Bundled.call(added_bundled_offer, @offer) }
+  end
```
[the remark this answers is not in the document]
jswk: Prettier styles check pass on this PR.

app/services/offer/update.rb
```diff
@@ -1,14 +1,36 @@
...
+class Offer::Update < ApplicationService
   def initialize(offer, params)
+    super()
     @offer = offer
     @params = params
   end
 
   def call
-    @offer.service.offers.published.size == 1 ? @offer.update(@params) : @offer.update(@params.merge(default: false))
+    effective_params = @offer.service.offers.published.size == 1 ? @params : @params.merge(default: false)
```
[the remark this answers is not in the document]
jswk: image

app/services/service/publish.rb
```diff
@@ -1,14 +1,30 @@
...
+
+  private
+
+  def notify_bundled_offers!
+    @service
+      .offers
+      .published
+      .filter(&:bundle?)
+      .each do |published_bundle|
+        published_bundle.bundled_offers.each do |bundled_offer|
```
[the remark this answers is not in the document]
jswk: I understand there's some repetition, but the contexts of both of the calls are different a little and may need adjusting on a per-case basis. I wouldn't like to introduce too much abstraction here.

Once we strike 3 repetitions the according to the Ro3 we may refactor this, extract this to a separate concern maybe with separate tests.

app/services/service/update.rb
jswk: Where are the other two repetitions?

Marcelinna:
1. I make a bundle from 4 offers
image
2. I go to the final step of ordering
image

In the summary, we show only bundled offer which has parameters?

Marcelinna (approved):

kmarszalek (approved):

PR #2563, opened 2022-02-22
[HOTFIX] Fix EGI Notebooks backoffice page

goreck888:
Add missing direct path to the `notebook_openaire_explore`
partial. Now it's impossible to see EGI Notebooks page
in the backoffice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 - Fix researching resources from nth page of resources (@danielkryska)
 - Don't depend on offer when determining whether a Project Item is a bundle (@jswk)
 - Fix displaying statistics of the resource for executive role (@goreck888)
+- Missing partial in the EGI Notebooks resource page (@goreck888)
 
 ### Security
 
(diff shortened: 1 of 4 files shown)


Marcelinna:
When I try to visit EGI Notebook in BO
image

Marcelinna:
When I visit EGI Notebook on BO
image

kmarszalek (changes requested):
bad PR instance for this PR, please check this out. some test?

Marcelinna:
Is there any other service with which we can have a similar problem?

jswk (changes requested):
Please add a check for this in the `features/service_spec.rb`.

This will be enough:
image
and we'll have this case covered.

**PG:** Not really, the problem was in the backoffice, so I just add similar test in the backoffice/ordering_configuration context

jswk:
> Is there any other service with which we can have a similar problem?

@Marcelinna: This error is specific to the "EGI Notebook" only, as only it has this openaire integration.

Marcelinna (approved):

jswk (approved):

kmarszalek (changes requested):
we change the scope of this from hotfix to normal bug, change the base to master.sorry.

kmarszalek (approved):

Issue #2544, opened 2022-02-16
Upgrade to rails 7.0


michal-szostak:
!!! BLOCKED ISSUE !!!

There are 2 libraries that don't support Rails 7.0:

- simple_token_authentication
- activestorage-validator

which blocks us from upgrading to the newest version of Rails.

EDIT: @kmarszalek 
- simple_token_authentication -> its on PR, waiting for merge + version
https://github.com/gonzalo-bulnes/simple_token_authentication/issues/385 
https://github.com/gonzalo-bulnes/simple_token_authentication/pull/391
- activestorage-validator -> its in master, waiting for version
https://github.com/aki77/activestorage-validator/commit/0e2f348fc89d36cb3f6a0d50a313f0d6e2aa0f91

danielkryska:
From a conversation with @mkasztelnik 
- most of the libraries are forward compatibile

Possible improvements for Rails 7.0:
- using `css-bundling` and `js-bundling` instead webpacker (https://esbuild.github.io/ can improve JS and CSS building speed something like x1000)
- replacement of `turbolinks` to `turbo`
- use [Hotwire](https://hotwired.dev/) to have SPA in Rails (required `turbo`)
-

mkasztelnik:
The migration path I'm performing right now in MEE is as follow:
  * dependencies upgrade in rails 6.1.x and used libraries
  * replace webpacker with `js-bundling`, `css-bundling`
  * replace `turbolinks` with `turbo`
  * upgrade to rails 7

kmarszalek:
- simple_token_authentication -> its on PR, waiting for merge + version
https://github.com/gonzalo-bulnes/simple_token_authentication/issues/385
https://github.com/gonzalo-bulnes/simple_token_authentication/pull/391
- activestorage-validator -> its in master, waiting for version
https://github.com/aki77/activestorage-validator/commit/0e2f348fc89d36cb3f6a0d50a313f0d6e2aa0f91

gonzalo-bulnes:
Hey @kmarszalek, Simple Token Authentication **v1.18.0** should get you one step closer to closing this issue 🙂

agpul:
invalid

PR #2564, opened 2022-02-23
[#2544] Rails 7.0 upgrade

danielkryska:


danielkryska:
Enums possible solution: 
https://stackoverflow.com/questions/36762779/ruby-on-rails-issues-using-enum-with-same-name

danielkryska:
There are 2 libraries that don't support Rails 7.0:

- simple_token_authentication
- activestorage-validator

which blocks us from upgrading to the newest version of Rails.

PR #2593, opened 2022-03-13
[#2558] Improve multicheckbox filters

### LINKED ISSUE #2558: Improve filter selection for scientific domain
After consideration about the operation of filters for scientific domains with @jarekzet , we came to the conclusion to improve the performance of the filters. (similar to how it is work on other websites e.g www.x-kom.pl)

Assumptions. 

- After selecting a parent, all his children are selected
image

- After unselecting a parent, all their children are unselected
image

- After unselecting one child
image

- After unselecting all children, parent is unselected
image

- After selecting all children, severally
image

- After selecting not all children
image

goreck888:
Add indeterminate state for multilevel checkboxes
Closes #2558

### DIFF
--- a/app/javascript/app/controllers/filter_controller.js
+++ b/app/javascript/app/controllers/filter_controller.js
@@ -10,12 +10,16 @@ export default class extends Controller {
 
   toggle(event) {
     event.preventDefault();
-    this.toggleParent(event);
-    this.toggleChildren(event);
-    this.reload();
+    const target = event.currentTarget;
+    const collapsed = !target.classList.contains("collapsed");
+
+    document.cookie = `${target.id}=${collapsed};expires=${this.expireAtString()}`;
   }
 
-  reload() {
+  reload(event) {
+    event.preventDefault();
+    this.toggleParent(event);
+    this.toggleChildren(event);
     let form = this.formTarget;
     for (const element of form) {
       if (
@@ -54,13 +58,11 @@ export default class extends Controller {
       return;
     }
     if (anyChildHasDifferentState) {
-      console.log("INDETERMINATE PARENT");
       parent.checked = false;
       parent.indeterminate = true;
       parent.classList.add("indeterminate");
       return;
     }
-    console.log("PARENT CHECK/UNCHECK");
     parent.classList.remove("indeterminate");
     parent.checked = state;
   }
--- a/app/javascript/app/controllers/filter_controller.js
+++ b/app/javascript/app/controllers/filter_controller.js
@@ -9,13 +9,13 @@ export default class extends Controller {
   }
 
   toggle(event) {
-    const target = event.currentTarget;
-    const collapsed = !target.classList.contains("collapsed");
-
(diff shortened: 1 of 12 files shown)


app/javascript/app/controllers/filter_controller.js
```diff
@@ -25,8 +24,49 @@ export default class extends Controller {
...
+    if (parent) {
+      if (anyChildHasDifferentState) {
+        console.log("INDETERMINATE PARENT");
+        parent.checked = false;
+        parent.indeterminate = true;
+      } else {
+        console.log("PARENT CHECK/UNCHECK");
+        parent.checked = state;
+      }
+      return;
```
danielkryska, line 68: Why do you call return in here?

app/javascript/app/controllers/filter_controller.js
```diff
@@ -25,8 +24,49 @@ export default class extends Controller {
...
+
+    if (parent) {
+      if (anyChildHasDifferentState) {
+        console.log("INDETERMINATE PARENT");
+        parent.checked = false;
+        parent.indeterminate = true;
+      } else {
+        console.log("PARENT CHECK/UNCHECK");
+        parent.checked = state;
+      }
```
danielkryska, line 67: In case of any problems with the settings state of checkboxes I suggest using one/both of the beneath methods to prevent events bubbling:

https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation
https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault

Marcelinna:
Backoffice filers don't work

Marcelinna:
- [ ] QA
- [ ] DEV

If the user clicks on scientific domain picture on https://marketplace-4.docker-fid.grid.cyf-kr.edu.pl/, the user should be redirected to /services, scientific domain with every child should be selected.  
image
image

Marcelinna:
Step to reproduce:
1. Go to Marketplace 
2. Select Parent Humanities - every child is selected
image
3. Unselect Parent Humanities - every child is unselected
image
4. Select Parent Humanities again - every child is selected
5. Delete from the selection under Active filter - Humanities - behaviour isn't the same, paren is unselected but every child is seleted
image
image

Marcelinna:
> * [ ]  QA
> 
>  * [x]  DEV
> 
> 
> Backoffice filers don't work

It isn't possible check elements without value. 
image

Marcelinna:
Steps to reproduce:

1. Go to .services
2. Click on the Natural Sciences filter
image
3. Unselect two children - Earth & Related Environmental Sciences, Biological Sciences using button under "Actvie filter" label
4. See error - despite filters being deleted under "Active filter" label, they are still selected inside filter
image

Marcelinna:
Steps to reproduce
1. Go to Backoffice and create new platform 
image
3. Go to Markteplace and find new platform in filters
image
5. Check filter - see error
image

kmarszalek (approved):

PR #2598, opened 2022-03-17
[#2537] Integrate user_actions with JMS

### LINKED ISSUE #2537: Wrap as a component the direct way of publishing user actions
We do not resign from the direct way of publishing user actions to the recommender.
1) This method should be packaged into a component,
2) An environment variable should be created that will allow defining how we want to publish user actions (direct, databus or both),
3) Possibility of connecting to any queue (databus).

wujuu:
closes #2537

### DIFF
--- a/config/stomp_publisher.yml
+++ b/config/stomp_publisher.yml
@@ -5,6 +5,7 @@ default: &default
   host: <%= ENV["MP_STOMP_HOST"] || Rails.application.credentials.stomp.dig(:host) || "''" %>
   ssl-enabled: <%= ENV["MP_STOMP_SSL"] || false %>
   topic: <%= ENV["MP_STOMP_PUBLISHER_TOPIC"] || Rails.application.credentials.stomp.dig(:publisher, :topic) || "''" %>
+  databus-topic: <%= ENV["MP_STOMP_PUBLISHER_DATABUS_TOPIC"] || "databus" %>
 
 test:
   <<: *default
--- a/README.md
+++ b/README.md
@@ -230,6 +230,9 @@ We are currently using the following ENV variables:
   * `MP_STOMP_PUBLISHER_ENABLED` (Optional) - turn on publishing with JMS if set to `true` (default `false`)
   * `MP_STOMP_PUBLISHER_LOGGER_PATH` (Optional) - path to the JMS log file (default `"RAILS_ROOT/log/jms.publisher.log"`)
   * `MP_STOMP_PUBLISHER_TOPIC` (Optional) - topic of a new message to be published (default `credentials.stomp.publisher.topic`)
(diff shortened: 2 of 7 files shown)


config/stomp_publisher.yml
```diff
@@ -5,6 +5,7 @@ default: &default
   host: <%= ENV["MP_STOMP_HOST"] || Rails.application.credentials.stomp.dig(:host) || "''" %>
   ssl-enabled: <%= ENV["MP_STOMP_SSL"] || false %>
   topic: <%= ENV["MP_STOMP_PUBLISHER_TOPIC"] || Rails.application.credentials.stomp.dig(:publisher, :topic) || "''" %>
+  recommender-topic: <%= ENV["MP_STOMP_PUBLISHER_RECOMMENDER_TOPIC"] || "recommender" %>
```
Michal-Kolomanski, line 8: - I wonder if "recommender-topic" is the best name. I would rather use something like "user_action_topic" in that case.
- Add a description of this variable to the README

Michal-Kolomanski (changes requested):
The [scope](https://github.com/cyfronet-fid/marketplace/issues/2537) of the ticket was also to create an env variable to control user_actions destination. This variable should support 3 values. Sending user_actions:
1)  directly to the recommender,
2)  sending to databus, 
3)  both ways at the same time.

For now, it should be set to the third option. 

Furthermore, the whole mechanism should be extensively tested.

jswk (commented):
~~Other than the misleading docs the PR looks ok.~~

jswk (changes requested):
Please also add the test for the added behaviour to `spec/services/jms/publisher_spec.rb`.

wujuu:
> Please also add the test for the added behaviour to `spec/services/jms/publisher_spec.rb`.

I've added tests, but to the UserActionController, because the Jms::Publisher didn't actually change. (it was Jms::Publish helper class). Left a TODO (for another PR) to expand the tests, possibly a task for @Michal-Kolomanski

jswk:
> > Please also add the test for the added behaviour to `spec/services/jms/publisher_spec.rb`.
> 
> I've added tests, but to the UserActionController, because the Jms::Publisher didn't actually change(rather Jms::Publish helper class, which we do not test as for now - I've felt that we should test the UserActionController, because it enapsulates more of the logic added). Left a TODO (for another PR) to expand the tests, possibly a task for @Michal-Kolomanski

Right, my bad. You're right that the Jms::Publish doesn't have tests.

jswk (approved):
Seems ok now.

Michal-Kolomanski (approved):

Michal-Kolomanski:
We need to inform Wojtek about the variables we've added and what values they should have.

PR #2599, opened 2022-03-21
[#2597] Rewrite jms handle of Provider msgs

### LINKED ISSUE #2597: Rewrite logic for provider - PC integration
We should simplify logic on our side.

MP statues: draft, published, deleted
draft - when it is hidden in mp 
published - when it is shown in mp
deleted - when it is deleted in PC

PC: Registered / Approved / Rejected / Deleted/ Deactivated + ACTIVE (true/false)

example:
Registered + false -> draft
deactivated + false -> draft
rejected + false -> draft
approved + true -> published
deleted -> stays as it is -> deleted

All providers from PC should be in MP with the proper status that will show them or hide.

kmarszalek:
All providers in all statuses should be in mp db, with proper mp status for showing them in FO.

Fixes #2597

### DIFF
--- a/app/services/jms/manage_message.rb
+++ b/app/services/jms/manage_message.rb
@@ -20,7 +20,7 @@ class Jms::ManageMessage
   def call
     body = JSON.parse(@message.body)
     action = @message.headers["destination"].split(".").last
-    log body
+    log @message
     resource = @parser.parse(body["resource"])
 
     raise ResourceParseError, "Cannot parse resource" if resource.empty?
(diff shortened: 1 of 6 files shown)


Marcelinna:
- [ ] Provider in pending status, active false in PC - should be in draft status in MP
- [ ] Provider in approved status, active true in PC - should be in published status in MP
- [ ] Provider in approved status, active false in PC - should be in draft status in MP
- [ ] Provider in deleted status in PC - should be in deleted status in MP

app/services/jms/manage_message.rb
```diff
@@ -41,9 +41,10 @@ def call
       end
     when "provider"
       modified_at = modified_at(resource, "providerBundle")
-      if action == "delete"
+      case action
+      when "delete"
         Provider::DeleteJob.perform_later(resource["providerBundle"]["provider"]["id"])
-      elsif action == "update" && process_provider_update?(resource)
+      when "update" || "create"
```
jswk, line 47: Remove this no-op. I.e. the `"create"` will be optimized away anyway, see:
```
irb(main):002:0> ("update" || "create") == "update"
=> true
```

Marcelinna:
Steps to reproduce 
1. Go to beta provider and deactivate Porvider UK Atomic Energy Authority
2. Go MP - see error, only Provider is deactivated, resource is still with published status

Marcelinna (approved):

Issue #2596, opened 2022-03-15
PC's importer should utilize refresh token to authenticate


michal-szostak:
Importer should require refresh token from aai and try to obtain valid authentication token from that. Then it should call PC's API to obtain resources / providers.

The scope of this issue contains:

- [ ] Fix existing importer code if it broke after long period of being unused
- [ ] Implement authentication via refresh token

---

- [ ] check if the code responsible for REST imports is working
- [ ] fix REST import if relevant
- [ ] fetch AAI JWT token based on the .env refresh token

danielkryska:
- file `lib/tasks/import.rake`
- tasks for resources and providers
- JWT token should be fetched from AAI based on refresh token (in `Importers:Provider` request)

danielkryska:
code source: https://docs.egi.eu/providers/check-in/sp/#refresh-request

```
curl -X POST "https://aai.com" -u "client_id":"client_secret" -d "grant_type=refresh_token" -d "refresh_token=4444.5555.6666" -d "scope=openid%20email%20profile"
```

```
fetch('https://aai.com', {
    method: 'POST',
    headers: {
        'Authorization': 'Basic ' + btoa('client_id:client_secret')
    },
    body: 'grant_type=refresh_token&refresh_token=4444.5555.6666&scope=openid%20email%20profile'
});
```

PR #2601, opened 2022-03-21
[#2596] Fetch JWT token using refresh token

danielkryska:
**IMPORTANT!!!**
Today (24.03.2022) the service hasn't worked. AAI successfully delivered an access token, but the provider service throw 404.

---

For testing purposes:

go to: https://aai.eosc-portal.eu/providers-api/
image

ENV:
MP_IMPORT_EIC_URL=https://beta.providers.eosc-portal.eu/api
UPSTREAM=eic
IMPORTER_AAI_REFRESH_TOKEN=<AS AT IMAGE>
IMPORTER_AAI_BASE_URL=aai.eosc-portal.eu
IMPORTER_AAI_CLIENT_ID=<AS AT IMAGE>

### DIFF
--- a/.run/Import providers.run.xml
+++ b/.run/Import providers.run.xml
@@ -1,0 +1,37 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Import providers" type="RakeRunConfigurationType" factoryName="Rake">
+    <module name="marketplace" />
+    <RAKE_RUN_CONFIG_SETTINGS_ID NAME="RUBY_ARGS" VALUE="" />
+    <RAKE_RUN_CONFIG_SETTINGS_ID NAME="WORK DIR" VALUE="$MODULE_DIR$" />
+    <RAKE_RUN_CONFIG_SETTINGS_ID NAME="SHOULD_USE_SDK" VALUE="false" />
+    <RAKE_RUN_CONFIG_SETTINGS_ID NAME="ALTERN_SDK_NAME" VALUE="" />
+    <RAKE_RUN_CONFIG_SETTINGS_ID NAME="myPassParentEnvs" VALUE="true" />
+    <EXTENSION ID="BundlerRunConfigurationExtension" BUNDLE_MODE="AUTO" bundleExecEnabled="true" />
+    <EXTENSION ID="JRubyRunConfigurationExtension" NailgunExecEnabled="false" />
+    <EXTENSION ID="RubyCoverageRunConfigurationExtension" runner="rcov">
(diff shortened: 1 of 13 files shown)


danielkryska:
**Issue blocked!!!**
To confirm the solution is working properly, an extension of grant types (refresh_token) is needed for all (prod, beta, dev, local) clients.

.run/Import providers.run.xml
```diff
@@ -0,0 +1,41 @@
+<component name="ProjectRunConfigurationManager">
```
jswk, line 1: Should this run configuration be commited? It even contains a JWT token (although an expired one).
If the IDE suddenly generates such files, then I would add `/.run` to the root `.gitignore`.

app/services/importers/token.rb
```diff
@@ -0,0 +1,36 @@
...
+        params: {
+          scope: SCOPE,
+          grant_type: GRANT_TYPE,
+          refresh_token: @refresh_token
+        },
+        headers: {
+          Authorization: "Basic #{Base64.strict_encode64("#{CLIENT_ID}:#{CLIENT_SECRET}")}"
+        }
+      )
+    response = connection.post("/#{AAI_TOKEN_PATH}", {}.as_json)
```
jswk, line 32: Wouldn't `@faraday.post(...)`, without a separate `.new` call, be more convenient?

app/services/importers/token.rb
```diff
@@ -0,0 +1,36 @@
...
+    raise Errno::ECONNREFUSED if response.blank? || response.status != 200
```
jswk, line 33: I would rather raise your custom error here instead of a system errorno (ECONNREFUSED is also rather used in cases when the target server doesn't accept connections).

app/services/importers/token.rb
danielkryska: done

app/services/importers/token.rb
danielkryska: I agree

.run/Import providers.run.xml
danielkryska: I've updated the files here. Those configurations are in here to help other developers using ruby mine to start faster into the project. You can just with 1 click run application as also importers, of course with the previous installation of tools.

danielkryska:
**IMPORTANT!!!**
Today the service hasn't worked. AAI successfully delivered an access token, but the provider service throw 404.

Marcelinna (approved):

jswk (approved):

PR #2610, opened 2022-03-25
[#2590] Fix REST import of Providers and Resources

### LINKED ISSUE #2590: Option to update single service from PC via task (importer)
In recent weeks there has been increased number of requests from EOSC onboarding to synchronize specific services which for some reason do not appear in MP. It would be useful to extend existing importer task to be able to specify single service which would be imported. This way this type of manual addition of the services would be less error prone.

from
Extend importer to fetch create / update single resource #2594
Add option for importer to accept argument RESOURCE_ID=... if this argument is provided only specified resource would be fetched and created / updated. This feature would allow us to quickly fix issues reported to us by onboarding team.

EDIT:
take into account #2595 status isn't there in previous api

kmarszalek:
*change API endpoint to newer that consist of status
and active flag in response
*adapt code to diff return value

USE:
* import resource/resources
bundle exec rake import:resources IDS=<RESOURCE_IDS> MP_IMPORT_TOKEN=<TOKEN>
* import provider/providers
bundle exec rake import:providers IDS=<RESOURCE_IDS> MP_IMPORT_TOKEN=<TOKEN>
IDS => comma separated PC ids
TOKEN => token with epot role.

Fixes #2590

### DIFF
--- a/lib/import/providers.rb
+++ b/lib/import/providers.rb
@@ -29,9 +29,13 @@ class Import::Providers
 
   def call
     log "Importing providers from EOSC Registry..."
-    @request_providers = external_providers_data.select { |id, _| @ids.empty? || @ids.include?(id) }
-    @request_providers.each do |eid, external_provider_data|
+    @request_providers = external_providers_data.select { |pro| @ids.empty? || @ids.include?(pro["provider"]["id"]) }
+
+    @request_providers.each do |external_data|
+      external_provider_data = external_data["provider"]
+      eid = external_provider_data["id"]
       parsed_provider_data = Importers::Provider.new(external_provider_data, Time.now.to_i, "rest").call
+      parsed_provider_data["status"] = external_data["active"] ? :published : :draft
       eosc_registry_provider =
         Provider.joins(:sources).find_by("provider_sources.source_type": "eosc_registry", "provider_sources.eid": eid)
       current_provider = eosc_registry_provider || Provider.find_by(name: parsed_provider_data[:name])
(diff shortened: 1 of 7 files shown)


jswk (commented):
I would consider extracting refactoring changes into a separate commit.

Also, I miss an example of the data that's to be parsed by the importer. It would be useful to have a basic test which would include such example, especially since the new version of the API is not public (as I understand it requires an EPOT role).

jswk:
The new API sometimes returns responses with the metadata=null, see example: [metadata-null.json](https://github.com/cyfronet-fid/marketplace/files/8370151/metadata-null.json.log).
In that case currently the resource importer stops with:
```
rails aborted!
NoMethodError: undefined method `[]' for nil:NilClass
/home/savi/src/marketplace/lib/import/resources.rb:50:in `block in call'
/home/savi/src/marketplace/lib/import/resources.rb:46:in `each'
/home/savi/src/marketplace/lib/import/resources.rb:46:in `call'
/home/savi/src/marketplace/lib/tasks/import.rake:17:in `block (2 levels) in <top (required)>'
/home/savi/src/marketplace/bin/rails:5:in `require'
/home/savi/src/marketplace/bin/rails:5:in `<top (required)>'
/home/savi/src/marketplace/bin/spring:13:in `block in <top (required)>'
/home/savi/src/marketplace/bin/spring:10:in `<top (required)>'
```

I guess we should handle such cases by logging them and proceed with updating the rest of the resources.

lib/import/providers.rb
```diff
@@ -29,9 +29,13 @@ def initialize(
 
   def call
     log "Importing providers from EOSC Registry..."
-    @request_providers = external_providers_data.select { |id, _| @ids.empty? || @ids.include?(id) }
-    @request_providers.each do |eid, external_provider_data|
+    @request_providers = external_providers_data.select { |pro| @ids.empty? || @ids.include?(pro["provider"]["id"]) }
+
+    @request_providers.each do |external_data|
```
danielkryska, line 34: Could you use `provider_data` instead `pro` and `external_data` or `external_provider_data`?

danielkryska (approved):

jswk (approved):

Marcelinna:
Why some resource has no offers? It is someting in logs?
image

Marcelinna (approved):

Issue #2621, opened 2022-04-04
Publish events from the users/services tables to the databus


Michal-Kolomanski:
[Specyfication](https://docs.google.com/document/d/1v5v9AuntpMXdJ58RYNpquQ5FI_0EoTMKZIu5nOiar5k/edit#heading=h.jbvt6xz5urxa) done by @JanKapala.
Split into more tickets if necessary.

PR #2649, opened 2022-04-22
[#2621] Publish model changes to JMS databus

wujuu:
Cetrain models (for now, only the ones that are being used in the Recommender training) now publish to JMS Databus queue after create, update and delete. Based on [this](https://docs.google.com/document/d/1v5v9AuntpMXdJ58RYNpquQ5FI_0EoTMKZIu5nOiar5k/edit#heading=h.jbvt6xz5urxa) spec.

So now we have 2 recomennder-related JMS topics (you can customize the names using env variables, see README):
- user_actions
- mp_db_events

NOTE: Not all model changes are now being published to Databus, but this can be easily expanded, just `include Publishable` in your model and write a serializer for the model under the `Recommender` namespace

### DIFF
--- a/lib/recommender/serialize_db.rb
+++ b/lib/recommender/serialize_db.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module Recommender
+module RecommenderLib
   class SerializeDb
     def initialize; end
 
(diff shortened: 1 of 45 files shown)


lib/recommender/serialize_db.rb
```diff
@@ -1,23 +1,38 @@
 # frozen_string_literal: true
 
-module Recommender
+module RecommenderLib
```
wujuu, line 3: Changed, to not conflict with the Recommender module namespace. Before, it caused the `Vocabulary` being treated as a `Recommender::Vocabulary` module instead of the Active Record `Vocabulary` model.

Michal-Kolomanski (changes requested):
1) Unfortunately, it seems that there is an error in this PR.
[Stack trace](https://pastebin.com/nj40PKKB)
2) Please add annotation on which ticket this PR is closing,
3) After discussing with @m.szostak it was decided to send DB events to another topic (different from user_actions).

Testing + checking the code will be done in the next iteration.

Michal-Kolomanski (commented):
It's challenging to test this feature in all scenarios, but I did what I could locally, and it worked flawlessly. Excellent work! :clap: 

However, we should have better integration tests for this feature (many services will rely on this), and that's the only thing stopping me from giving it my full approval. I'm talking about integrating with any queue, making a modification to users and services and determining whether it was propagated to that queue.

Open question
- should changes in offers be published to the databus?

Also please make a rebase. Changelog?

Marcelinna:
image
You don't need more information for provider changes?

Marcelinna:
For target user you take description
image
For category and scientific domain not, why?

Michal-Kolomanski:
> image You don't need more information for provider changes?

From the point of our recommender, we do not (we store only id and name).
image

Michal-Kolomanski:
> example for vocabularies: target user you take the description image For category and scientific domain not, why? image

It's also based on that which properties of objects we exactly store in the recommender's database. In that case, Patrick's implementation reflects the recommender's needs. 

BUT, on the other hand, we are introducing NLP, and the issue raised by Marceline is very intriguing. The category description, scientific domain description, providers description, etc., seem to be extremely helpful and valuable in this circumstance. We definitely should store information like descriptions from now. Please address that in this PR.

wujuu:
Also added integration tests:
- ActiveMQ to docker-compose.yml
- `stomp_publisher.yml` and `stomp_subscriber.yml` test configurations

Please check it out @wziajka

michal-szostak (approved):

Marcelinna (approved):

PR #2690, opened 2022-05-30
[#2446] Fix providers import on JMS Resource update

### LINKED ISSUE #2446: Error when updating resource caused by wrong provided by

goreck888:
Removed try to import Provider via REST
when Resource is updated via JMS
Closes #2446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Please view this file on the master branch, on stable branches it's out of date.
 ### Removed
 
 ### Fixed
+- JMS Resource update on missing providers (@goreck888)
 
 ### Security
 
--- a/app/models/concerns/importable.rb
+++ b/app/models/concerns/importable.rb
@@ -111,37 +111,11 @@ module Importable
     Catalogue.find_by(pid: catalogue)
   end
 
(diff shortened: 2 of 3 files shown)


Marcelinna:
- [ ] QA
- [ ] DEV

Steps to reproduce
1. Select resource, edit and add to field Resource Providers - providers which don't exist in MP
2. Click Submit in PC
3. Go to Marketplace - Resource doesn't updated

Field "Provided by" affects the update of the resource on updating resource

Marcelinna (approved):

kmarszalek (approved):

PR #2699, opened 2022-06-07
Add status 301 to url validator

goreck888:
A lot of websites has many aliases
and status 301 for validator should be correct code
Fixes error causes invalid_url for correct addresses,
e.g. http://google.com


app/controllers/backoffice/providers_controller.rb
```diff
@@ -99,6 +99,7 @@ def valid_model_and_urls?
 
     invalid_multimedia =
       @provider.link_multimedia_urls.reject { |media| media.url.blank? ? true : UrlHelper.url_valid?(media.url) }
+    puts "JESTEM TU"
```
kmarszalek, line 102: ...

kmarszalek (approved):

PR #2711, opened 2022-06-27
Add integration test with Provider Portal - providers

Marcelinna:


test/system/cypress/integration/provider_portal/provider_rejected.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+describe("Provider Portal - rejecte provider", () => {
+
+  const token = Cypress.env("PROVIDER_PORTAL_ACCESS_TOKEN");
+  const providerPortalURL = Cypress.env("PROVIDER_PORTAL_URL");
+  const marketplaceURL = Cypress.env("MARKETPLACE_URL");
+  const authorization = `Bearer ${token}`;
+  const provider = providerJson
+
+  it("create provider, rejecte, approve and delete it", { tags: '@extended-test' }, () => {
```
kmarszalek, line 11: type:rejecte

test/system/cypress/support/provider_portal.ts
```diff
@@ -0,0 +1,104 @@
...
+  cy.get("a[data-e2e='more-link-providers']")
+    .click();
+  cy.contains("a", provider)
+    .should("be.visible")
+    .click();
+  cy.contains("a", "Browse resources")
+    .should("be.visible");
+});
+
+Cypress.Commands.add("checkInvisibilityOfProvider", (provider: string) => {
```
kmarszalek, line 37: invisibility may be understood as a provider that exists but is not visible, you use it as a check when there is a provider and when there isn't. 
not sure if there already are these tests, but I would let this one be as it is and add tests that check if a provider exists or not (BO?)

Marcelinna: I will add a test for status in BO, when in the BO we add a search for a providers, now I don't know on which page provider will be displayed in BO so test it will be very hard

kmarszalek (approved):

goreck888 (approved):

Issue #2716, opened 2022-07-04
Include projects data in the recommender's database


Michal-Kolomanski:
The additional data (projects) must be added to the recommender database so the Athena team can read it for their recommender's use case: project completion.

Data which they need: 
For our use cases we need parts of the tables:

- projects
- project_items
- offers

Specifically, the following query retrieves all the information we currently need from the MP database.

```
select projects.user_id, projects.id as project_id, offers.service_id 
from projects, project_items, offers
where projects.id = project_items.project_id and project_items.offer_id = offers.id
```

An example mongo document that covers our needs:
```
{
   user_id: 123
   project_id: 321
   services: [
        12,
        19,
        127
    ]
}
```

**Furthermore, changes should be published via events to the databus.**

This [PR](https://github.com/cyfronet-fid/marketplace/pull/2649) should be merged before starting this ticket .

PR #2722, opened 2022-07-08
[#2716] Add project to recommender dump

wujuu:
As the title says. Added project_id, user_id and ids of services used in a project to the serializer.

NOTE: If a service is included more than one time in a project, (for example it has two offers with the same service), it is serialized as many times as it appears in the project - this is to provide additional data to the recommender team.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -3,6 +3,7 @@
 class Project < ApplicationRecord
   include Eventable
   include Messageable
+  include Publishable
 
   CUSTOMER_TYPOLOGIES = {
     single_user: "single_user",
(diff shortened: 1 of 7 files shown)


michal-szostak (approved):

Michal-Kolomanski (changes requested):
We merged this PR https://github.com/cyfronet-fid/marketplace/pull/2649/files
Please add the module Publishable to this feature.

Michal-Kolomanski (changes requested):
I have tested this PR and it seems to not publish any event when updating the project (for example, adding a service to the already existing project). Events to the **mp_db_events** topic are published only during the creation of the project. 

The dump which is sent to the recommender looks good to me :clap:

Michal-Kolomanski (approved):

PR #2745, opened 2022-08-18
[#2742] Add AAI Token to the Recommender dump

wujuu:


Michal-Kolomanski (changes requested):
Thats my bad but please change aai_token to uid

wujuu:
Can we actually not rename the field? user_id and uid seem way to similar in my opinion

Michal-Kolomanski (changes requested):
The name **aai_token** somewhat distorts its content, and other services would also have to use such terminology, therefore I'm not totally sure we should use it. Maybe we should just call this variable **user_unique_id** similarly to what we did for user actions? I am open to discussion

wujuu:
Changed `aai_token` to `aai_uid`

wziajka (approved):

PR #2748, opened 2022-08-24
Workflow PC integration

Marcelinna:


danielkryska (approved):

danielkryska:
Before merging please add npm install cache
```
 name: Install e2e dependencies
        run: npm i --prefix ./test/system
```

Here is the example: https://github.com/cyfronet-fid/eosc-search-service/blob/master/.github/workflows/ui-styles.yml

goreck888 (approved):

PR #2753, opened 2022-08-29
Add json endpoint for services

goreck888:


goreck888:
to open for user-dashboard services displaying

PR #2755, opened 2022-09-06
Workflow PC integration

Marcelinna:


michal-szostak (approved):

test/system/cypress/integration/provider_portal/provider_approved.spec.ts
```diff
@@ -41,7 +45,7 @@ describe("Provider Portal - approve provider", () => {
       }
     });
 
-    cy.visit(`${marketplaceURL}`);
+    cy.visitPage(`${marketplaceURL}`);
```
goreck888, line 48: Please remove including variable value in string like in line 34 and other lines

goreck888 (approved):

PR #2770, opened 2022-09-29
[#2754] Add similar services recommendations

### LINKED ISSUE #2754: Present similar services from Athena Recommender
1) Firstly, Athena Recommender must be deployed @wziajka 
2) Integrate
3) Based on compatible services (already done), present similar services

goreck888:
Add recommendations to the service `show` page
You can set similar_services_host
using SIMILAR_SERVICES_HOST env variable
Closes #2754

### DIFF
--- a/app/controllers/concerns/service/recommendable.rb
+++ b/app/controllers/concerns/service/recommendable.rb
@@ -42,6 +42,26 @@ module Service::Recommendable
     end
   end
 
+  def fetch_similar(service_id, quantity = 6)
+    user_id = current_user.id
+
+    return if user_id.blank?
+
+    url = "#{Mp::Application.config.similar_services_host}/similar_services/recommendation"
+    body = { user_id: user_id, service_id: service_id, num: quantity }.to_json
+    headers = { "Content-Type": "application/json", Accept: "application/json" }
+    response = Faraday.post(url, body, headers)
+    body = JSON.parse(response.body)
+
+    return if response.status != 200
+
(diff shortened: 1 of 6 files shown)


goreck888:
NOTICE!
Similar services recommendation works only for logged in users. It's caused by specific implementation of user_id including only ids from marketplace's user ids.

Marcelinna (approved):

app/controllers/concerns/service/recommendable.rb
```diff
@@ -42,6 +42,27 @@ module Service::Recommendable
...
+        Faraday.post(
+          url,
+          { user_id: user_id, service_id: service_id, num: quantity }.to_json,
+          { "Content-Type": "application/json", Accept: "application/json" }
+        )
+      body = JSON.parse(response.body)
+      if response.status == 200
+        ids = body.is_a?(Array) ? body.each.map { |e| e["service_id"].to_i } : []
+        Service.where(id: ids, status: %i[published unverified])
+      end
```
danielkryska, line 59: You may flatten nested logic like this:

```
return if not user_id.present?

body = { user_id: user_id, service_id: service_id, num: quantity }.to_json
headers = { "Content-Type": "application/json", Accept: "application/json" }
response = Faraday.post(url, body, headers)
body = JSON.parse(response.body)

return if response.status != 200

ids = body.is_a?(Array) ? body.each.map { |e| e["service_id"].to_i } : []
Service.where(id: ids, status: %i[published unverified])

danielkryska (approved):

PR #2780, opened 2022-10-18
[#2769] Adjust user actions

### LINKED ISSUE #2769: Adjust user_actions
User_actions will be added to SS and UD.
The schema of UA in MP should be adjusted appropriately https://docs.google.com/document/d/1uZPnsj1fG0cTQsuS4XeDSVfLVGO4qFHnXpJ5wqDoL9A/edit#heading=h.8mwkstyi9zyk

PR in which UA are added to SS https://github.com/cyfronet-fid/eosc-search-service/pull/192

Scope:
- Check if `aai_uid`, `client_id` are sent
- `source.root:` no matter which root type it is, always return `resource_id` and add a new field `resource_type` which will be `"service"` in that case
image

Root type: sorted can be omitted right now, as we do not have sort by relevance yet.

wujuu:
Closes #2769.

Not every user-action probe sends information about the service id. It's not how the user actions were originally designed. The original use case involved constructing a tree out of user acitons where the service id would be in the root user action. We do not have access to service object in each of the views or sometimes it is not clear which service.id should be sent.

Alternative approach would be to redesign user action system, where each user action would be independent. "root" key in "source" would be changed to context, to add information about what situation the user action was clicked

### DIFF
--- a/app/javascript/app/user_action.js
+++ b/app/javascript/app/user_action.js
@@ -2,6 +2,8 @@ import Rails from "@rails/ujs";
 import Cookies from "js-cookie";
 import { v1 as uuidv1 } from "uuid";
 
+const ROOT_KNOWN_LABELS = ["recommendation-panel"];
+
 const WINDOW_EVENTS_HANDLERS = {
   beforeunload: async () => {
     if (history.length !== 1) {
@@ -262,19 +264,22 @@ function get_source_by(element, pageIdOverride) {
 }
 
 function get_source_root_by(element) {
+  const resource_type_object = { resource_type: "service" };
   if (!element) {
-    return { type: "other" };
+    return { ...resource_type_object, type: "other" };
   }
 
-  const is_recommendation_panel = element.getAttribute("data-probe") === "recommendation-panel";
-  if (is_recommendation_panel) {
-    return {
-      type: "recommendation_panel",
-      service_id: parseInt(element.getAttribute("data-service-id")),
-    };
-  }
+  // We can add other labels if we need to determine more root types
+  const dataProbeLabel = element.getAttribute("data-probe");
+  const dataProbeType = ROOT_KNOWN_LABELS.includes(dataProbeLabel) ? dataProbeLabel : "other";
 
-  return { type: "other" };
+  const serviceId = parseInt(element.getAttribute("data-service-id"));
+
+  return {
+    type: dataProbeType,
+    ...(!Number.isNaN(serviceId) && { resource_id: serviceId }),
+    ...resource_type_object,
+  };
 }
 
 function get_target_url(element) {
--- a/app/views/comparisons/_service.html.haml
+++ b/app/views/comparisons/_service.html.haml
@@ -8,6 +8,6 @@
       %span.middle
         = service.name.truncate(25, separator: " ")
     .comparison-link.float-right
-      %a{ value: "#{service.slug}", "data-probe": "", "data-e2e": "delete-service-btn",
+      %a{ value: "#{service.slug}", "data-probe": "", "data-service-id": service.id, "data-e2e": "delete-service-btn",
       "data-action": "click->comparison#update" }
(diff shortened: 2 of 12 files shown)


app/javascript/app/user_action.js
```diff
@@ -262,19 +262,26 @@ function get_source_by(element, pageIdOverride) {
...
-      service_id: parseInt(element.getAttribute("data-service-id")),
-    };
-  }
+  // We can add other labels if we need to determine more root types
+  const knownLabels = ["recommendation-panel"];
+  const dataProbeLabel = element.getAttribute("data-probe");
+  const dataProbeType = knownLabels.includes(dataProbeLabel) ? dataProbeLabel : "other";
+
+  // this will be NaN when getAttribute return null
+  const serviceId = parseInt(element.getAttribute("data-service-id"));
```
danielkryska, line 277: The comment on the line is redundant with functional well-written documentation. Please remove it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

app/javascript/app/user_action.js
```diff
@@ -262,19 +262,26 @@ function get_source_by(element, pageIdOverride) {
...
-  if (is_recommendation_panel) {
-    return {
-      type: "recommendation_panel",
-      service_id: parseInt(element.getAttribute("data-service-id")),
-    };
-  }
+  // We can add other labels if we need to determine more root types
+  const knownLabels = ["recommendation-panel"];
+  const dataProbeLabel = element.getAttribute("data-probe");
+  const dataProbeType = knownLabels.includes(dataProbeLabel) ? dataProbeLabel : "other";
```
danielkryska, line 274: If there is a possibility, move the `knownLabels` up to configuration (in class or something) as a frozen const, to split different concerns (data, and implementation).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

danielkryska: The comment is unnecessary if you rename the `knownLabels` const to something like `ROOT_TYPES_LABELS`.

danielkryska: Combining the const name change with placing it higher in the class/structure hierarchy will improve readability and configurability.

app/javascript/app/user_action.js
```diff
@@ -262,19 +262,26 @@ function get_source_by(element, pageIdOverride) {
 }
 
 function get_source_root_by(element) {
+  const resource_type_object = { resource_type: "service" };
+  // If there's no element to reference - we cannot determine the serviceId
   if (!element) {
-    return { type: "other" };
+    return { ...resource_type_object, type: "other" };
   }
```
danielkryska, line 269: You can replace the comment with:

```javascript
const hasServiceId = !!element;
if (!hasServiceId) { ... }

danielkryska (approved):
I'm very impressed by your code quality. You created a lot of variables to improve readability and you named them accordingly. It's written almost as a poem. :p

On the other hand, I see that you overuse comments. Those are required in a very extreme situation, for example:
- specific configuration parameters setup for external libraries
- very weird bug in language and/or framework

In all other cases, the best choice will be to:
- assign values to variables with proper names
- split code into chunks and put them into functions with the specific atomic action names

In the end, I'd like to say that you can leave the code as it is because Marketplace as a repository will be exchanged slowly with other solutions we provide. So clean code practices at a very high level aren't required. 

But... I encourage you to clean this up and learn something for your future self. :)

Additional sources of knowledge:
- https://towardsdatascience.com/why-good-codes-dont-need-comments-92f58de19ad2
- https://medium.com/@bpnorlander/stop-writing-code-comments-28fef5272752

wujuu:
@danielkryska thanks for the comments - refactored the js function a bit

Michal-Kolomanski (approved):

PR #2784, opened 2022-10-21
fix:[HOTFIX] Import `catalogue_id` on REST import

goreck888:
- Add logs for catalogues import
- Fix assigning catalogue_id to the service

### DIFF
--- a/lib/import/catalogues.rb
+++ b/lib/import/catalogues.rb
@@ -25,7 +25,12 @@ class Import::Catalogues
 
   def call
     log "Importing catalogues from EOSC Registry..."
+
     @request_catalogues = external_catalogues_data.select { |pro| @ids.empty? || @ids.include?(pro["id"]) }
+
+    total_catalogue_count = @request_catalogues.length
+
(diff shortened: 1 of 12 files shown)


lib/import/catalogues.rb
```diff
@@ -25,8 +25,13 @@ def initialize(
 
   def call
     log "Importing catalogues from EOSC Registry..."
+
     @request_catalogues = external_catalogues_data.select { |pro| @ids.empty? || @ids.include?(pro["id"]) }
 
+    total_service_count = @request_catalogues.length
```
michal-szostak, line 31: `total_catalogue_count` would be better

michal-szostak (approved):

Marcelinna:
When I visit pr https://marketplace-2.docker-fid.grid.cyf-kr.edu.pl/
image

Marcelinna:
1. I Import catalogues
2. I Import provider from beta.provider provider_2309
3. Why in details catalogue not be visible?

PR 
image

When catalogue is importer, should be visible in Provider details
example from beta
image

Marcelinna (approved):

PR #2795, opened 2022-10-28
[HOTFIX] `catalogue_data` task

goreck888:
Add migration task `rake migration:catalogue_data` 
for import missing catalogue for existing services and providers 

NOTICE!
Task depends on `MP_IMPORT_EOSC_REGISTRY_URL` env variable

### DIFF
--- a/lib/tasks/migration.rake
+++ b/lib/tasks/migration.rake
@@ -1,6 +1,24 @@
 # frozen_string_literal: true
 
 namespace :migration do
+  desc "Import catalogue info for services and providers"
+  task catalogue_data: :environment do
+    types = [{ clazz: Service, method: "resource" }, { clazz: Provider, method: "provider" }].freeze
+    ActiveRecord::Base.transaction do
+      types.each do |type|
+        url = (ENV["MP_IMPORT_EOSC_REGISTRY_URL"] || "https://beta.providers.eosc-portal.eu/api") + "/#{type[:method]}"
+        type[:clazz].find_each do |obj|
+          r = Faraday.get("#{url}/#{obj.pid}")
+          b = JSON.parse(r.body)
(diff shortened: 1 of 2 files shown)


Marcelinna (approved):

lib/tasks/migration.rake
```diff
@@ -1,6 +1,24 @@
 # frozen_string_literal: true
 
 namespace :migration do
+  desc "Import catalogue info for services and providers"
+  task catalogue_data: :environment do
+    types = [{ clazz: Service, method: "resource" }, { clazz: Provider, method: "provider" }].freeze
+    ActiveRecord::Base.transaction do
+      types.each do |type|
+        url = ENV["MP_IMPORT_EOSC_REGISTRY_URL"] || "https://api.eosc-portal.eu/#{type[:method]}/"
```
michal-szostak, line 9: I think this would not work for env variable - `#{type[:method]}` should be concatenated to the whole URL

```
url = (ENV["MP_IMPORT_EOSC_REGISTRY_URL"] || "https://api.eosc-portal.eu") + "/#{type[:method]}/"
```

Also - in all other places where `MP_IMPORT_EOSC_REGISTRY_URL ` is used the default is `https://beta.providers.eosc-portal.eu/api` - maybe we should be consistent here.

lib/tasks/migration.rake
```diff
@@ -1,6 +1,24 @@
...
+        type[:clazz].find_each do |obj|
+          r = Faraday.get("#{url}/#{obj.pid}")
```
michal-szostak, line 11: `/` is added after `url` but above `url` ends with `/`

michal-szostak (changes requested):

lib/tasks/migration.rake
goreck888: Done

michal-szostak (approved):

PR #2799, opened 2022-11-07
Add research category filter

goreck888:
Add research category filter to the service search


agpul (commented):
From the PO point of view, I suggest moving this filter to the top. 
image

michal-szostak (approved):

PR #2809, opened 2022-11-11
feat: add research step links to home page

michal-szostak:


roksanaer (approved):
all good. just the mahage research categories icon needs to be updated and the button under icon categories (Browse all resources -> https://search.eosc-portal.eu/search/all?q=*)

Marcelinna (approved):

PR #2815, opened 2022-11-14
fix:[#HOTFIX] Displaying datasources in main search

goreck888:
Fix displaying datasources in resources search

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Please view this file on the master branch, on stable branches it's out of date.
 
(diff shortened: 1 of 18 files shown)


Marcelinna:
Beta
image
PR
image

we have a regression on this pr about recalculating filter values ​​in Backoffice.
For me, regression should not go into master. @goreck888 he said it was hard to fix, @michal-szostak maybe you could help?
@roksanaer @agpul If the decision will be to merge this, please comment

michal-szostak (approved):

michal-szostak:
I agree with @Marcelinna - we should not merge it into the master. The current code is ok, but that issue must be fixed before proceeding.

PR #2858, opened 2023-01-03
Fix integration test with PC

Marcelinna:


goreck888 (approved):

test/system/cypress/integration/provider_portal/provider_approved.spec.ts
```diff
@@ -32,6 +32,7 @@ describe("Provider Portal - approved provider", () => {
     });
 
     cy.visitPage(marketplaceURL);
+    cy.wait(40000)
```
michal-szostak, line 35: I think that instead of always waiting maybe it would be possible to modify 

```
cy.checkInvisibilityOfProviderInMarketplace(provider.name)
```

in `test/system/cypress/support/provider_portal.ts`.

It should be possible to use timeout in `get`:

```
Cypress.Commands.add("checkInvisibilityOfProviderInMarketplace", (provider: string) => {
  cy.get("a[data-e2e='more-link-providers']", {timeout: 40000})
    .click();
  cy.contains("a", provider)
    .should("not.exist");
  cy.get("[data-e2e='searchbar-input']").
    type(provider, { force: true });
  cy.contains("[data-e2e='autocomplete-results'] li", "Provider")
    .should("not.exist");
});
```

In worst case scenario it would be the same, but in best case we should save some time.

michal-szostak (changes requested):

test/system/cypress/integration/provider_portal/provider_approved.spec.ts
Marcelinna: @michal-szostak  adding timeout doesn't resolve my problem. Adding a timeout increases the wait for an element. 
But in my case "a[data-e2e='more-link-providers' is visible immediately, so cypress will go to the /provider page right away, where provider will not be visible because it has not yet been imported from the PC.

I can't wait longer for provider on provider page too, because until the page is not refreshed, the provider will not be visible. anyway

michal-szostak: Alright, I see the issue - in this case I'd at least move the `.wait` to the `cy.checkVisibilityOfProviderInMarketplace` so we do not need to change the wait values every time there is a need to fine tune.

Marcelinna: @michal-szostak done

michal-szostak (approved):

PR #2866, opened 2023-01-09
Fix deprecated scopes in AAI

goreck888:
Fix the possibility of login from AAI,
updated scopes available for the new version of oidc


Michal-Kolomanski (approved):
LGTM

PR #2900, opened 2023-02-20
feat: [#2884] Update blue bar and commons

### LINKED ISSUE #2884: Change commons and the blue bar in all services managed by Cyfronet
1. White bar - commons

- [ ] change the white upper bar which is in commons according to the project from figma (top bar sequence of buttons: EOSC logo - EOSC Portal - Browse Marketplace - Providers Dashboard - Monitoring - Status - Contact Us - Log-in dropdown pic)

- [ ] Under my name (for eg. Agnieszka Pułpa) move drop-down menu from "My EOSC", see picture: 
image
 (It is valid only for MP, SS and UD - it should be configurable, not avaiable for eg from Providers Dashboard )
Screenshot 2023-01-30 at 14 17 11

2. Blue bar with search
- [ ] add blue bars according to the project from Figma (in most cases it will back to search [search.marketplace.eosc-portal.eu](http://search.marketplace.eosc-portal.eu/) ), search remains only on the landing page of the Marketplace and on the search page
Behavior:
- [ ] Service detail page in MP: "Go to search" bar, which redirects to: http://search.marketplace.eosc-portal.eu/
- [ ] On the project page in MP: "Go to search"
- [ ] project detail page in SS: "Go to search bar"
- [ ] In EOSC Portal, we don't add the blue bar at all
- [ ] EOSC Explore - has to add itself, independently of us, "Go to search"
- [ ] Search Service - do nothing
- [ ] MP homepage - new landing page
- [ ] User Dashboard - lack of blue bar

Figma project:

https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Search-Service?node-id=672%3A4150&t=Q3TnbXEpg63tp8ue-0

michal-szostak:
Closes #2884

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -92,5 +92,9 @@ module Mp
       else
         Rails.env.test?
       end
+
+    config.enable_external_search = ENV.fetch("MP_ENABLE_EXTERNAL_SEARCH", false)
+
+    config.whitelabel = ENV.fetch("MP_WHITELABEL", false)
   end
 end
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ We are currently using the following ENV variables:
   * `MP_STOMP_PUBLISHER_TOPIC` (Optional) - topic of a new message to be published (default `credentials.stomp.publisher.topic`)
   * `MP_STOMP_PUBLISHER_MP_DB_EVENTS_TOPIC` (Optional) - topic for publishing database create, update and destroy events (default `mp_db_events`)
   * `MP_STOMP_PUBLISHER_USER_ACTIONS_TOPIC` (Optional) - topic for publishing user actions events(default `user_actions`)
+  * `MP_ENABLE_EXTERNAL_SEARCH` - should be set to `1` if external search is used. If set top bars will be replaced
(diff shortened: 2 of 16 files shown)


config/application.rb
```diff
@@ -92,5 +92,8 @@ class Application < Rails::Application
       else
         Rails.env.test?
       end
+
+    config.whitelabel =
+      ENV["MP_WHITELABEL"].present? ? ENV["MP_WHITELABEL"] : false
```
goreck888, line 97: You can use ENV.fetch("MP_WHITELABEL", false) instead of this.

Marcelinna:
@michal-szostak blue bar should be visible on main page ?
image

goreck888 (approved):

michal-szostak:
@Marcelinna - this page would not be visible in the production instance, right? If so then blue bar is not an issue.

Marcelinna:
> @Marcelinna - this page would not be visible in the production instance, right? If so then blue bar is not an issue.

@michal-szostak this page it will not be visible, but visible will be another main page. 

If there will be another page at the same address, the blue bar will not be visible?

Marcelinna (approved):

PR #2910, opened 2023-03-03
[#2896] Send user action to the databus on project addition

### LINKED ISSUE #2896: User actions - parameter Order is not set to true after adding the service to the project
When the user adds a service to the project, the parameter order should always be set to true, for all types of services

wujuu:
Closes #2896.

### DIFF
--- a/app/controllers/services/summaries_controller.rb
+++ b/app/controllers/services/summaries_controller.rb
@@ -92,6 +92,12 @@ class Services::SummariesController < Services::ApplicationController
       client_id: "marketplace"
     }
 
-    Probes::ProbesJob.perform_later(request_body.to_json)
+    if %w[all recommender].include? Mp::Application.config.user_actions_target
+      Probes::ProbesJob.perform_later(request_body.to_json)
+    end
+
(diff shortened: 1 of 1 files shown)


Marcelinna:
@wujuu When I add service to the project I see 2 messages, one with order -> true, one with order -> false
image

Michal-Kolomanski:
> @wujuu When I add service to the project I see 2 messages, one with order -> true, one with order -> false image

Those messages have different panel_id in their target. Please correct me @wujuu if I am wrong, but:
1) First UA is about the order=True
2) Second UA is about the redirection to the /projects page (there are also UA)

Michal-Kolomanski (approved):

wujuu:
> > @wujuu When I add service to the project I see 2 messages, one with order -> true, one with order -> false image
> 
> 
> 
> Those messages have different panel_id in their target. Please correct me @wujuu if I am wrong, but:
> 
> 1) First UA is about the order=True
> 
> 2) Second UA is about the redirection to the /projects page (there are also UA) 

Yes, I confirm

Marcelinna (approved):

PR #2941, opened 2023-03-22
feat: add search service terms of use and privacy policy

michal-szostak:
ENV variable should be set to:

```
EOSC_COMMONS_BASE_URL=https://s3.cloud.cyfronet.pl/eosc-portal-common/pr-229-add-possibility-to-customize-privacy-policy-and-term-of-use-in-the-footer-per-service/
```


Marcelinna (approved):

Marcelinna:
@michal-szostak 
tests are run on wrong instance, we should have MP search bar in localhost
image

PR #2999, opened 2023-05-31
feat: Add API endpoints for mp collections

goreck888:
Add new endpoints for rendering data to SOLR
Data is accessible via the path:
`api/v1/ess/{{ collection }}`
where the collection can be:
- services
- datasources
- providers
- offers
- bundles There is a possibility to get data by id,
and in case of service/datasource/provider
slug or pid
just by add `/{{ id }}` to path,
eg. `/api/v1/ess/services/egi-cloud-compute`

Api documentation available [here](https://marketplace-1.docker-fid.grid.cyf-kr.edu.pl/api_docs/swagger/index.html)

NOTICE!
To get the data you have to login
or authenticate via the user's access_token
with service_portfolio_manager role


Michal-Kolomanski (changes requested):
Good job! I have a few remarks. 
1) I would rather see 1 endpoint with a parameter describing which type of data I want to receive, not N-endpoints where N is the number of data types.

2) Retrieving all data sources seems to not work
image

3) There are some minor differences between dump data and data from the API.

Services:
Left side data from API, right side data from the dump
image
image

Providers and bundles:
- scientific_domains - the same case

4) Retrieving data seems to be slow. I am a little bit worried that if I would try to retrieve all types of data at once it would blow up the whole API, probably causing timeouts. Maybe it's not a big deal but I wonder why this is the case? @goreck888 

Summarize:
I was testing, the code was not checked by me. Point 2) is critical, others may not be considered ASAP.

michal-szostak (approved):

Michal-Kolomanski (approved):
Please add also a section about this API in the README. Also please note there that you need a token to get data.

PR #3003, opened 2023-06-07
feat:[] Add tags to offers and bundles

goreck888:
Add tags (keywords) to the backoffice forms for offers and bundles. It allows sending tags to the SOLR directly from an offer/bundle.
This PR Is created upon @roksanaer's request.

We can also change a schema for SOLR in this PR, but it should be discussed with @Michal-Kolomanski


Marcelinna:
1. Add a tag to offer and save 
image
2. Click edit button offer
3. See problem - tag is not visible
image

the same situation for bundle

Marcelinna (approved):

michal-szostak (approved):

PR #3037, opened 2023-07-20
Fix zammad form

michal-szostak:


goreck888 (approved):
Could you add a changelog entry?

Marcelinna (approved):

PR #3045, opened 2023-08-06
[#3023] Hadle suspended status in PC

### LINKED ISSUE #3023: Suspended status
You need to handle the new status Suspended

- when we get information suspended = true, service, data source, provider should have draft status in MP

we need to add changes to JMS and Import (add &suspended=false)

image

goreck888:
Add `suspended` flag to handling
resource status in JMS and REST communication
with the Provider's component
Closes #3023


Marcelinna:
I import provider from PC with suspended status and has publish status in MP

image
image
image

.gitignore
```diff
@@ -45,4 +45,4 @@ yarn-debug.log*
 # Ignore IDE configuration
 /.idea
 /.env
-/.run
```
michal-szostak, line 48: This should ignore everything in `/.run` I'm not sure why should it be more specific?

michal-szostak (approved):

Marcelinna (approved):

PR #3111, opened 2023-10-16
[#3094] Datasources as services subprofiles

### LINKED ISSUE #3094: Datasources as a subServices
- [ ] As a User I would like to see the data sources as a sub catalog of services

goreck888:
Add necessary changes to importers
Add new PID to import
Closes #3094


Marcelinna:
duplication
image

Marcelinna:
@goreck888 it is okay that this value is different? it was always the same value
image

Marcelinna:
1. Create service and add sub profile service
2. Change service status to suspended or inactive in PC (In PC service and sub profile data source have the same status)
image
3. in Marketplace data source is changed to service and have published status (should be dataosurce and have draft status)
image

Marcelinna:
Maybe in other PR
should be added to datasource scheme
image

Marcelinna:
Should be added to Services scheme
image
Marketplace Locations (*)  == Research steps

Marcelinna:
When I approve Provider with Alternative Identifiers, then it won't show up in the Marketplace, I have to delete it Alternative Identifiers
image

Marcelinna:
Second and the second and subsequent service are created without offers 
image

config/brakeman.ignore
```diff
@@ -3,58 +3,20 @@
     {
```
wujuu, line 3: qq: is this file generated or we have to define the lines and everything manually?

lib/tasks/import.rake
```diff
@@ -7,62 +7,62 @@ namespace :import do
 
   task resources: :environment do
     Import::Resources.new(
-      ENV["MP_IMPORT_EOSC_REGISTRY_URL"] || "https://beta.providers.eosc-portal.eu/api",
-      dry_run: ENV["DRY_RUN"] || false,
-      default_upstream: (ENV["UPSTREAM"] || "eosc_registry").to_sym,
-      ids: (ENV["IDS"] || "").split(","),
-      filepath: ENV["OUTPUT"],
-      token: ENV["MP_IMPORT_TOKEN"]
+      ENV.fetch("MP_IMPORT_EOSC_REGISTRY_URL", "https://beta.providers.eosc-portal.eu/api"),
```
wujuu, line 10: nit: shouldn't we have all constants defined in [Rails::config](https://github.com/cyfronet-fid/marketplace/blob/master/config/application.rb) ?

Unless this config is not availble at rake task exec

wujuu (approved):
LGTM, 1 nit and 1 question

config/brakeman.ignore
goreck888: It's automatically generated

lib/tasks/import.rake
goreck888: Rake tasks don't use Application config

Marcelinna (approved):

PR #3118, opened 2023-10-30
HOTFIX:[] Fix views cache uploading and CatalogueSerializer

goreck888:
Now store_analytics casts Google API response to integer 
Change serializing catalogue as pid instead of name


tomaszandel (approved):

Marcelinna:
EGI Nottebok 
image

beta
image
production
image

Marcelinna (approved):

PR #3133, opened 2023-11-15
fix: Fix JMS import

goreck888:
This PR fixes an error caused by a change in the JMS message schema.
Removed `resourceExtras` from the `serviceBundle` entity.


Michal-Kolomanski (approved):
Not sure what it is exactly fixing but I assume its necessary

PR #3135, opened 2023-11-21
fix:[#3123] Fix displaying monitoring data in datasources

### LINKED ISSUE #3123: Add Monitoring data for Datasources
Acceptance criteria:

- [ ] As a User I would like to see the monitoring data for data sources (similar like for services)

goreck888:
Unified sidebar for services and datasources
Now monitoring data and statistics are displayed
Closes #3123


datasource_ids.json
```diff
@@ -0,0 +1,149 @@
+{
```
wujuu, line 1: shouldn't we fetch this from somewhere? what if a new datasource is added that is not in this mapping?

goreck888: I just received a static list of migrated objects from Athena. It's a temporary file needed for one migration and I could ask @wziajka to load it after the production deploy on the machine and before merging I can delete it.

Marcelinna (approved):

PR #3140, opened 2023-11-27
fix:[HOTFIX] Datasources import fix

goreck888:
Fix `import:datasources` task
Add missing change of `type`
from `Service` to `Datasource`


Michal-Kolomanski (approved):

Marcelinna:
In Mp I see service after import, but I should see datasource
image
image

PR #3150, opened 2023-12-05
fix:[]Omission of `EOSC::` prefix in serialized eosc_if

goreck888:
Remove the `EOSC::` prefix from the
interoperability patterns serialized
to the Search Service
Add task `rake viewable:heal_tags` with default values 
which can be changed by the `HEAL_TAG_LIST` env variable


Michal-Kolomanski (changes requested):
1) 
```
"eosc_if": [
      "ro-crate"
    ],
```
it should be "RO-crate"

2) 
```
    "eosc_if": [
      "galaxy workflow"
    ],
```
it should be "Galaxy Workflow" 

In general, those values should be capitalized, like here (without EOSC::)
image

- please also adjust detail pages and redirections

Marcelinna (approved):

Issue #3212, opened 2024-02-13
Suspended view for Catalogue


Marcelinna:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

goreck888:
Resolved in #3218

Issue #3201, opened 2024-02-07
Add suspended feature for Provider 


agpul:
Acceptance criteria:
- [x] As a User I would like to see suspended page for Provider page (all page should be grey and inactive)

PR #3218, opened 2024-02-14
feat:[#3201, #3212] Handle `suspended` status in importable

goreck888:
Add `suspended` grey view for `Provider`
and `Catalogue`

Closes
- #3201
- #3212


Marcelinna:
should be similar to service Manage the provider button should be not visible
image

Issue #3269, opened 2024-04-19
Visual adjustments for offers


AlicjaSwierad:
Figma: 
https://www.figma.com/file/0F53l8h6P2Q7TpzLuFZ7qX/Offers%3A-Adding-and-Editing?type=design&node-id=516%3A55179&mode=design&t=jrj3zifEgwbknQcS-1

- [ ] Remove lines
![image](https://github.com/cyfronet-fid/marketplace/assets/145433362/e2273be1-fb1d-44b5-ab19-04303b23c516)

- [ ] Change tooltip style (add) to URL restrictions
![image](https://github.com/cyfronet-fid/marketplace/assets/145433362/8f8730fa-a167-4faa-9a39-cb01793f39c0)

PR #3295, opened 2024-05-28
styles:[#3269] visual adjustments rails 7

goreck888:


Marcelinna (changes requested):
Please delete section about URL when ordering is EOSC OMS, in this case why don't have URL input
image

Marcelinna (approved):

Michal-Kolomanski (approved):

Issue #3299, opened 2024-05-31
Covered list


Marcelinna:
In form list of elements are covered
![image](https://github.com/cyfronet-fid/marketplace/assets/59872096/1da758b2-248a-4060-871b-4c4b375f6653)

PR #3305, opened 2024-06-06
fix:[#3299] overflow accordion fix

jarekzet:
Fix for #3299 

image


goreck888 (approved):
Tested on localhost because CheckIn has some issues and can't login on PR instance.

Marcelinna (approved):

PR #3317, opened 2024-07-05
fix: Remove eosc message

goreck888:


Michal-Kolomanski (approved):
If there are some styles made only for `top-info-wrapper` please delete also them

Marcelinna:
I still see Importance inscription
image

wziajka (approved):

PR #3369, opened 2024-12-16
duplicate offer

maria-j-k:


app/controllers/backoffice/services/offers_controller.rb
```diff
@@ -76,6 +76,22 @@ def fetch_subtypes
     render json: json
   end
 
+  # POST /backoffice/services/<service_slug>/offers/:id/duplicate
+  def duplicate
+    original_offer = @service.offers.find_by(iid: params[:offer_id])
```
goreck888, line 81: You should create `duplicates_controller.rb` (near the `publishes_controller`) with create method. That is more clean than your idea of putting non-CRUD method in the offers_controller.

goreck888 (approved):

goreck888 (commented):
Just add a changelog entry.
I think we can improve routes a little, but it's not necessary now.
And please check the failing test.

app/controllers/backoffice/services/offers_controller.rb
```diff
@@ -6,15 +6,15 @@ class Backoffice::Services::OffersController < Backoffice::ApplicationController
   before_action :load_form_data, only: %i[fetch_subtypes]
   after_action :reindex_offer, only: %i[create update destroy]
 
-  helper_method :exit_title, :save_as_draft_title, :save_title
+  helper_method :exit_title, :save_as_draft_title
```
goreck888, line 9: I think it should be the same code like in whitelabel :)
Create offers_helper and move this code there.

PR #3521, opened 2025-09-13
feat:[##1556] Add offer support to Deployable Services

wujuu:


goreck888 (approved):
Probably this PR needs a considered refactoring in the future. Check my comments.
Approved, but before merge please run `bundle exec rbprettier --write .` to fix the styles task and then amend commit without `--no-verify` option 🙂

PR #3570, opened 2025-11-28
fix:[#3569] Fix projects

### LINKED ISSUE #3569: Fix projects
**Describe the bug**
There's no possibility of getting projects path after making any order.

**To Reproduce**
1. Order any service
2. Click on the projects path in the menu
4. See an error

**Expected behavior**
We should achieve the projects path after the order

goreck888:
Allow user to see projects path.
Before the patch we've got
an error page.
Discovered probable reason: 
missing service_id parameter in recommendations

Closes #3569


Michal-Kolomanski:
I guess its impossible to test on the PR instance

AlicjaSwierad (approved):

PR #3584, opened 2025-12-15
feat: Add polymorphic orderable association and infrastructure tracking

wujuu:
Refactor Offer model to use polymorphic association for Service and DeployableService, and add infrastructure lifecycle management for deployable service deployments.

Polymorphic Orderable
---------------------
- Replace service_id and deployable_service_id columns with polymorphic orderable_type/orderable_id on offers table
- Add parent_service method as unified accessor for both types
- Add backward-compatible service/deployable_service accessors
- Create reusable join scopes: join_service, join_orderable, with_published_orderable and SQL constants for external use
- Update all policies, controllers, and serializers to use new scopes

OrderableResource Concern
-------------------------
- Extract shared interface for Service and DeployableService
- Define required methods: name, description, offers, owned_by?, etc.
- Add common scopes: visible, active

Infrastructure Tracking
-----------------------
- Add Infrastructure model with state machine (pending -> creating -> configured -> running -> failed/destroyed)
- Add InfrastructureManager::Client for IM API integration with token caching and refresh
- Add DeploymentJob for creating infrastructures
- Add StatePollingJob for monitoring infrastructure state
- Add DestroyJob and UI endpoint for infrastructure teardown
- Extract IM configuration to config/deployable_services.yml

Database Migrations
-------------------
- 20251213211539_make_offer_orderable_polymorphic
- 20251214080854_create_infrastructures
- 20251214143117_remove_deprecated_offer_columns

### DIFF
--- a/app/controllers/projects/services_controller.rb
+++ b/app/controllers/projects/services_controller.rb
@@ -33,11 +33,14 @@ class Projects::ServicesController < ApplicationController
 
   def sample_recommended_offers
     if @project_items.present?
-      domains = @project.scientific_domains.presence || @project_items.first.service.scientific_domains
+      domains = @project.scientific_domains.presence || @project_items.first.service&.scientific_domains
+
+      return Offer.all.sample(2) if domains.blank?
 
       Offer
-        .joins(service: :scientific_domains)
-        .where(scientific_domains: { id: domains.map(&:id) })
+        .join_service
+        .joins("INNER JOIN service_scientific_domains ON service_scientific_domains.service_id = services.id")
+        .where(service_scientific_domains: { scientific_domain_id: domains.map(&:id) })
         .distinct
         .sample(2)
         .presence || Offer.all.sample(2)
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -64,7 +64,8 @@ class ServicesController < ApplicationController
     @similar_services = fetch_similar(@service.id, current_user&.id)
     @related_services = @service.related_services
 
-    @service_opinions = ServiceOpinion.joins(project_item: :offer).where(offers: { service_id: @service })
+    @service_opinions =
+      ServiceOpinion.joins(project_item: :offer).where(offers: { orderable_type: "Service", orderable_id: @service.id })
     @question = Service::Question.new(service: @service)
     @favourite_services =
       current_user&.favourite_services || Service.where(slug: Array(cookies[:favourites]&.split("&") || []))
--- a/app/jobs/deployable_service/deployment_job.rb
+++ b/app/jobs/deployable_service/deployment_job.rb
@@ -1,32 +1,19 @@
 # frozen_string_literal: true
 
 class DeployableService::DeploymentJob < ApplicationJob
-  queue_as :default
+  queue_as :deployments
 
   def perform(project_item)
     Rails.logger.info "Starting deployment for ProjectItem #{project_item.id} (#{project_item.offer&.name})"
 
-    filled_template = DeployableService::ToscaTemplateFiller.call(project_item)
-    deployment_address = deploy_to_infrastructure_manager(project_item, filled_template)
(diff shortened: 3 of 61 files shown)


app/controllers/projects/services_controller.rb
```diff
@@ -33,11 +33,14 @@ def load_projects
...
+      domains = @project.scientific_domains.presence || @project_items.first.service&.scientific_domains
+
+      return Offer.all.sample(2) if domains.blank?
 
       Offer
-        .joins(service: :scientific_domains)
-        .where(scientific_domains: { id: domains.map(&:id) })
+        .join_service
+        .joins("INNER JOIN service_scientific_domains ON service_scientific_domains.service_id = services.id")
+        .where(service_scientific_domains: { scientific_domain_id: domains.map(&:id) })
```
goreck888: What is the reason for this change? Does it increase speed of queue or something like that?

app/models/offer.rb
```diff
@@ -13,26 +13,60 @@ class Offer < ApplicationRecord
   searchkick word_middle: %i[offer_name description], highlight: %i[offer_name description]
 
   def search_data
```
goreck888: This model is now very fat.

app/controllers/services_controller.rb
```diff
@@ -64,7 +64,8 @@ def show
     @similar_services = fetch_similar(@service.id, current_user&.id)
     @related_services = @service.related_services
 
-    @service_opinions = ServiceOpinion.joins(project_item: :offer).where(offers: { service_id: @service })
+    @service_opinions =
+      ServiceOpinion.joins(project_item: :offer).where(offers: { orderable_type: "Service", orderable_id: @service.id })
```
goreck888: I suppose that query isn't optimal, but I have no idea how make it better.

app/jobs/deployable_service/deployment_job.rb
```diff
@@ -6,27 +6,14 @@ class DeployableService::DeploymentJob < ApplicationJob
   def perform(project_item)
     Rails.logger.info "Starting deployment for ProjectItem #{project_item.id} (#{project_item.offer&.name})"
 
+    # Create Infrastructure record to track the deployment
```
goreck888: Maybe we could use a dedicated :deployments queue? Just a suggestion.

app/serializers/ess/offer_serializer.rb
```diff
@@ -15,6 +14,9 @@ class Ess::OfferSerializer < ApplicationSerializer
              :parameters,
              :updated_at
 
+  # Resource ID (service or deployable_service) - uses polymorphic orderable
+  attribute :orderable_id, key: :service_id
```
goreck888: Does need the TransformService to include this serialization change? I suppose so.

goreck888 (approved):
I think it's quite good. In the comments you can find some questions and improvement requests.

app/serializers/ess/offer_serializer.rb
wujuu: I've added orerable_type to the seralizer. Probably this will need to be adjusted in the TransformService as well @Michal-Kolomanski

app/models/infrastructure.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  # State predicates
+  STATES.each { |state_name| define_method("#{state_name}?") { state == state_name } }
+
+  def deployable_service
+    project_item.offer.parent_service
+  end
+
+  def mark_created!(infrastructure_id)
+    update_columns(im_infrastructure_id: infrastructure_id, state: "creating", last_state_check_at: Time.current)
```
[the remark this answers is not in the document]
wujuu: We use update_columns here because these are background job state transitions that happen rapidly (pending → creating → configured → running). The state values are hardcoded strings, so validation would always pass anyway. Using update! would add overhead and potentially trigger callbacks we don't want during these internal state machine updates.

PR #3587, opened 2025-12-31
fix: [#3581] Fix the "node" in DS Serializer

### LINKED ISSUE #3581: Fix node value for deployable services returned from API
How to reproduce:
1) Get all deployable services from MP API
2) see that in the response body you have: `"node": "node-sandbox"`. It should be `Sandbox`

Goal:
- [ ] replace raw value with a pretty name

wujuu:
Adjust the "node" value to return:
- "node-sandbox" → "Sandbox"
- "docker" → "Docker"
- "kubernetes" → "Kubernetes"

Closes #3581


Michal-Kolomanski (commented):
I thought that we store node's vocabulary and can return the pretty name directly.
If not, the proposed transformation is okay

Issue #3531, opened 2025-09-25
Fix deployable services synchronization


Michal-Kolomanski:
Steps to reproduce:
1) onboard new DS - its returned by PC API, but we are not returning it in MP API
2) update DS - changes are not reflected on MP API

wujuu:
Fixed in https://github.com/cyfronet-fid/marketplace/pull/3588

PR #3588, opened 2025-12-31
fix:[#3531] Fix deployable services synchronization

wujuu:
Create DeployableServiceSource when syncing new DeployableServices from Provider Catalog, mirroring the pattern used by Service sync.


goreck888 (approved):
I have just one improvement request. In the project details the password is public. I think we should fix that.
And could also you fix styles check? `bundle exec rubocop -A` should do the work.

Image

Issue #3601, opened 2026-01-27
Deployable services - user dashboard


AlicjaSwierad:
In this task, we focus on the details page of the deployable service. Most of the changes are styling. Besides styling, we need to add a banner that shows when the deployment process is in progress, and a link to email the provider. In addition, we need to rename “Service Organisation” to “Provider” to keep the naming consistent across the service. Additionally, we prefer to keep status names consistent with other places in the system.

- [x] styling
- [x] service Organisation -> Provider
- [ ] status names 
- [x] Deployment in progress banner

Figma: https://www.figma.com/design/AkfoCc611AN1FuMfHbDVr1/Dashboard?node-id=9417-6084&t=j4WJpRVzKaOd8ebh-1

PR #3605, opened 2026-01-29
3601 deployable services user dashboard

jarekzet:
Closes: #3601 

added content mods due to Figma. Deploy banner shows when status of project.item is = project_item.in_progress?

with baner:
image

without:
image

### DIFF
--- a/app/views/projects/services/_layout.html.haml
+++ b/app/views/projects/services/_layout.html.haml
@@ -21,15 +21,15 @@
           %h1.text-break
             #{project_item.service.name}
 
-            -# haml-lint:disable LineLength
-            - if policy(project_item).destroy_infrastructure?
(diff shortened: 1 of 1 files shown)


AlicjaSwierad:
1. The “Go to Service” button should appear once the DS is deployed. @Michal-Kolomanski can we do it? 
image

2. Where does this link lead? I think we can hide it and add it after creating a DS list page/view.
image

3. Can we rename the statuses so they match the rest of the portal? @Michal-Kolomanski  And then style it? :D

4. Error message
image

5. Underline links
image

_______
**To sum up, it looks nice. 
For now, we should focus on fixing backend issues (the button and statuses).
The only other thing we could still improve is the styling of the error message and maybe add an underline to links.**

jarekzet:
added mods:

image

image

PR #3716, opened 2026-06-24
fix: federated search api changes

### LINKED ISSUE #3699: Federated Search Final
http://federatedsearch.service.eosc-beyond.eu/federation/services

What has changed:
- Now we have a `pid` attribute in both `nodes` and `node` - we should use this attribute to map one to another

Scope:
- [ ] Construct the Federated Search filter on `metadata -> nodes -> name`. The content of the filter itself should be static, and those values should be a list of names 
- [ ] But in the backend, the search query should be constructed based on the facet. Mapping should be done using `pid`

sqrasdf:
closes #3699

### DIFF
--- a/app/controllers/federation/services_controller.rb
+++ b/app/controllers/federation/services_controller.rb
@@ -115,26 +115,24 @@ class Federation::ServicesController < ApplicationController
     facet_values.to_h { |v| [v["value"], v["label"]] }
   end
 
-  def map_results(json)
+  def map_results(json, pid_to_name)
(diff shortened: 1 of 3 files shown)


Michal-Kolomanski:
Please hide this social media component from federated search.
It should be shown only on the landing page, so here https://userspace.sandbox.eosc-beyond.eu/
and not on different routes - e.g. detail pages, ordering, federated search. Check out the routing 
image

Michal-Kolomanski (approved):
👏

Michal-Kolomanski (approved):
👏

PR #3741, opened 2026-08-05
fix: repair dev:prime seed task crashing on fresh setup

zlekki:
## Summary
- `db/data.yml` was missing its `target_users:` top-level key, leaving that block nested under `providers:`; `create_providers` then tried to build a provider out of each target-user entry and crashed on the missing `multimedia` key.
- `Catalogue#affiliations=` called `value.compact_blank` without a nil guard, unlike its sibling `participating_countries=`, so catalogues without affiliations in the seed data raised `NoMethodError`.
- `create_target_users` didn't guard against a missing `target_users` hash, unlike the other seed methods in the same file.
- Documented in the README that `dev:prime` requires Elasticsearch indices to exist beforehand (`searchkick:reindex:all`), since it saves Searchkick-backed records and otherwise aborts with `Searchkick::ImportError` partway through.

## Test plan
- [x] `bundle exec rails db:drop db:create db:migrate`
- [x] `bundle exec rails searchkick:reindex:all`
- [x] `bundle exec rails dev:prime` completes end-to-end with `Done!`
- [x] Verified record counts (58 services, 45 providers, 78 offers, 24 target users) and that all are synced to their Elasticsearch indices

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,14 @@ you can add new records by editing `db/data.yml`.
 It is important to remember that if some record
 is a parent for another, it must be written above its child.
 
+`dev:prime` saves records that are synced to Elasticsearch via Searchkick
+(`Service`, `Offer`, `Bundle`, `Provider`), so their indices must already
+exist beforehand, otherwise the task aborts partway through with
+`Searchkick::ImportError: index_not_found_exception`. Make sure Elasticsearch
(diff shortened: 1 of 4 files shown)


schodevio-ackcyf (approved):
Looks good

PR #3742, opened 2026-08-06
fix: auth mock login redirect and regenerate schema from migrations

zlekki:
## Summary
- `Users::AuthMockController#login` (the dev-only `AUTH_MOCK=true` login bypass at `/users/login`) never called `render` or `redirect_to`, so hitting it from a browser failed with "missing template" instead of signing the user in. It now redirects to `/backoffice` after a successful sign-in (or immediately if already signed in), and returns 403 when the mock login is disabled.
- `db/schema.rb` had `orderable_type`/`orderable_id` on `offers` marked `null: false`, which doesn't match what `20251213211539_make_offer_orderable_polymorphic.rb` actually produces — those columns are intentionally nullable there, to support orphan offers during the polymorphic migration, with no later migration adding the constraint. Regenerated `schema.rb` from a clean `db:migrate` run so it reflects real migration history instead of stale/hand-edited state.

## Test plan
- [x] `bundle exec rails db:drop db:create db:migrate` produces the committed `schema.rb` with no further diff
- [x] Visited `/users/login?email=...&password=...&first_name=...&last_name=...` with `AUTH_MOCK=true` and confirmed it signs in and redirects to `/backoffice` instead of raising `ActionController::MissingExactTemplate`


schodevio-ackcyf (approved):
Looks good

PR #3751, opened 2026-08-12
fix: [#3725] logos for deployable applications

### LINKED ISSUE #3725: Fix logos for Deployable Applications
Issue:
https://search.userspace.sandbox.eosc-beyond.eu/search/deployable_service?q=*&standard=true&exact=false&radioValueAuthor=A&radioValueExact=A&radioValueTitle=A&radioValueKeyword=A
Image

Scope:
- [x] Identify the issue 
- [ ] fix the issue. DA should have their own Logo or a default one

schodevio-ackcyf:
Resolves #3725 

NOTE:
This is a hotfix to display logos correctly. Ideally fetching logos should be moved to background job

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logo importing across services, providers, datasources, catalogues, and deployable services.
  * Invalid, unavailable, blank, or non-image logo URLs are now skipped safely.
  * Improved handling of timeouts, unreachable sources, and unsupported image files.
  * Deployable services now use the correct dedicated fallback logo.

* **Improvements**
  * Logos are converted to PNG format when needed.
  * Logo attachments are applied consistently during creation and updates.
  * Deployable service logos now render with resizing when available.


Michal-Kolomanski:
To test: https://marketplace-6.docker-fid.grid.cyf-kr.edu.pl/deployable_services

spec/services/importers/logo_spec.rb
```diff
@@ -3,30 +3,70 @@
...
 
-    expect(service.logo).to_not be_nil
-  end
+      attachable = described_class.new(url).call
+
+      expect(Vips::Image).to have_received(:new_from_buffer).with("png-bytes", "")
+      expect(attachable[:io]).to be_a(StringIO)
+      expect(attachable[:filename]).to end_with(".png")
+      expect(attachable[:content_type]).to eq("image/png")
+    end
```
zlekki, line 36: if its ok with you I would like us to follow standards from https://docs.rubocop.org/rubocop-rspec/latest/cops_rspec.html, creating clean test that make it easy to spot bugs as responses are more readable, minimum I would like to see here is using aggregated failures for readability, here is example of changed tests. It would be great if you can change those tests as we need setup rubocop correctly on CI for the future purposes but it's not there yet. It would be great to have new code to following some standards instead of previous style

```suggestion
context 'when image is valid' do
    before do
          stub_request(:get, url).to_return(body: "png-bytes", headers: { "Content-Type" => "image/png" })
    end
    
    let(:attachable) { described_class.call(url) }
    
    it "returns element of proper content type" do
      expect(attachable[:content_type]).to eq("image/png")
    end
    
    it 'calls Vips' do
          expect(Vips::Image).to have_received(:new_from_buffer).with("png-bytes", "")
    end
    
    it 'returns element of a class StringIO' do
          expect(attachable[:io]).to be_a(StringIO)
    end
    
    it 'file extention is png' do
      expect(attachable[:filename]).to end_with(".png")
    end
end

```

Michal-Kolomanski: Still double quotes here:
`    it "returns element of proper content type" do`

spec/services/importers/logo_spec.rb
```diff
@@ -3,29 +3,93 @@
...
+  describe "#call" do
+    it "returns an attachable png for a valid png image" do
+      stub_request(:get, url).to_return(body: "png-bytes", headers: { "Content-Type" => "image/png" })
+
+      attachable = described_class.call(url)
+
+      expect(Vips::Image).to have_received(:new_from_buffer).with("png-bytes", "")
+      expect(attachable[:io]).to be_a(StringIO)
+      expect(attachable[:filename]).to end_with(".png")
+      expect(attachable[:content_type]).to eq("image/png")
```
zlekki: I think we're missing some `end` here

PR #3759, opened 2026-08-28
feat: [#3633] replace AMS subscriber daemon with job-based sync pipeline

### LINKED ISSUE #3633: Fix AMS for data sources, deployable applications and catalogues
Currently, there is no live update for data sources, deployable software and catalogues. 

Scope:
- [ ] find the issue. Messages are ariving to the topic? Are they correctly processed? 
- [ ] fix the issue 
- [ ] test on pre-prod

@Michal-Kolomanski add better description
@alicjaswieradcyfronet linked table with current status

schodevio-ackcyf:
## Summary
- Replace the long-running AMS subscriber daemon (`lib/ams/subscriber.rb`, `bin/ams-subscriber`) with a job-based pull/acknowledge pipeline: `Ams::SubscribeAllJob` fans out one `Ams::SubscribeJob` per configured topic, each pulling and acking messages via a new `Ams::Client`
- Route pulled messages to per-resource handlers (catalogue, datasource, deployable application, interoperability record, organisation, service) that create/update/delete the corresponding local records
- Simplify AMS configuration into a single `config/ams.yml`, replacing `config/ams_subscriber.yml`
- Schedule `Ams::SubscribeAllJob` to run every 5 minutes via sidekiq-cron, and mount the sidekiq-cron admin UI alongside the existing Sidekiq dashboard
- Cap Sidekiq retries on the AMS jobs to 3, since the 5-minute cron cadence and AMS's own broker-level redelivery of unacked messages already provide retry semantics
- Fix a bug where the pull response body was parsed twice (once by Faraday's JSON middleware, once manually), which raised a `TypeError` on every real AMS response
- Fix message decoding to operate on the unwrapped message envelope instead of digging through a nested structure that no longer matched the real AMS payload shape

## Test plan
- [x] Run the full AMS spec suite (`bundle exec rspec spec/jobs/ams spec/services/ams spec/lib/tasks/ams_spec.rb`)
- [x] Start the app with `sidekiq-cron` scheduled and confirm `Ams::SubscribeAllJob` fires every 5 minutes without error in the Sidekiq log
- [x] Confirm catalogue, datasource, and deployable application updates from AMS are reflected locally (the scenario originally reported in #3633)

## Summary by CodeRabbit

* **New Features**
  * Added ARGO Messaging System integration for receiving, decoding, acknowledging, and processing messages.
  * Added synchronization support for services, providers, catalogues, datasources, guidelines, and deployable services.
  * Added scheduled subscription polling and commands to create or delete topic subscriptions.
  * Added configurable endpoints, topics, authentication, and subscription settings.
* **Improvements**
  * Background processing now handles subscription work automatically.
  * Synchronization updates now respect source modification timestamps.
  * Simplified service synchronization configuration by removing redundant connection parameters.


schodevio-ackcyf:
QA: Tested with providers - changes mirrored to local database - PASS

- [x] providers
- [x] services
- [x] datasources
- [x] guidelines
- [x] deployable applications
- [x] catalogue (pull is working, but message data is broken)

config/schedule.yml
```diff
@@ -0,0 +1,4 @@
+ams_subscribe_all:
+  cron: "*/5 * * * *"
```
Michal-Kolomanski: Expose via .env variable

Michal-Kolomanski (approved):
Conceptually it is written very well 👏

config/schedule.yml
Michal-Kolomanski: by default lets aim for 2-3 minutes

Michal-Kolomanski:
> QA: Tested with providers - changes mirrored to local database - PASS

please check other data types as well. After testing, delete all your artefacts

spec/jobs/ams/subscribe_all_job_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+require "rails_helper"
+
+RSpec.describe Ams::SubscribeAllJob, type: :job do
+  describe "#perform_now" do
+    before do
+      allow(Ams::SubscribeJob).to receive(:perform_later)
+    end
+
+    it "calls Ams::SubscribeJob once per topic" do
+      described_class.perform_now
```
zlekki: I think it will be cleaner if we add `described_class.perform_now` to before block

app/concerns/callable.rb
```diff
@@ -0,0 +1,15 @@
...
+module Callable
+  extend ActiveSupport::Concern
+
+  included do
+    private_class_method :new
+  end
+
+  class_methods do
+    def call(...)
+      new(...).call
```
zlekki: this is already implemented in `app/services/application_service.rb` which is better place for service related logic, please decide for one place for this code

app/services/ams/process_message.rb
```diff
@@ -0,0 +1,47 @@
...
+    def call
+      handler_klass = RESOURCE_HANDLERS_MAP[subscription_parts[:resource]]
+
+      unless handler_klass
+        resource = subscription_parts[:resource]
+        Ams::Logger.warn("No handler found for resource '#{resource}' (subscription: #{subscription_name})")
+        return false
+      end
+
+      handler_klass.call(
```
zlekki: you can do similar thing for `Ams::Handlers` `create_or_update` and `delete`

app/services/ams/subscribe.rb
```diff
@@ -0,0 +1,44 @@
...
+    def call
+      response = client.pull(subscription_name)
+      raise "AMS pull failed for #{subscription_name}: #{response.status}" unless response.success?
+
+      messages = get_messages(response)
+      ack_ids = messages.filter_map { |message| process(message) }
+      return if ack_ids.blank?
+
+      ack_response = client.acknowledge(subscription_name, ack_ids: ack_ids)
+      raise "AMS acknowledge failed for #{subscription_name}: #{ack_response.status}" unless ack_response.success?
```
zlekki: this raise standard error which makes it impossible to rescue selectively, I suggest some custom errors

app/services/ams/process_message.rb
```diff
@@ -0,0 +1,47 @@
...
+      "service" => Ams::Handlers::Service
+    }.freeze
+
+    def initialize(subscription_name, message:)
+      @subscription_name = subscription_name
+      @message = message
+    end
+
+    def call
+      handler_klass = RESOURCE_HANDLERS_MAP[subscription_parts[:resource]]
```
zlekki: you use RESOURCE_HANDLERS_MAP here, but each handler calls the same group of methods, so you don't need to implement these classes at all, one handler can handle all of it

zlekki (changes requested):
Nice work! Just a few small changes from my side, but overall looks great

app/concerns/callable.rb
schodevio-ackcyf: I would prefer this one to meet the composition over inheritance approach

app/services/ams/process_message.rb
schodevio-ackcyf: I'm not sure about that. `create_or_update` and `delete` take a different set of parameters

app/services/ams/process_message.rb
schodevio-ackcyf: But each method uses a different job to apply changes, so some MAP is still needed

app/concerns/callable.rb
zlekki: This is a broader design discussion, but in Ruby composition usually means composing objects, often via dependency injection, modules are generally mixins.

My main concern here is duplication. Since this is already implemented in ApplicationService, please either use the existing implementation or replace it with this approach, rather than keeping the same behavior in two places.

app/services/ams/process_message.rb
zlekki: Datasource, Provider, Guideline, Service, and DeployableService handlers don't implement different algorithms. They all perform exactly the same operations and only differ in which job namespace they use. So we're representing configuration as subclasses and duplicating the implementation each time. We could express the varying part as data/dependency and have one implementation, or, if keeping the named handler classes is useful, generate them from a mapping instead of copy/pasting them. 

you can do this with config, for example:
```RESOURCE_CONFIG = {
  "datasource" => {
    delete_job: ::Datasource::DeleteJob,
    create_or_update_job: ::Datasource::PcCreateOrUpdateJob
  },
...
```
  
```class Ams::Handlers::Resource < Ams::Handlers::Base
  def initialize(config)
    @config = config
  end

  private

  attr_reader :config

  def delete(id)
    config[:delete_job].perform_later(id)
  end
...
```

app/concerns/callable.rb
zlekki: Beside the mixin vs inheritence discussion, I meant `app/services/*` as a place for a service related logic, I did not meant exactly this class at this point

Michal-Kolomanski:
Please update README.md as well

PR #3760, opened 2026-08-28
feat: add MARKETPLACE_VARIANT support for repo consolidation

zlekki:
## Summary

One codebase serves the marketplace, pl-marketplace and whitelabel-marketplace deployments. The node is selected at boot by `MARKETPLACE_VARIANT` (`marketplace` / `pl` / `whitelabel`, required in production) and exposed as `Mp::Variant`. Behavior that really differs between the repositories is gated inline; code that only looked different was unified.

## Included

- **Variant mechanism** – routes, backoffice policies and permitted attributes, vocabulary sets, delete/suspend/unpublish lifecycle, PL login identities, PL catalogue API, ESS and ordering-API differences, BOS integration, PL registry import mapping, configuration fallbacks.
- **Schema parity** – every variant runs the same schema: PL profile tables, restored PL-only tables, `user_identities`, nullable `users.uid`. The migration history is squashed, so fresh databases come from `db/schema.rb`.
- **Per-deployment UI** – `CUSTOMIZATION_PATH` overrides views, locales, images, JavaScript and stylesheets (esbuild and Sass resolve through it); ViewComponents use template variants; the helpers those views call are consolidated here. The pl and whitelabel customization directories live outside the repository.
- **Dev seeds** per variant and a working ordering-API test setup.
- **CI** – the `variants` job boots pl and whitelabel and runs the request, lib, helper and routing specs; marketplace-only examples are tagged `variant: :marketplace`. Spring removed.

## Docs

- `docs/deployment_variants.md` – per-variant behavior and deployment settings.
- `docs/repository_consolidation_status.md` – what is done and what remains (deployed database checks, per-deployment settings, JMS with real messages, PL search index, pages not covered by request specs).

## Before deploying

Set `MARKETPLACE_VARIANT` and `CUSTOMIZATION_PATH` per node; check the deployed databases (`schema_migrations`, PL profile rows, duplicate provider pids) before migrating.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


app/controllers/backoffice/services_controller.rb
```diff
@@ -6,30 +6,32 @@ class Backoffice::ServicesController < Backoffice::ApplicationController
...
   before_action :find_and_authorize, only: %i[show edit update destroy]
   before_action :sort_options, :favourites
   before_action :load_query_params_from_session, only: :index
   before_action :provider_scope
   before_action :catalogue_scope
-  prepend_before_action(only: [:index]) { authorize(Service) }
+  # pl authorizes the list after authenticate_user!, so an unauthenticated user
+  # goes through Check-in and comes back; the others authorize first.
+  before_action :authorize_collection, only: :index, if: -> { Mp::Variant.pl? }
+  prepend_before_action(only: [:index], unless: -> { Mp::Variant.pl? }) { authorize(Service) }
```
schodevio-ackcyf: This approach should be the same for all instances. In backoffice we cannot authorize Service without authentication

PR #3761, opened 2026-09-02
fix: [#3744] rework check-in membership verification

### LINKED ISSUE #3744: New added VO membership not working on the same session
To reproduce:
- log in with account without VO membership
- try "run the service"
- request VO membership
- wait for the request approval 
- try "run the service" again - it will redirect to get VO membership again (log out and log in fix the bug)

schodevio-ackcyf:
## Summary
- Replace the inline Net::HTTP token-introspection logic in `check_vo_membership!` with `Checkin::CheckVoMembership`, backed by `Checkin::Client` (Faraday), refreshing the access token and re-introspecting once — fixes VO membership checks failing on an existing/aging session (#3744)
- Store the checkin `refresh_token` in the session on sign-in, and persist refreshed tokens back into the session after every membership check
- Sign the user out when the session has expired, rather than only redirecting
- Source Check-in's OIDC endpoints from live discovery (`CHECKIN_ISSUER_URI`/`CHECKIN_DISCOVERY`) instead of a hardcoded old/new AAI endpoint toggle, and read `VO_GROUP_NAME`/`BECOME_VO_MEMBER_URL` directly from `ENV`; drop `Checkin::Config` now that nothing needs it
- Return a `:misconfiguration` status (and a dedicated user-facing alert) when the VO group name or become-member URL isn't configured, instead of silently doing nothing
- Move `check_vo_membership!` out of `Services::ApplicationController` (where it ran on every wizard step) into `Services::ChooseOffersController#show`, so the VO check now happens once per order instead of on every step; `prepend_before_action` plus a signed-in guard keep it running before service authorization instead of after
- Disable turbo-prefetch on the "Run the service" link so hovering it no longer triggers a VO membership check
- Load `.env.build` during Docker asset precompilation, since Check-in env vars are now mandatory
- Document the Check-in integration and its env vars in the README, next to the AMS subscriptions section

Also on this branch (unrelated to #3744):
- fix(message-policy): correct a broken join that silently dropped the `messageable_type` filter on OMS admin message scoping
- refactor(config): load locale files via `Rails.root.glob`

## Test plan
- [x] Log in with an account without VO membership, run a service, and confirm the "request VO membership" redirect still works
- [x] Reproduce the original bug scenario from #3744 (VO membership check on an existing/aging session) and confirm it now refreshes the token instead of failing
- [x] `bundle exec rspec spec/services/checkin spec/requests/services/ spec/controllers/services/choose_offers_controller_spec.rb`

Closes #3744

## Summary by CodeRabbit

* **New Features**
  * Added Check-in authentication and VO membership verification before users can choose offers.
  * Sessions now retain refreshed authentication tokens after sign-in or verification.
  * Added configurable Check-in/OIDC setup with documented environment variables.

* **Bug Fixes**
  * Improved message filtering for project and project-item offers.
  * Prevented unnecessary prefetching when running a deployable service.

* **Documentation**
  * Added guidance for Check-in authentication, membership enforcement, redirects, and configuration.


app/services/checkin/check_vo_membership.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def call
+      if refresh_token.blank?
+        Checkin::Logger.warn("Missing credentials")
+        return session_expired_result
+      end
+
+      # 1. Refresh token
+      response = client.refresh_token(refresh_token)
+      return session_expired_result unless response.success?
```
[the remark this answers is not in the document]
schodevio-ackcyf: It's not relevant for now

app/services/checkin/check_vo_membership.rb
```diff
@@ -0,0 +1,72 @@
...
+        refresh_token: refresh_token,
+        become_vo_member_url: become_vo_member_url
+      )
+    end
+
+    def status(introspection)
+      entitlements = introspection["entitlements"] || []
+      group_tag = "group:#{vo_group_name}"
+
+      entitlements.any? { _1.include?(group_tag) } ? :member : :not_member
```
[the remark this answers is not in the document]
schodevio-ackcyf: This approach was agreed with Checkin provider

.env.test
```diff
@@ -0,0 +1,32 @@
...
+VO_GROUP_NAME=testgroup
+
+### Checkin (Keycloak) envs - MANDATORY if CHECKIN_DISCOVERY is false
+### ---------------------------------------------------------------------------
+# CHECKIN_AUTHORIZATION_ENDPOINT=/authorize
+# CHECKIN_INTROSPECTION_ENDPOINT=/token/introspect
+# CHECKIN_TOKEN_ENDPOINT=/token
+# CHECKIN_USERINFO_ENDPOINT=/userinfo
+# CHECKIN_JWKS_URI=/jwk
+# CHECKIN_END_SESSION_ENDPOINT=/logout
```
zlekki: blank line is missing

.env.example
```diff
@@ -6,4 +6,27 @@ AMS_SUBSCRIBE_ALL_ENABLED=true
...
+VO_GROUP_NAME=<YOUR_VO_GROUP_NAME>
+
+### Checkin (Keycloak) envs - MANDATORY if CHECKIN_DISCOVERY is false
+### ---------------------------------------------------------------------------
+# CHECKIN_AUTHORIZATION_ENDPOINT=<YOUR_KEYCLOAK_ISSUER_URI>/authorize
+# CHECKIN_INTROSPECTION_ENDPOINT=<YOUR_KEYCLOAK_ISSUER_URI>/token/introspect
+# CHECKIN_TOKEN_ENDPOINT=<YOUR_KEYCLOAK_ISSUER_URI>/token
+# CHECKIN_USERINFO_ENDPOINT=<YOUR_KEYCLOAK_ISSUER_URI>/userinfo
+# CHECKIN_JWKS_URI=<YOUR_KEYCLOAK_ISSUER_URI>/jwk
+# CHECKIN_END_SESSION_ENDPOINT=<YOUR_KEYCLOAK_ISSUER_URI>/logout
```
zlekki: blank line is missing

spec/services/checkin/client_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    it "returns the refreshed tokens in the response body" do
+      expect(JSON.parse(response.body)).to eq("access_token" => "new-token", "refresh_token" => "new-refresh")
+    end
+
+    it "authenticates with HTTP basic auth using the checkin client credentials" do
+      response
+
+      expect(WebMock)
+        .to have_requested(:post, "#{issuer}/token")
+        .with(headers: { "Authorization" => "Basic #{Base64.strict_encode64("#{identifier}:#{secret}")}" })
```
zlekki: You shouldn't run test on testing library, you should do `a_request`, instead of that you can use for example something like
 ```
expect(a_request(:post, "www.example.com").
  with(:body => {"a" => ["b", "c"]},
    :headers => {'Content-Type' => 'application/json'})).to have_been_made
```

The target is the fake here
The bahaviour that matter is not tested, like "the token was fetched",  or "a 401 is handled".
The stub accepts any credentials - no check if correct values were picked
Last but not least is just not how Webmock works https://github.com/bblimke/webmock#stubbing-requests-based-on-method-uri-body-and-headers

this problem occur in different places, please fix all of them

File: spec/services/checkin/check_vo_membership_spec.rb [hunk 6]
```diff
@@ -0,0 +1,214 @@
...
+
+  let(:access_token) { "a-token" }
+  let(:refresh_token) { "a-refresh-token" }
+  let(:client) { instance_double(Checkin::Client) }
+
+  before do
+    allow(Checkin::Client).to receive(:new).and_return(client)
+    allow(ENV).to receive(:fetch).and_call_original
+    allow(ENV).to receive(:fetch).with("VO_GROUP_NAME", nil).and_return("eosc-beyond.eu")
+    allow(ENV).to receive(:fetch).with("BECOME_VO_MEMBER_URL", nil).and_return("https://example.com/enroll")
```
zlekki: during tests best way to handle such values is to do some dependency injection and create cofig object that your class will use. 
There is a nice way for doing this: https://www.rubydoc.info/docs/rails/8.0.1/Rails%2FApplication:config_for

You can use erb in your yml that will handle env variable and use `Rails.configuration` for some mocks instead o overriding env variables as they should be treated out of app scope.

Best would be to setup some small config class for this and mock it in tests

app/controllers/services/choose_offers_controller.rb
```diff
@@ -40,6 +42,30 @@ def update
...
+    when :misconfiguration
+      redirect_to root_path, alert: _("We can't verify your VO membership. Please contact admin.")
+    when :session_expired
+      sign_out(current_user)
+      redirect_to root_path, alert: _("Your session has expired. Please sign in again.")
+    when :verification_failed
+      redirect_to root_path, alert: _("Your VO membership verification has failed.")
+    when :not_member
+      redirect_to result.become_vo_member_url, allow_other_host: true
+    end
```
zlekki, line 70: Maybe it would be good to add some additional behaviour when status is return but don't match any of those?

app/policies/api/v1/message_policy.rb
```diff
@@ -8,10 +8,24 @@ def resolve
...
+          .or(scope.where(offers_project_items: { primary_oms_id: user.administrated_oms_ids }))
+          .joins(<<~SQL.squish)
+            LEFT OUTER JOIN project_items ON project_items.id = messages.messageable_id
+              AND messages.messageable_type = 'ProjectItem'
+            LEFT OUTER JOIN offers ON offers.id = project_items.offer_id
+            LEFT OUTER JOIN projects ON projects.id = messages.messageable_id
+              AND messages.messageable_type = 'Project'
+            LEFT OUTER JOIN project_items project_items_projects ON project_items_projects.project_id = projects.id
+            LEFT OUTER JOIN offers offers_project_items ON offers_project_items.id = project_items_projects.offer_id
+          SQL
```
zlekki, line 28: Policies aren't a place for building complicated queries, please move to model or to query object.

There is a problem with relations in our codebase that would be nice to fix, then you cand do something like that: ```oms_ids = user.administrated_oms_ids

project_item_ids =
  ProjectItem
    .joins(:offer)
    .where(offers: { primary_oms_id: oms_ids })
    .select(:id)

project_ids =
  Project
    .joins(project_items: :offer)
    .where(offers: { primary_oms_id: oms_ids })
    .select(:id)

scope
  .where(
    messageable_type: "ProjectItem",
    messageable_id: project_item_ids
  )
  .or(
    scope.where(
      messageable_type: "Project",
      messageable_id: project_ids
    )
  )
  .distinct```
as Rails is lazy loading the queries so it won't create multiple queries but one and is much more clear whats happening here

those relations on message model are a big problem and need refactoring: ```  belongs_to :messageable, polymorphic: true
  belongs_to :project_item,
             -> { where(messages: { messageable_type: "ProjectItem" }).includes(:messages) },
             foreign_key: "messageable_id",
             optional: true
  belongs_to :project,
             -> { where(messages: { messageable_type: "Project" }).includes(:messages) },
             foreign_key: "messageable_id",
             optional: true
  belongs_to :approval_request,
             -> { where(messages: { messageable_type: "ApprovalRequest" }).includes(:messages) },
             foreign_key: "messageable_id",
             optional: true```
it's rewriting methods that are already created by `belongs_to :messageable, polymorphic: true` 

they're structured in a way that may cause a collision of elements

app/services/checkin/check_vo_membership.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module Checkin
+  class CheckVoMembership < ApplicationService
+    Result = Struct.new(:status, :access_token, :refresh_token, :become_vo_member_url, keyword_init: true)
```
zlekki: the name Result is in my opinion is confusing, too generic, maybe some name relate to token or config would be better.
maybe 
`AuthorizationModel`, or `AuthorizationResulltl` or `AuthorizationConfig`

Can you explain why do you decided to use Struct instead of Hash here?

app/services/checkin/check_vo_membership.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      # 2. Introspect token
+      response = client.introspect(access_token)
+      return result_for(:verification_failed) unless response.success?
+
+      introspection = JSON.parse(response.body)
+      return result_for(:session_expired) unless introspection["active"]
+
+      result_for(status(introspection))
+    rescue Faraday::Error, JSON::ParserError => e
```
zlekki: `Faraday::Error` is too generic, some errors unrelated to the verification will go through it

app/services/checkin/client.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
```
zlekki, line 1: it's not a service but client implementation that should go somewhere else, for example `lib/`

app/services/checkin/check_vo_membership.rb
```diff
@@ -0,0 +1,76 @@
...
+      return result_for(:misconfiguration) if misconfigured?
+      return result_for(:session_expired) if refresh_token.blank?
+
+      # 1. Refresh token
+      response = client.refresh_token(refresh_token)
+      return result_for(:session_expired) unless response.success?
+
+      credentials = JSON.parse(response.body)
+      @access_token = credentials["access_token"].presence || access_token
+      @refresh_token = credentials["refresh_token"].presence || refresh_token
```
zlekki: as you're changing instance variable here you can easy move this part to some method which will make it more clear whats going on here

zlekki (changes requested):
There are a few small things to fix, but the main issue is the relations/models and related scopes — there are some serious problems there that need to be addressed.

If it helps, we can also pair on this or have a quick chat and work through the solution together.

app/policies/api/v1/message_policy.rb
schodevio-ackcyf: This is out of the scope and can be done in separate ticket. It was fixed with Claude's help due to failing spec

app/services/checkin/client.rb
schodevio-ackcyf: It was done in similar way like AMS, I don't want to make a mess with different approaches

app/policies/api/v1/message_policy.rb
zlekki: @schodevio-ackcyf real fix for it is to add `belongs_to :messageable, polymorphic: true` where necessary, if you think this is out of a scope of that ticket, this is too.

app/services/checkin/check_vo_membership.rb
schodevio-ackcyf: - **Fixed shape, typo-safe**: `result.status` fails loudly (`NoMethodError`) if you typo it; `result[:staus]` on a Hash silently returns `nil` and the bug surfaces somewhere downstream, harder to trace.
- **Method-call syntax**: `result.access_token` reads better than `result[:access_token]` at every call site (controller, specs).
- **Self-documenting contract**: the four fields are declared once, in one place — anyone reading the class sees the full shape of what `call` can return, instead of inferring it from scattered hash literals.
- **`keyword_init: true`** forces named args at construction (`Result.new(status: :member)`), preventing accidental positional mix-ups.

app/services/checkin/client.rb
zlekki: I'm afraid that mess is already there, for example here is client connection in model: https://github.com/cyfronet-fid/marketplace/blob/development/app/models/report/client.rb#L23
same with Jira https://github.com/cyfronet-fid/marketplace/blob/development/app/models/jira/client.rb
so theoretically you should put this code in a models directory
but you can find connection to third party in lib too:
https://github.com/cyfronet-fid/marketplace/blob/development/lib/import/catalogues.rb#L10 but currently most of this calls client class that are in models

zlekki (approved):

zlekki (approved):

PR #3777, opened 2026-09-11
feat(ams): add tagged logging to sync pipeline

schodevio-ackcyf:
## Summary
- Log enqueued create/update/delete jobs and unsupported resource/action warnings, tagged `[AMS]` via `Rails.logger.tagged`
- Log message counts pulled and processed on each subscription cycle
- Fix the test environment's logger config so `Rails.logger.tagged` doesn't raise `NoMethodError` under RSpec
- Update decode_message/process_message specs to assert against the tagged logger instead of the removed `Ams::Logger` module

## Test plan
- [x] `bundle exec rspec spec/services/ams/` — 44 examples, 0 failures
- [ ] Trigger an AMS subscription pull and confirm `[AMS]`-tagged logs appear with resource ids and message counts

### DIFF
--- a/app/services/ams/logger.rb
+++ b/app/services/ams/logger.rb
@@ -1,19 +1,21 @@
 # frozen_string_literal: true
 
+require "colorize"
+
 module Ams
   module Logger
     module_function
 
     def info(message)
-      Rails.logger.info("[AMS] #{message}")
+      Rails.logger.info("[AMS] #{message}".cyan)
     end
 
     def warn(message)
-      Rails.logger.warn("[AMS] #{message}")
+      Rails.logger.warn("[AMS] #{message}".cyan)
     end
 
     def error(message)
-      Rails.logger.error("[AMS] #{message}")
+      Rails.logger.error("[AMS] #{message}".cyan)
     end
   end
 end
--- a/app/services/ams/process_message.rb
+++ b/app/services/ams/process_message.rb
@@ -61,14 +61,22 @@ module Ams
 
     def create_or_update_later
       job = JOBS_MAP.dig(resource, :create_or_update_job)
-      return job.perform_later(message.dig("data", resource_key), status, modified_at) if job
+
+      if job.present?
+        Ams::Logger.info("Enqueue #{resource} id=#{message.dig("data", "id")} for create/update")
+        return job.perform_later(message.dig("data", resource_key), status, modified_at)
+      end
 
       Ams::Logger.warn("Unsupported '#{action}' for resource '#{resource}'")
     end
 
     def delete_later
       job = JOBS_MAP.dig(resource, :delete_job)
-      return job.perform_later(message.dig("data", "id")) if job
+
+      if job.present?
+        Ams::Logger.info("Enqueue #{resource} id=#{message.dig("data", "id")} for delete")
+        return job.perform_later(message.dig("data", "id"))
+      end
 
       Ams::Logger.warn("Unsupported '#{action}' for resource '#{resource}'")
     end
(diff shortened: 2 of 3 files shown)


zlekki (changes requested):
Short summary with Claude:

**Problems with the current version**

- `.cyan` puts colour codes into the log text. They end up in log files and in any tool that collects logs, where they show up as `\e[36m` garbage.
- All levels use the same colour, so warnings and errors do not stand out.
- `require "colorize"` here adds it to the whole app. The Gemfile has `require: false` so it is only used in console scripts.
- The `Enqueue …` lines repeat what ActiveJob already logs (`Enqueued <Job> (Job ID: …) to Sidekiq(…) with arguments: …`).

**Proposal**

- `Ams::Subscribe` wraps the pull in `ActiveSupport::Notifications.instrument("pull.ams", …)` and puts the message counts in the payload. Unsupported actions and decode errors are sent as events in the same way.
- One `AmsLogSubscriber` turns these events into log lines like `[AMS] subscription=service-update received=3 acked=2`.
- `Ams::Logger` and `colorize` are removed.
- Colour by level (yellow WARN, red ERROR) is added in `config/environments/development.rb`, only when writing to a terminal.

**Why**

- Services report events, and one class decides how they look in the log.
- When we add OpenTelemetry, it will subscribe to the same events. Nothing in AMS needs to change.
- Plain `key=value` lines are easy to grep now and easy for log tools to parse later.

As discussed in DMs, if you have any questions do not hesitate to ask, we can chat or jump on a call.

zlekki:
As discussed before I've prepared (with Claude) information that separate cleanly elements that are bugs that need to be changed for approve and merge, and suggested changes that may be done later on, discussed and planned.

## Required before merge

- **Remove `colorize` and `.cyan` from `Ams::Logger`.** Colour codes go into the log text, so they end up in log files and in every tool that collects logs as `\e[36m` garbage. `require "colorize"` also loads a global `String` patch into web and Sidekiq processes, while the Gemfile has `require: false` on purpose.
- **Remove the two `Enqueue … id=…` lines.** ActiveJob already logs `Enqueued <Job> (Job ID: …) to Sidekiq(…) with arguments: …` for every `perform_later`, with more detail.

The `Found …` / `Processed …` counts can stay.

## Follow-up (its ok as a separate ticket)

- `Ams::Subscribe` reports the pull with `ActiveSupport::Notifications.instrument("pull.ams", …)` and puts the counts in the payload. Unsupported actions and decode errors become events in the same way.
- One `AmsLogSubscriber` turns these events into lines like `[AMS] subscription=service-update received=3 acked=2`. `Ams::Logger` goes away.
- Colour by level (yellow WARN, red ERROR) is added in `config/environments/development.rb`, only when writing to a terminal.

Why: services report events, one class decides how they look in the log. When we add OpenTelemetry(it's already planned), it subscribes to the same events and nothing in AMS changes. `key=value` lines are easy to grep now and easy for log tools to parse later.

## Reading AMS logs in the console today

All AMS lines already carry the `[AMS]` prefix, so filtering and highlighting is done in the terminal, not in the code:

```sh
# only AMS lines
bin/dev | grep '\[AMS\]'

# from sidekiq
bundle exec sidekiq -q ams_subscriber | grep '\[AMS\]'
# all lines, AMS highlighted
bin/dev | grep --color=always -E '\[AMS\]|$'

# from the log file
tail -f log/development.log | grep '\[AMS\]'

# after the fact
grep '\[AMS\]' log/development.log

As before I'm opened to some pair programming session or discussion/call where we can discuss these suggestions.

schodevio-ackcyf:
The failing MessagePolicy spec is already fixed in another PR: https://github.com/cyfronet-fid/marketplace/pull/3761

Michal-Kolomanski (approved):

